PR #1, opened 2018-05-02
Clone full history of dlpx-app-gate repository

prakashsurya:
Originally, we performed a very shallow clone of the dlpx-app-gate
repository, optimizing for latency of the build, and the size of the
resultant VM artifacts.

Not having the repository's history on our internal VM images is a
divergence from what we do for other images, and can pose a problem for
some of our internal developer tooling.

Thus, this change reverts this functionality; now we do a full clone
of the dlpx-app-gate repository, optimizing for developer productivity
and utility of the resultant VM images, rather than build latency.


prakashsurya:
bors ping

PR #2, opened 2018-05-03
install perf

ahrens:
test results: http://ops.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build/job/master/job/pre-push/3/

I realize that this package is specific to the kernel version, so we'd need to update it when we update the base kernel.  Is there a better way to do this?


prakashsurya:
@ahrens see #3 .. I still need to test it, but my hope is that approach will work better than explicitly listing the kernel version like what is done here.

PR #3, opened 2018-05-03
Explicitly install "generic" kernel and tools

prakashsurya:
Previously we were installing whatever kernel version that was being
included by the "ubuntu-minimal" package. This poses a problem since we
also need to include the linux-tools package (e.g. to obtain "perf"),
but we have to specify the specific kernel verison of linux-tools that
we want (e.g. "linux-tools-generic", "linux-tools-lowlatency", etc).

Thus, to ensure we match the kernel version, and the kernel specific
tools, this change moves us to be explicit about the kernel package that
we install. This way, it's clear that the installed kernel version will
be compatible with the kernel tools also installed.


ahrens:
looks great to me!

PR #6, opened 2018-05-03
Add fio to the image

jwk404:
fio is used by the ZFS performance suites as well as the storage tests.
Also updated the README, to clarify which ansible roles are used when.


prakashsurya:
I’ll plan on moving this list of virtualization package dependencies into the actual package definition of the virtualization package tomorrow.

Currently we have this list here, and then basically the same list in the app-gate. I think we can start to rely solely on the app-gate list now.

PR #8, opened 2018-05-03
Rely on virtualization package dependencies

prakashsurya:
This change removes the hard coded list of dependencies for the
virtualization package. The virtualization package should now contain
the list of its dependencies, so there's no reason to have a hard coded
list here. When pulling in the virtualization package, we'll pull in
that package's dependencies as well.

### DIFF
--- a/live-build/base/config/package-lists/tools.list.chroot
+++ b/live-build/base/config/package-lists/tools.list.chroot
@@ -24,7 +24,10 @@ atop
 dstat
 emacs
 gdb
+glances
 iftop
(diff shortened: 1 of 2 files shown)


prakashsurya:
I built the `internal-qa` variant with this change, and verified that no dependencies were left unsatisfied:
```
$ sudo apt-get -f install
Reading package lists... Done
Building dependency tree
Reading state information... Done
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```

PR #25, opened 2018-05-17
Use "--enable-debuginfo" when building SPL and ZFS

prakashsurya:
By default, when building linux kernel modules, the compiler is pretty
aggressive when inlining functions, which can make it more difficult to
observe and debug a running system. In the interest of being able to
better debug production systems, this change modifies the configure flags
used when building the SPL and ZFS repositories.

By enabling the "--enable-debuginfo" configure flag, the two repositories
will be compiled using the "-fno-inline" compiler flag, which will make
stack traces and tools that leverage function boundary kprobes more
useful.

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.install-delphix-zfs-packages/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.install-delphix-zfs-packages/tasks/main.yml
@@ -24,7 +24,7 @@
 # compile the ZFS on Linux repositories from source, and install the
 # resultant build artifacts. Below is the logic to that work.
 #
-- shell: "sh autogen.sh && ./configure --prefix= && make"
+- shell: "sh autogen.sh && ./configure --enable-debuginfo --prefix= && make"
   args:
     chdir: "/export/home/{{ lookup('env', 'APPLIANCE_USERNAME') }}/spl"
     creates: "/export/home/{{ lookup('env', 'APPLIANCE_USERNAME') }}/spl/module/Module.symvers"
@@ -37,7 +37,7 @@
     chdir: "/export/home/{{ lookup('env', 'APPLIANCE_USERNAME') }}/spl"
   when: spl.changed
 
-- shell: "sh autogen.sh && ./configure --prefix= && make"
+- shell: "sh autogen.sh && ./configure --enable-debuginfo --prefix= && make"
   args:
     chdir: "/export/home/{{ lookup('env', 'APPLIANCE_USERNAME') }}/zfs"
     creates: "/export/home/{{ lookup('env', 'APPLIANCE_USERNAME') }}/zfs/module/Module.symvers"


prakashsurya:
I still need to test and verify this works as I expect. I'll post a comment after I do that.

prakashsurya:
I've built the internal-qa variant locally using the process described in the README. Here's some before/after verification:

Before:
```
$ readelf -s /lib/modules/4.15.0-20-generic/extra/zfs/zfs.ko | grep zil_process_commit_list | wc -l
0
```

After:
```
$ readelf -s /lib/modules/4.15.0-20-generic/extra/zfs/zfs.ko | grep zil_process_commit_list
  6456: 00000000000ee4f0   488 FUNC    LOCAL  DEFAULT    2 zil_process_commit_list
```

This is the main reason for this change; in the "before" case the `zil_process_commit_list` function was inlined, whereas it was not after applying this change.

Additionally, since this change will also enable the `-g` compile-time flag, it's worth pointing out the size differences in the binaries.

Before:
```
$ du -s /lib/modules/4.15.0-20-generic/extra/*/* | sort -n
12      /lib/modules/4.15.0-20-generic/extra/avl/zavl.ko
101     /lib/modules/4.15.0-20-generic/extra/nvpair/znvpair.ko
102     /lib/modules/4.15.0-20-generic/extra/zcommon/zcommon.ko
259     /lib/modules/4.15.0-20-generic/extra/lua/zlua.ko
259     /lib/modules/4.15.0-20-generic/extra/spl/spl.ko
387     /lib/modules/4.15.0-20-generic/extra/icp/icp.ko
387     /lib/modules/4.15.0-20-generic/extra/splat/splat.ko
387     /lib/modules/4.15.0-20-generic/extra/unicode/zunicode.ko
3461    /lib/modules/4.15.0-20-generic/extra/zfs/zfs.ko
```

After:
```
$ du -s /lib/modules/4.15.0-20-generic/extra/*/* | sort -n
259     /lib/modules/4.15.0-20-generic/extra/spl/spl.ko
387     /lib/modules/4.15.0-20-generic/extra/avl/zavl.ko
387     /lib/modules/4.15.0-20-generic/extra/splat/splat.ko
1155    /lib/modules/4.15.0-20-generic/extra/nvpair/znvpair.ko
1155    /lib/modules/4.15.0-20-generic/extra/unicode/zunicode.ko
4101    /lib/modules/4.15.0-20-generic/extra/zcommon/zcommon.ko
8200    /lib/modules/4.15.0-20-generic/extra/lua/zlua.ko
11658   /lib/modules/4.15.0-20-generic/extra/icp/icp.ko
54052   /lib/modules/4.15.0-20-generic/extra/zfs/zfs.ko
```

This increase in size is negligible considering our root image is over 6GB in size:
```
$ df -h /
Filesystem         Size  Used Avail Use% Mounted on
rpool/ROOT/ubuntu   68G  6.3G   62G  10% /
```
and the virtualization package is well over 1GB in size:
```
$ apt-cache show delphix-virtualization | grep Installed-Size
Installed-Size: 1878187
```

prakashsurya:
@ahrens Yes. That's my intention, to use this flag for production builds. While I haven't measured the performance penalty this may (or may not) have, I assume it's a reasonable cost to pay, in order to better debug and observe production systems.

I agree, we should update `git-zfs-make` to always use `--enable-debuginfo`. I can probably get to that next week; I'll make a note so I don't forget.

sdimitro:
So looking at the ZoL wiki:
```
tip: --enable-debug can be set to enable all ASSERTs and additional correctness tests. This option is also supported when building ZFS.
```

I'm pretty sure we have some heavy-weight verification functions in ZFS that take place only in DEBUG kernels of OpenZFS (either through _KERNEL_DEBUG macros or from calls within ASSERT()). Is this the case also in ZoL. If yes I think that it may worth rethinking this.

For example, if our target is to not be aggressive at inlining can we just pass the specific flag to the C compiler and if we still want all the debug-info we can make them optional in `git-zfs-make` and for production builds.

sdimitro:
oops.. ok nevermind then :-)

PR #26, opened 2018-05-21
Add "Statement of Support" section to README.md

prakashsurya:


prakashsurya:
See [here](https://github.com/prakashsurya/appliance-build/blob/statement-of-support/README.md) for an example of what this looks like.

PR #29, opened 2018-05-30
Remove SPL

ahrens:
Since the SPL repo has been merged to ZFS, we don't need to build it any
more.  Building the ZFS repo includes SPL.


ahrens:
superseded by #31

PR #33, opened 2018-06-01
Modify run-live-build to default to new linux buckets

grodr:
To test this change I:
1. cloned the bootstrap-18-04 vm
2. cloned this repo
3. applied my diff
4. ran the live build with my JENKINSID and GITBRANCH set


prakashsurya:
bors delegate+

PR #35, opened 2018-06-01
Modify run-live-build to default to new linux buckets

prakashsurya:


prakashsurya:
bors delegate=grodr

@grodr this is identical to the change that was just reverted. when all the other dependencies for this are landed, you can comment with "bors r+" and the bot will test and land this.

PR #37, opened 2018-06-06
LX-1039 enable auditd on linux

don-brady:
Install the `auditd` package as part of the _minimal-common_ ansible role. Update the `/etc/audit/` config parameters for log rotation and add rules for recording all commands that are executed (i.e. via execve).

**Testing**
Booted  the build artifacts with QEMU and verified in the console that `auditd` was installed and that the config files were updated as expected.  Also used `ausearch(8)` and `aureport(8)` tools to confirm that command auditing is occurring.

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
@@ -233,3 +233,36 @@
   with_items:
     - { key: "net.ipv4.conf.all.promote_secondaries", value: "1" }
     - { key: "net.ipv4.conf.default.promote_secondaries", value: "1" }
+
+#
+# Configure command audit logging
+#
+# We want to record all commands executed on the appliance. Opt out for
+# setsid since all ExecuteUtils.execute wrap each call with setsid.
+#
+# Note we can't use the "service" module, as that fails when running
+# inside the chroot environment, so we enable auditd using "command".
+#
+- command: systemctl enable auditd.service
+
+- lineinfile:
+    dest: /etc/audit/auditd.conf
+    regexp: "{{ item.regex }}"
+    line: "{{ item.line }}"
+  with_items:
+    - { regex: "^num_logs =", line: "num_logs = 6" }
+    - { regex: "^max_log_file =", line: "max_log_file = 3072" }
+    - { regex: "^max_log_file_action =", line: "max_log_file_action = rotate" }
+    - { regex: "^log_format =", line: "log_format = RAW" }
+
+- blockinfile:
+    path: /etc/audit/rules.d/audit.rules
+    insertafter: EOF
+    block: |
+      ## Record all executed commands (excluding setsid)
+      -a exit,never -S execve -F exe=/usr/bin/setsid
+      -a exit,always -S execve
+
+      ## Record command exit failures (execve result is not command result)
+      -a exit,always -F a0!=0 -S exit_group
+      -a exit,always -F a0!=0 -S exit
--- a/live-build/base/config/package-lists/minimal.list.chroot
+++ b/live-build/base/config/package-lists/minimal.list.chroot
@@ -42,3 +42,4 @@ net-tools
 open-vm-tools
 openssh-server
 ubuntu-minimal
+auditd


live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -233,3 +233,41 @@
...
+- lineinfile:
+    dest: /etc/audit/rules.d/audit.rules
+    line: "{{item}}"
+    insertafter: EOF
+    state: present
+  with_items:
+    - "## Record all executed commands"
+    - "-a exit,never -S execve -F exe=/usr/bin/setsid"
+    - "-a exit,always -S execve"
+    - "-a exit,always -F a0!=0 -S exit_group"
```
sebroy: Maybe add a comment here explaining that this enables the tracking of the exit code of commands since the execve syscall will succeed if it's able to execute a command even if that command exits with a failure.

don-brady:
@sebroy good point, added:
```
+    - "## Record command exit failures (execve result is not command result)"
     - "-a exit,always -F a0!=0 -S exit_group"
     - "-a exit,always -F a0!=0 -S exit"
```

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -233,3 +233,42 @@
...
+# Configure command audit logging
+#
+# We want to record all commands executed on the appliance. Opt out for
+# setsid since all ExecuteUtils.execute wrap each call with setsid.
+#
+# Note we can't use the "service" module, as that fails when running
+# inside the chroot environment, so we enable auditd using "command".
+#
+- apt:
+    name: auditd
```
prakashsurya: This isn't _wrong_, but I'd suggest we install this by adding the package to the file: `live-build/base/config/package-lists/minimal.list.chroot` .. just so it's (imo) more consistent with how we install other packages for the "minimal" variant. What do you think?

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -233,3 +233,42 @@
...
+    state: installed
+    update_cache: no
+
+- command: /bin/systemctl enable auditd.service
```
prakashsurya: Can we do this with the [service module](https://docs.ansible.com/ansible/latest/modules/service_module.html)? or does that not work when using a chroot like we're doing?

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -233,3 +233,42 @@
...
+
+- lineinfile:
+    dest: /etc/audit/auditd.conf
+    regexp: "{{item.regex}}"
```
prakashsurya: To be consistent with the style used elsewhere, can you put a space after `{{` and before `}}`, here and elsewhere that you use substitution? e.g.
```
regexp: "{{ item.regex }}"
```

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -233,3 +233,42 @@
...
+    line: "{{item.line}}"
+  with_items:
+    - { regex: "^num_logs =", line: "num_logs = 6" }
+    - { regex: "^max_log_file =", line: "max_log_file = 3072" }
+    - { regex: "^max_log_file_action =", line: "max_log_file_action = rotate" }
+    - { regex: "^log_format =", line: "log_format = RAW" }
+
+- lineinfile:
```
prakashsurya: This seems like it'd be better suited to use the [copy module](https://docs.ansible.com/ansible/latest/modules/copy_module.html), rather than `lineinfile`; e.g.
```
copy:
  dest: /etc/audit/rules.d/audit.rules
  mode: 0644
  content: |
    ## Record all executed commands (excluding setsid)
￼    -a exit,never -S execve -F exe=/usr/bin/setsid
    -a exit,always -S execve

    ## Record command exit failures (execve result is not command result)
    -a exit,always -F a0!=0 -S exit_group
    -a exit,always -F a0!=0 -S exit
```
What do you think?

We already do this sort of thing, to insert whole files, like [here](https://github.com/delphix/appliance-build/blob/master/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml#L82).

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
don-brady: @prakashsurya  Seems logical.  I moved the auditd package install location as suggested above.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
don-brady: Will do.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
don-brady: This is appending content to an existing file. Would copy preserve and append here?

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
don-brady: Doesn't seem to work as a service module in our context.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
prakashsurya: oh, ok. I didn't realize the file will already exist. in that case, I'd suggest we use [blockinfile](https://docs.ansible.com/ansible/2.3/blockinfile_module.html) since it seems like we're trying to append a whole block, rather than inserting independent lines. This might not have a practical difference, but "blockinfile" feels like a better fit than "lineinfile" to me. What do you think?

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -233,3 +233,37 @@
...
+#
+# Configure command audit logging
+#
+# We want to record all commands executed on the appliance. Opt out for
+# setsid since all ExecuteUtils.execute wrap each call with setsid.
+#
+# Note we can't use the "service" module, as that fails when running
+# inside the chroot environment, so we enable auditd using "command".
+#
+- command: /bin/systemctl enable auditd.service
```
prakashsurya: let's use `systemctl` rather than `/bin/systemctl` to be consistent with other invocations of `systemctl` elsewhere in our ansible roles.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
don-brady: I confirmed that using copy as above will over-write the original rules file (not append) which is not what we want.

prakashsurya: Yes. "copy" will replace with the contents provided. "blockinfile" won't do that though; it's similar to "lineinfile" but intended to be used when multiple lines are wanted, like what I think we want to do here.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
don-brady: Git it.  Thanks

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
don-brady: I think blockinfile is what I need (append content to an existing file).
```
+- blockinfile:
+    path: /etc/audit/rules.d/audit.rules
+    insertafter: EOF
+    block: |
+      ## Record all executed commands (excluding setsid)
+      -a exit,never -S execve -F exe=/usr/bin/setsid
+      -a exit,always -S execve
+
+      ## Record command exit failures (execve result is not command result)
+      -a exit,always -F a0!=0 -S exit_group
+      -a exit,always -F a0!=0 -S exit
```

PR #39, opened 2018-06-07
Remove "bcc-development" Ansible role

### LINKED ISSUE #32: Remove "bcc-development" Ansible role
As of commit f85b5cbe50798e13ca9f4808b7ee4827bdaabd8f we no longer need to be providing/building the BCC code base as part of the build process. Folks should now be using the tools provided by the `bpfcc-tools` package, rather than using the tools directly from the BCC source repository.

Thus, we should remove the "bcc-development" Ansible role, and update all consumers of that role.

pzakha:
We now install bcc from an Ubuntu package, `bpfcc-tools`, so we do not need to build it anymore.
We can also remove `llvm` which was only used to build bcc.

Closes #32


prakashsurya:
bors delegate+

@pzakha leave a comment with "bors r+" when you're ready for this to land.

pzakha:
Tested build of internal-dev here:
http://collins.d.delphix.com:32423/job/devops-gate/job/projects/job/dx4linux/job/appliance-build/job/master/job/post-push/3/console (pass)

bors r+

PR #41, opened 2018-06-07
LX-1098 Add swap device to engine

pcd1193182:
This patch adds a swap device to the engine early in the build process. We create the zvol with appropriate configuration and add it to the new machine's /etc/fstab file. In order to format the device, the /dev/zvol pseudo-device needs to exist. However, that device only exists if the bootstrap machine (not the docker container) has the zfsutils package installed. To solve this, an ansible playbook and task were added, which should be executed by the automation to install the package.  Finally, when we've finished the build we have to stop the zed service in order to unload the zfs kernel modules.

### DIFF
--- a/bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
+++ b/bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
@@ -24,6 +24,7 @@
   with_items:
     - docker.io
     - python3-docker
+    - zfsutils-linux
 
 - docker_image:
     path: "{{ toplevel.stdout }}/docker"
--- a/live-build/misc/live-build-hooks/90-raw-disk-image.binary
+++ b/live-build/misc/live-build-hooks/90-raw-disk-image.binary
@@ -115,6 +115,15 @@ zfs mount rpool/ROOT/ubuntu
 rsync --info=stats3 -Wa binary/* /mnt/
 
 #
+# Set up a swap device.
+#
+zfs create -V 4G -b "$(getconf PAGESIZE)" \
+	-o logbias=throughput -o sync=always \
+	-o primarycache=metadata rpool/swap
+mkswap -f /dev/zvol/rpool/swap
+echo /dev/zvol/rpool/swap none swap defaults 0 0 >>/mnt/etc/fstab
+
+#
 # Now we need to install the bootloader. In order to do that, we'll chroot
 # into the newly populated root filesystem, so that we use the grub-install
 # and update-grub binaries installed in that filesystem.  Additionally, we


tools/ansible/roles/dlpx.appliance-build-bootstrap/tasks/install_zfsutils.yml
```diff
@@ -0,0 +1,10 @@
+#
+# Copyright (c) 2018 by Delphix. All rights reserved.
+#
+
+- name: Install zfsutils | Install dependencies
+  apt:
+    name: "{{ item }}"
+  with_items:
+    - zfsutils-linux
```
basil: I think we also depend on the ZFS kernel modules (`zfs-dkms`?). This dependency is currently implicit, but it would be great to make it explicit in this Ansible role. Of course, this is a pre-existing issue, so I'm fine if you don't want to deal with it in this change.

tools/ansible/roles/dlpx.appliance-build-bootstrap/tasks/main.yml
```diff
@@ -0,0 +1,9 @@
+#
+# Copyright (c) 2018 by Delphix. All rights reserved.
+#
+
+#
+# This role configures a VM for building the Virtualization Engine.
```
basil: Update the comment to read "for building the Delphix Appliance"?

tools/ansible/roles/dlpx.appliance-build-bootstrap/tasks/install_zfsutils.yml
prakashsurya: the kernel module comes by default with the ubuntu kernel package.. thus, we'd have to explicitly depend on the kernel package, which seems like that's a little unnecessary? maybe just a comment is good enough? I could go either way.

scripts/kernel-modules-unload.sh
```diff
@@ -23,4 +23,5 @@ fi
 set -o errexit
 set -o xtrace
 
+systemctl stop zfs-zed.service
```
prakashsurya: why do we need this?

pcd1193182: I explained in the last sentence of the description; in order to unload the zfs kernel modules, the zed service must be stopped first. If it isn't stopped, modprobe -r fails with EBUSY. It wasn't an issue before, because the utils were only installed inside the container, but now that they're installed in the bootstrap VM itself we need to handle this.

prakashsurya: I guess I was more curious why we couldn't unload the kernel module without stopping this service.. like, what does this do to prevent EBUSY? do we know?

tools/ansible/ansible.cfg
```diff
@@ -0,0 +1,6 @@
+#
+# Copyright (c) 2018 by Delphix. All rights reserved.
```
prakashsurya: The copyright on the new files needs to match what we're using elsewhere in this file; to adhere to the Apache 2.0 requirements, which is different than what we use for our internal repositories.

tools/ansible/roles/dlpx.appliance-build-bootstrap/tasks/main.yml
```diff
@@ -0,0 +1,9 @@
...
+#
+---
+- include: install_zfsutils.yml
```
prakashsurya: IMO, we should just remove the `install_zfsutils.yml` file and put those contents directly in this file.. I don't see a need for two files.. but I'm OK with this.

tools/ansible/roles/dlpx.appliance-build-bootstrap/tasks/main.yml
prakashsurya: I'd be in favor of just removing the comment...

tools/ansible/playbooks/bootstrap.yml
```diff
@@ -0,0 +1,9 @@
+#
+# Copyright (c) 2018 by Delphix. All rights reserved.
+#
+---
+- hosts: all
```
prakashsurya: can we add the `become: true` line here? rather than having it in the `apt` invocation? this way, if/when we add more tasks, we don't have to add `become: true` to all of them.

tools/ansible/playbooks/bootstrap.yml
```diff
@@ -0,0 +1,9 @@
...
+  vars:
+    ansible_python_interpreter: /usr/bin/python3
+  roles:
+    - dlpx.appliance-build-bootstrap
```
prakashsurya: I'd rather adopt a name like `appliance-build.bootstrap`, which doesn't have "dlpx" in the name, and is more consistent with the naming of other roles in this repository.

prakashsurya:
Thinking a little more, it seems a little awkward how we now would have multiple ways to configure the bootstrap system. We'd do part of the configuration with scripts (e.g. `docker-build.sh`, `kernel-modules-load.sh`, etc), and then part of the configuration with this new Ansible role.

Personally, since we already do stuff with these scripts, I think it'd be better to add a new script (e.g. `zfsutils-install-ubuntu.sh` to match the already existing `docker-install-ubuntu.sh` convention) that'd be used to install the package. This would be more simple and consistent with other things in this repo, and accomplish the same goal.

Alternatively, I think we could try to remove some of the scripts used to configure the bootstrap system, and instead move that logic into this new Ansible role.

It just seems a little awkward to use Ansible to install the `zfsutils` package, but not for any of the other configuration of the bootstrap system that's also required. The consumer of this new role is already tailored specific to this repository's scripts/steps, so I don't view adding another script as any worse than what we are already doing.

@basil @pcd1193182 What do y'all think?

pcd1193182:
I don't have strong feelings on how we want to install the packages. I prototyped this change with a line in a shell script installing the package, but I felt that the idempotence and industry-standardness of ansible was a benefit. On the other hand, for this fairly limited use case, adding an outside usage instruction (you must invoke the ansible tools before building) is a bit cumbersome. And it does match how we do other things in this repo. I could go either way.

basil:
I don't have strong feelings either, but I mentored Paul to do it this way because it matches the design used for building the Virtualization and Masking packages. Having consistency across various teams and products that use the `bootstrap` VM is a plus, in my opinion. I don't think the existing Docker shell scripts are particularly elegant. I have a mind for folding them into this Ansible role and/or the [Jenkins Docker Pipeline Plugin](https://github.com/jenkinsci/docker-workflow-plugin) eventually.

prakashsurya:
OK, fair enough. Let me see if I can pull the scripts into Ansible as well; with this shift to Ansible, I think it'd be good to try and move the existing configuration into Ansible too.. I'll report back tomorrow.. I don't think it'll be too difficult to move everything into Ansible and then remove the scripts, but need some time to experiment and try it out.

bootstrap/roles/appliance-build.bootstrap/tasks/install_zfs.yml
```diff
@@ -0,0 +1,23 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+- name: Install zfs packages | Install dependencies
+  apt:
+    name: "{{ item }}"
+  with_items:
+    - zfs-dkms
```
prakashsurya: why are we pulling in "zfs-dkms"? I don't think that's necessary (it might actually be harmful).. iirc, the zfs kernel module comes from the ubuntu kernel package, not zfs-dkms.. or am I mistaken?

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
```diff
@@ -29,6 +29,8 @@
     path: "{{ toplevel.stdout }}/docker"
     name: appliance-build
 
+- include: install_zfs.yml
```
prakashsurya: I would prefer just to add "zfsutils-linux" to list of packages we're adding a few lines above.. do you think it's better to add a new file, and include it here?

live-build/misc/live-build-hooks/90-raw-disk-image.binary
```diff
@@ -114,6 +114,15 @@ zfs mount rpool/ROOT/ubuntu
 #
 rsync --info=stats3 -Wa binary/* /mnt/
 
+#
+# Set up a swap device.
+#
+zfs create -V 4G -b $(getconf PAGESIZE) \
```
prakashsurya: These lines are failing the "shfmt" check; see [here](https://travis-ci.com/delphix/appliance-build/jobs/128450236)
```
 zfs create -V 4G -b $(getconf PAGESIZE) \
-    -o logbias=throughput -o sync=always \
-    -o primarycache=metadata rpool/swap
+	-o logbias=throughput -o sync=always \
+	-o primarycache=metadata rpool/swap
 mkswap -f /dev/zvol/rpool/swap
-echo /dev/zvol/rpool/swap none swap defaults 0 0 >> /mnt/etc/fstab
+echo /dev/zvol/rpool/swap none swap defaults 0 0 >>/mnt/etc/fstab
```

If you have a bootstrap system set up, like the README describes, you can run the check like this:
```
./scripts/docker-run make shfmtcheck
```

You can also have shfmt automatically fix the files too, with something like this:
```
shfmt -w $(find scripts -type f -executable) \
    $(find live-build/misc/live-build-hooks -type f -executable)
```

prakashsurya: It looks like "shellcheck" also wants some changes (see [here](https://travis-ci.com/delphix/appliance-build/jobs/128450235))
```
In live-build/misc/live-build-hooks/90-raw-disk-image.binary line 120:
zfs create -V 4G -b $(getconf PAGESIZE) \
                    ^-- SC2046: Quote this to prevent word splitting.
```

bootstrap/roles/appliance-build.bootstrap/tasks/install_zfs.yml
pcd1193182: As you saw above, this was requested by @basil. My understanding is that this will effectively result in a no-op on ubuntu machines since the zfs-dkms package is already installed on all machines. Having it listed here merely makes the dependency explicit. I would be fine including it or not; I was merely attempting to address review feedback.

basil: I think @prakashsurya's point was that this isn't actually the correct package. I was never sure that it was, hence the question mark I used when making the request.

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
pcd1193182: My understanding (admittedly limited) of ansible is that it's best to split logically separate dependencies into separate files and include as necessary. If that's not the convention, or overkill for this small of a dependency list, I'm fine including it in the same file as well.

bootstrap/roles/appliance-build.bootstrap/tasks/install_zfs.yml
pcd1193182: Update: Actually, it looks like in some cases this can cause issues, so I'm going to remove it from the file.

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
prakashsurya: I think it's overkill in this case, since we're just installing a single package, and we already have some logic to do that for other packages above. But I'll defer to your judgement here; if you think a new file is best, that's OK with me.

bootstrap/roles/appliance-build.bootstrap/tasks/install_zfs.yml
prakashsurya: Yea, I think "zfs-dkms" is actually the wrong package.. the correct package would be `linux-modules` I think:
```
$ dpkg-query -S /lib/modules/$(uname -r)/kernel/zfs/zfs/zfs.ko
linux-modules-4.15.0-1009-aws: /lib/modules/4.15.0-1009-aws/kernel/zfs/zfs/zfs.ko
```
But IMO, I don't think we gain a whole lot by explicitly listing that dependency here.

prakashsurya:
bors delegate+

Looks like there’s still a small shfmt issue to address. Add a comment of “bors r+” to land when that’s fixed. Thanks!

PR #42, opened 2018-06-07
Perform bootstrap configuration using Ansible

prakashsurya:


prakashsurya:
@basil @pcd1193182 When you have time, please give this a look. I still need to fully test this, but my idea is, we'd land this change and update the Jenkins job, and then rebase #41 on top of this. This way, we use Ansible to do all the configuration of the bootstrap system, rather than mix and matching Ansible with bash scripts. Also, I didn't adopt the `tools/ansible` directory hierarchy here, but I can do that if it's required.

basil:
Looks great to me! Happy to approve this once it's done.

prakashsurya:
I've tested this manually by running through the process described in the README (that I've updated), as well as via Jenkins (see [here](http://collins.d.delphix.com:29627/job/devops-gate/job/projects/job/dx4linux/job/appliance-build/job/master/job/post-push/2/console)).

PR #43, opened 2018-06-08
LX-1100 appliance-build portion of LX-788

jwk404:
When testing OOM conditions in the stack, it was discovered that the
script which creates the hprof dump could fail to be run because the JVM
uses fork(2) rather than vfork(2). In order to allow the dump creation
to proceed, the default settings for vm.overcommit_memory needed to
change from the default.

Additionally, this change enables application core dumps for all users
and creates a common directory in which all cores are written. Cores
will use the same filename pattern as on illumos.

Tested that an hprof dump is successfully created when the stack
encounters an OOM error using the fixed test_out_of_memory blackbox
test.

Created a live build, and tested that cores are created for root and
user processes in the correct location.

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
@@ -266,3 +266,50 @@
       ## Record command exit failures (execve result is not command result)
       -a exit,always -F a0!=0 -S exit_group
       -a exit,always -F a0!=0 -S exit
+
+#
+# By default, the ulimit for core files is set to 0, and the default
+# filename and location for a core file is 'core' in the cwd. Update
+# limits.conf to allow processes running as root or a regular user to
+# make core files.
+#
+- lineinfile:
+    create: yes
+    dest: /etc/security/limits.conf
+    line: "{{ item }} soft core unlimited"
+  with_items:
+    - 'root'
+    - '*'
+
+#
+# Create a world writeable directory for application core dumps. We
+# want it world writeable because we're sharing one directory for
+# corefiles from any user. Unlike illumos where all cores are written
+# out as root, linux cores are written with the UID of the running
+# process.
+#
+- file:
+    path: /var/crash
+    state: directory
+    mode: 0777
+
+#
+# Update sysctl.conf such that application cores are written into
+# /var/crash and tagged with the execname, process id, and seconds since
+# the epoch.
+#
+# Additionally in this file, we force the kernel to assume memory
+# allocations will succeed until we actually run out of memory. The
+# default heuristic can cause failure to generate an hprof dump on OOM
+# in the stack. Allowing overcommit lets the fork(2) succeed despite not
+# having enough memory which allows the script that generates the hprof
+# dump to run.
+#
+- lineinfile:
+    create: yes
+    dest: /etc/sysctl.conf
+    regexp: "^#?{{ item.key }} ="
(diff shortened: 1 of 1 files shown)


live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -266,3 +266,45 @@
...
       -a exit,always -F a0!=0 -S exit_group
       -a exit,always -F a0!=0 -S exit
+
+#
+# By default, the ulimit for core files is set to 0, and the default
+# filename and location for a core file is 'core' in the cwd. Update
+# limits.conf to allow processes running as root or a regular user to
+# make core files.
+#
+- lineinfile:
```
prakashsurya: this could be simplified:
```
+- lineinfile:
+    create: yes
+    dest: /etc/security/limits.conf
+    line: "{{ item }} soft core unlimited"
+  with_items:
+    - root
+    - *
```
but if you prefer to use `domain`, that's fine with me.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -266,3 +266,45 @@
...
+    create: yes
+    dest: /etc/security/limits.conf
+    line: "{{ item.domain }} soft core unlimited"
+  with_items:
+    - { domain: "root" }
+    - { domain: "*" }
+
+#
+# Create a world writeable directory for application core dumps.
```
prakashsurya: I think it'd be useful to put your analysis in a comment here, about how core dumps are created/written as the user of the process, rather than always using "root" like how illumos does it; which is why we want it to be world writable.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -266,3 +266,45 @@
...
+# the epoch.
+#
+# Additionally in this file, we force the kernel to assume memory
+# allocations will succeed until we actually run out of memory. The
+# default heuristic can cause failure to generate an hprof dump on OOM
+# in the stack. Allowing overcommit lets the fork(2) succeed despite not
+# having enough memory which allows the script that generates the hprof
+# dump to run.
+#
+- lineinfile:
```
prakashsurya: you could add a `regexp` field, so that any default value will get replaced by the new value. I'm OK leaving it out, but that'd make things a little more elegant; e.g. something like this:
```
+- lineinfile:
+    create: yes
+    dest: /etc/sysctl.conf
+    regexp: "^#?{{ item.key }} ="
+    line: "{{ item.key }} = {{ item.value }}"
+  with_items:
+    - { key: 'kernel.core_pattern', value: '/var/crash/core.%e.%p.%t' }
+    - { key: 'vm.overcommit_memory', value: '1' }
```

prakashsurya:
```
ERROR! Syntax Error while loading YAML.
  did not find expected alphabetic or numeric character
The error appears to have been in '/opt/appliance-build/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml': line 282, column 8, but may
be elsewhere in the file depending on the exact syntax problem.
The offending line appears to be:
    - root
    - *
       ^ here
E: config/hooks/80-ansible-playbook.binary failed (exit non-zero). You should check for errors.
```
I guess `*` is special? maybe it's getting parsed as a multiplication operator? idk... I think if you just add back the `"`'s to make sure it's interpreted as a string, that'll fix the errors.

PR #45, opened 2018-06-11
Appliance-build should use packages produced by the ZFS package build job

### LINKED ISSUE #44: Appliance-build should use packages produced by the ZFS package build job
Now that we have a job that produces ZFS packages, we should leverage it instead of relying on `make install`. Note that this depends on the commit below to be pushed first: 
```
TOOL-5821 ZFS packages should be automatically built on new commits
```

pzakha:
With this change, appliance build will use latest ZFS packages instead of checking out ZFS and launching `make install`.

Note that we still run the `appliance-build.zfsonlinux-development` role on the internal-dev variant to allow git-zfs-make to work.

**Required devops change**
http://reviews.delphix.com/r/40866

**Note**
In run-live-build.sh, we are calling:
```export AWS_S3_BUCKET="snapshot-de-images"```
This is not ideal as it forces the same bucket on all s3 packages (virtualization, masking, zfs). Some work is being done by Garret to clean this up.

**Testing**
http://collins.d.delphix.com:32423/job/devops-gate/job/projects/job/dx4linux/job/appliance-build/job/master/job/pre-push/7/console
linux-pre-checkin with an appliance-build made with ZFS packages:
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/blackbox-smoke/36481/consoleFull

Closes #44

### CODE REVIEW

File: live-build/misc/ansible-roles/appliance-build.install-delphix-s3-packages/tasks/main.yml [hunk 1]
```diff
@@ -42,7 +42,8 @@
   register: tempdir
   delegate_to: localhost
 
-- command: aws s3 sync --only-show-errors "s3://{{ bucket }}/{{ prefix }}" .
+- name: Download packages
```
prakashsurya: IMO, I'd leave off the name, to be consistent with the other tasks..

I'd like to make a change that adds the name to all the tasks, and then enables the ansible lint check to ensure we always name tasks moving forward.. but until all other tasks are updated, I vote to be consistent, and not add a name here (and below).

live-build/misc/ansible-roles/appliance-build.install-delphix-s3-packages/tasks/main.yml
```diff
@@ -56,10 +57,15 @@
   args:
     chdir: "{{ tempdir.path }}"
 
-- apt:
-    deb: "{{ item }}"
-  with_fileglob:
-    - "{{ tempdir.path }}/*.deb"
+#
+# We use apt-get here instead of ansible's apt module to install all the debs
+# at the same time, so that any dependencies between provided packages are
```
prakashsurya: Can you try to expand on this comment a bit, and more clearly define what "dependencies between provided packages" means? perhaps with an example.. I think I understand, but having an example would probably make things more clear, especially months later when I forget how this works..

live-build/misc/ansible-roles/appliance-build.install-delphix-s3-packages/tasks/main.yml
```diff
@@ -56,10 +57,15 @@
...
+# resolved. We use 'shell' instead of 'command' to resolve the glob (*.deb).
+#
+- name: Install packages
+  shell: "apt-get install -y ./*.deb"
```
prakashsurya: We might also want to set the environment variable: `DEBIAN_FRONTEND=noninteractive ` .. and use the apt-get options: `-o DPkg::Options::=--force-confold -o DPkg::Options::=--force-confdef`

pzakha:
I've ran `debsums -c` and it came back clean.
Added Prakash's suggestions in latest update. New build:
http://collins.d.delphix.com:32423/job/devops-gate/job/projects/job/dx4linux/job/appliance-build/job/master/job/pre-push/8/console

pzakha:
Latest version: http://collins.d.delphix.com:32423/job/devops-gate/job/projects/job/dx4linux/job/appliance-build/job/master/job/pre-push/10/consoleFull (passed)

PR #47, opened 2018-06-12
LX-1108 swap zvol formatting sometimes fails

pcd1193182:
This retry loop will help is handle the case where the `/dev/zvol` device takes a few seconds to be created. It will delay for a maximum of 20 seconds, and will exit if the device doesn't appear.

### DIFF
--- a/live-build/misc/live-build-hooks/90-raw-disk-image.binary
+++ b/live-build/misc/live-build-hooks/90-raw-disk-image.binary
@@ -22,6 +22,11 @@
 # contents are that of the "binary" directory).
 #
 
+die() {
+	echo "$*" 1>&2
+	exit 1
+}
+
 #
 # When running in Travis CI, we cannot expect this script to be running
 # on a host that has the ZFS kernel modules installed *and* for those
@@ -45,8 +50,7 @@ external-*)
 	RAW_DISK_SIZE_GB=127
 	;;
 *)
-	echo "Invalid variant specified: '$APPLIANCE_VARIANT'" 1>&2
-	exit 1
+	die "Invalid variant specified: '$APPLIANCE_VARIANT'"
 	;;
 esac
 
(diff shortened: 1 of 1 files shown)


live-build/misc/live-build-hooks/90-raw-disk-image.binary
```diff
@@ -120,6 +120,21 @@ rsync --info=stats3 -Wa binary/* /mnt/
...
+# We need to wait for the /dev/zvol device to appear before we proceed; this
+# can take a few seconds, so we retry until it succeeds.
+#
+tries=0
+while [[ ! -e /dev/zvol/rpool/swap && $tries -lt 5 ]]; do
+	sleep 5
+	tries=$((tries + 1))
+done
+if [[ ! -e /dev/zvol/rpool/swap }]; then
+	echo "/dev/zvol device failed to appear before timeout"
```
jgallag88: [nit] It would be nice to add a `die` function

live-build/misc/live-build-hooks/90-raw-disk-image.binary
```diff
@@ -120,6 +120,21 @@ rsync --info=stats3 -Wa binary/* /mnt/
...
+#
+# We need to wait for the /dev/zvol device to appear before we proceed; this
+# can take a few seconds, so we retry until it succeeds.
+#
+tries=0
+while [[ ! -e /dev/zvol/rpool/swap && $tries -lt 5 ]]; do
+	sleep 5
+	tries=$((tries + 1))
+done
+if [[ ! -e /dev/zvol/rpool/swap }]; then
```
prakashsurya: I think you need to replace `}` with `]`

live-build/misc/live-build-hooks/90-raw-disk-image.binary
```diff
@@ -120,6 +120,21 @@ rsync --info=stats3 -Wa binary/* /mnt/
 zfs create -V 4G -b "$(getconf PAGESIZE)" \
 	-o logbias=throughput -o sync=always \
 	-o primarycache=metadata rpool/swap
+
+#
+# We need to wait for the /dev/zvol device to appear before we proceed; this
+# can take a few seconds, so we retry until it succeeds.
+#
+tries=0
+while [[ ! -e /dev/zvol/rpool/swap && $tries -lt 5 ]]; do
```
prakashsurya: Are we sure that waiting a maximum of 25 seconds is enough time to guarantee the device will appear when everything is working correctly?

I'd rather wait a long time and have it eventually work, than have to deal with flaky failures due to not waiting long enough. If 25 seconds is sufficient, I'm good with this; if we're not sure, I'd suggest bumping this to something on the order of minutes (I assume minutes is long enough, since we're starting with seconds here).

jgallag88: When I was looking at the same issue in the app-gate, I think I found that the device typically appeared within a couple ms, and the outliers were still less <100ms. I think I ended up setting the timeout there to 20s.

prakashsurya: Perfect, thanks! Based on that, I'm good with this.

prakashsurya:
@pcd1193182 you'll need to fix the test failure and retry.

PR #49, opened 2018-06-13
Remove PostgreSQL repository from APT sources

prakashsurya:
Since moving to PostgreSQL 10, we no longer need to install PostgreSQL
9.4 on the appliance, so we don't need this 3rd party APT source line.


prakashsurya:
`git-ab-pre-push` is running [here](http://ops.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build/job/master/job/pre-push/40/)

PR #53, opened 2018-06-14
Add "qemu" package to bootstrap ansible role

prakashsurya:
This change adds the "qemu" package installation into the ansible role
for the bootstrap system. As a result, we can remove the commands from
the README.md that instruct the user to explicitly install this package.


prakashsurya:
`git-ab-pre-push` is [here](http://ops.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build/job/master/job/pre-push/44/)

PR #54, opened 2018-06-14
Allow checking format of scripts locally with: "make checks"

pzakha:
I've adapted the Makefile's find commands to also work properly under OSX.
We can now run all the checks locally by executing `make checks` without needing a docker container
The prerequisites on OSX is to install shellcheck, shfmt and ansible-lint. This can be done easily with brew:
```
brew install shellcheck shfmt ansible-lint
```
Note that I've also added a small change to `92-ova-machine-image.binary` to make shfmt pass.

**Testing**
I've tested `make checks` on OSX and tested `./scripts/docker-run.sh make checks` on Linux.
I've also confirmed that the format change in `92-ova-machine-image.binary` doesn't affect the contents of the "$OVA_DIRECTORY/$APPLIANCE_VARIANT.mf" file created.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,10 @@ ALL_VARIANTS = $(shell find live-build/variants -maxdepth 1 -mindepth 1 -exec ba
 ALL_INTERNAL = $(shell find live-build/variants -maxdepth 1 -mindepth 1 -name 'internal-*' -exec basename {} \;)
 ALL_EXTERNAL = $(shell find live-build/variants -maxdepth 1 -mindepth 1 -name 'external-*' -exec basename {} \;)
 
+FINDEXEC.Darwin := -perm +111
+FINDEXEC.Linux := -executable
+FINDEXEC := $(FINDEXEC.$(shell uname -s))
+
 all-variants: $(ALL_VARIANTS)
 all-internal: $(ALL_INTERNAL)
 all-external: $(ALL_EXTERNAL)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -61,8 +65,10 @@ shellcheck:
...
 shfmtcheck:
-	! shfmt -d $$(find scripts -type f -executable) \
-		$$(find live-build/misc/live-build-hooks -type f -executable) | grep .
+	! shfmt -d $$(find scripts -type f $(FINDEXEC)) \
+		$$(find live-build/misc/live-build-hooks -type f $(FINDEXEC)) | grep .
 
 ansiblecheck:
 	ansible-lint $$(find bootstrap live-build/variants -name playbook.yml)
+
+checks: shellcheck shfmtcheck ansiblecheck
```
prakashsurya: I think the normal convention is to use `make check` rather than `make checks`..

alternatively, I see ZOL has a `make checkstyle` target; I'd be OK using that to be consistent with another codebase we'll frequently be using.

live-build/misc/live-build-hooks/92-ova-machine-image.binary
```diff
@@ -74,7 +74,7 @@ find "$OVA_DIRECTORY" -type f | while read -r file; do
 	NAME=$(basename "$file")
 	HASH=$(sha1sum "$file" | awk '{print $1}')
 	cat <<-EOF >>"$OVA_DIRECTORY/$APPLIANCE_VARIANT.mf"
-	SHA1($NAME)= $HASH
+		SHA1($NAME)= $HASH
```
prakashsurya: Is this change intentional?

prakashsurya: Do you know why this wasn't failing the automated checks? We don't run shfmt against this file via the automated PR checks?

pzakha: It's an issue with shfmt version: v2.30 on our docker image vs v.2.40 on OSX through brew.
I'll update the docker image to use 2.40

prakashsurya: Ah, OK. Thanks, that makes sense.

PR #55, opened 2018-06-14
LX-1114 Missing utilities causing zfstest failures

jwk404:
The utilities `bc` `attr` and `setfattr` are missing from the build,
in turn causing 28 tests in the suite to fail. After this change,
those tests passed.

### DIFF
--- a/live-build/base/config/package-lists/minimal.list.chroot
+++ b/live-build/base/config/package-lists/minimal.list.chroot
@@ -43,3 +43,10 @@ open-vm-tools
 openssh-server
 ubuntu-minimal
 auditd
+
+#
+# The following packages are required for the ZFS test suite to run
+# properly.
+#
+attr
+bc


live-build/base/config/package-lists/minimal.list.chroot
```diff
@@ -43,3 +43,10 @@ open-vm-tools
 openssh-server
 ubuntu-minimal
 auditd
+
+#
+# The following packages are required for the ZFS test suite to run
```
prakashsurya: wouldn't the correct fix, be to list these as dependencies of the zfstest package? if we did that, they'd get automatically pulled in when we install the zfstest package.

If we did it this way, and this list ever changed because the zfstest requirements changed, we'd run into a flag day scenario which I'd like to avoid if possible.

jwk404: ok, I'll do that instead.

PR #56, opened 2018-06-15
Label make targets as .PHONY

prakashsurya:
All of our current make targets are actually "phony targets", since
they're not intended to result in a file of that same target name. See
also: https://www.gnu.org/software/make/manual/make.html#Phony-Targets

This is a simple change to list all of our make targets as ".PHONY".


prakashsurya:
`git-ab-pre-push` is [here](http://ops.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build/job/master/job/pre-push/50/)

PR #57, opened 2018-06-15
Add initial support for building metapackages

prakashsurya:
This change adds some initial support for building a new metapackage
named "delphix-foundation". The intention is for this package to depend
on other packages, such that this list of dependencies builds the
"foundation" of the appliance. The current list of dependencies is
defined in the "metapackages/seeds/bionic/foundation" file.

This new package isn't yet consumed by anything, but the plan is to
migrate the live-build base configuration over to using this package;
i.e. the files in "live-build/base/config/package-lists" would be
replaced with a single file that only listed "delphix-foundation".

The new package can be easily generated via "make metapackages":

    $ ./scripts/docker-run.sh make metapackages

    $ ls -l *.deb
    -rw-r--r-- 1 root root 1.3K Jun 15 22:07 delphix-foundation_1.0.0_amd64.deb

    $ dpkg-deb -I delphix-foundation_1.0.0_amd64.deb
     new Debian package, version 2.0.
     size 1252 bytes: control archive=628 bytes.
         531 bytes,    10 lines      control
          80 bytes,     1 lines      md5sums
     Package: delphix-foundation
     Source: delphix-meta
     Version: 1.0.0
     Architecture: amd64
     Maintainer: Delphix Engineering <eng@delphix.com>
     Installed-Size: 7
     Depends: atop, auditd, bpfcc-tools, cloud-init, debsums, dnsutils, dstat, emacs, gdb, glances, grub-pc, iftop, inotify-tools, linux-generic, linux-tools-generic, net-tools, open-vm-tools, openssh-server, procinfo, sg3-utils, strace, tshark, ubuntu-minimal, vim, zfs-initramfs, zfsutils-linux
     Section: metapackages
     Priority: optional
     Description: Foundation for Delphix Appliance

In addition to the ".deb" file, these other files are also generated:

    $ ls -l delphix-*
    -rw-r--r-- 1 root root 1.3K Jun 15 22:07 delphix-foundation_1.0.0_amd64.deb
    -rw-r--r-- 1 root root  575 Jun 15 22:07 delphix-meta_1.0.0.dsc
    -rw-r--r-- 1 root root 1.9K Jun 15 22:07 delphix-meta_1.0.0.tar.xz
    -rw-r--r-- 1 root root 5.6K Jun 15 22:07 delphix-meta_1.0.0_amd64.buildinfo
    -rw-r--r-- 1 root root 1.6K Jun 15 22:07 delphix-meta_1.0.0_amd64.changes


prakashsurya:
`git-ab-pre-push` is [here](http://ops.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build/job/master/job/pre-push/51/)

.travis.yml
```diff
@@ -13,6 +13,7 @@ services:
 env:
   - TARGET=ansiblecheck
   - TARGET=internal-minimal
+  - TARGET=metapackages
```
pzakha: Do you think this is something we need to do on every build for now?

prakashsurya: Is there a good reason not to do it for every build?

pzakha: I don't mind either way since it's done in parallel and should not be too flaky. I was just wondering what was the motivation of running it by travis, since we do not consume build products from travis.

prakashsurya: Without this, it's basically dead code as there's no other mechanism to ensure it continues to work moving forward. Further, since we don't have a way to kick of specific tests based on the changes that are being made, it's either we test on all changes (even diffs that don't are irrelevant to the test being performed), or rely on reviews to enforce manual testing (which usually means tests don't get run).

IMO, more testing is better than less testing, so long as adding more tests doesn't negatively impact the test results (e.g. introduce flakey tests), or negatively impact the test latency (e.g. add long running tests for little benefit). In this case, I don't think this additional test does either, so it's a prime candidate for running on each PR. This is similar to how I chose to run the style checks (shfmt, shellcheck, ansiblelint) for each PR, even though those tests may be irrelevant for some changes.

Relying on peer reviews to enforce (and communicate) which tests need to be run for any given change is a poor system, imo, so I prefer to avoid that whenever possible; even if it means sometimes/often running irrelevant tests (that ultimately pass, of course).

PR #60, opened 2018-06-19
Allow specifying different buckets and prefixes for each s3 component

pzakha:
Right now, we cannot specify different buckets for the various delphix components fetched from s3. 
This change simplifies what is being passed down to live-build: instead of having separate prefix and bucket components, we pass the s3 url directly.
Note that the jenkins job can still pass us a prefix instead of a full URL, so we make sure that we can still handle that case.

### Testing
Custom ZFS s3 URL: http://ops.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build/job/master/job/pre-push/60/parameters/ (pass)
git-ab-pre-push: http://ops.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build/job/master/job/pre-push/59/ (pass)

### DIFF
--- a/scripts/run-live-build.sh
+++ b/scripts/run-live-build.sh
@@ -61,6 +61,43 @@ else
 	cd "$TOP/live-build/variants/$1"
 fi
 
+function get_s3_uri() {
+
+	local def_bucket="snapshot-de-images"
+	local jenkinsid="jenkins-ops"
+
+	local prefix="$1"
+	local latest="s3://$def_bucket/builds/$jenkinsid/$2"
+	local uri
+
+	if [[ -z "$prefix" ]]; then
+		#
+		# prefix is empty, so get latest version
+		#
+		aws s3 cp --quiet "$latest" .
+		prefix=$(cat latest)
+		uri="s3://$def_bucket/$prefix"
+		rm -f latest
+	elif [[ "$prefix" == s3* ]]; then
+		#
+		# prefix was set to full s3 URI: "s3://<bucket>/<prefix>"
+		#
+		uri="$prefix"
+	else
+		#
+		# We assume prefix is inside default bucket
+		#
+		uri="s3://$def_bucket/$prefix"
+	fi
+
+	if aws s3 ls "$uri" >/dev/null; then
+		echo "$uri"
+	else
+		echo "$uri not found." 1>&2
+		exit 1
+	fi
+}
+
 #
 # When performing minimal testing from within Travis CI, we won't have
 # access to the Delphix internal infrastructure. Thus, we want to skip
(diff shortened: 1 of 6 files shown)


scripts/run-live-build.sh
```diff
@@ -85,41 +122,18 @@ if ! [[ -n "$CI" && -n "$TRAVIS" ]]; then
...
-		URI="s3://$BUCKET/builds/$JENKINSID/dlpx-app-gate/"
-		URI+="projects/dx4linux/build-package/post-push/latest"
-
-		aws s3 cp "$URI" .
-		AWS_S3_PREFIX_VIRTUALIZATION=$(cat latest)
-		export AWS_S3_PREFIX_VIRTUALIZATION
-		export AWS_S3_BUCKET="snapshot-de-images"
-		rm -f latest
-	fi
+	AWS_S3_URI_VIRTUALIZATION=$(get_s3_uri "$AWS_S3_PREFIX_VIRTUALIZATION" \
```
prakashsurya: I don't really like how we're now using a variable named "prefix" to potentially specify the full URI, not a prefix.

I think I'd rather do something like:
```
if [[ -z "$AWS_S3_URI_VIRTUALIZATION" ]]; then
    export AWS_S3_URI_VIRTUALIZATION=$(...)
fi
```

Then we expose the `AWS_S3_URI_VIRTUALIZATION` variable, and expect consumers to actually populate it with a URI.. rather than mixing the terms "prefix" and "URI", which can be confusing for the reader and consumers of this script.

scripts/run-live-build.sh
```diff
@@ -61,6 +61,43 @@ else
 	cd "$TOP/live-build/variants/$1"
 fi
 
+function get_s3_uri() {
+
+	local def_bucket="snapshot-de-images"
+	local jenkinsid="jenkins-ops"
+
+	local prefix="$1"
+	local latest="s3://$def_bucket/builds/$jenkinsid/$2"
```
prakashsurya: if we're going to "bake in" parts of the URI like "builds" here, should we also do that with "post-push/latest"?

scripts/run-live-build.sh
pzakha: Sorry, I forgot to mention that devops is working to refactor the jenkins job.
The goal is indeed to rename PREFIX to URI once that is done. I think for now we should be able to live with prefix accepting URIs (note that prefixes are still accepted).

In my first implementation I actually did create AWS_S3_BUCKET_VIRTUALIZATION and AWS_S3_BUCKET_MASKING but realized that is not useful, as we only need the full URI, and it simplifies things.

scripts/run-live-build.sh
pzakha: We could do that.

PR #62, opened 2018-06-22
LX-1092 create dcenter build variant

grwilson:
This adds a new dcenter variant that is based off the internal-dev variant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,3 +149,74 @@ console and perform any post-boot verification that's required (e.g.
 verify certain packages are installed, etc).
 
 To exit "qemu", one can use "Ctrl-A X".
+
+## Creating new build variants
+
+This repository contains different build variants which are used to
+generate customized images, leveraging different Ansible roles to customize
+those variants. The following instructions can be used to create a new
+build variant and roles.
+
+### Step 1. Create the build variant
+
+The different variants are located in the "live-build/variants" directory:
+
+    $ ls -l live-build/variants
+    total 0
+    drwxr-xr-x  5 root  root  160 Jun  4 22:56 external-standard
+    drwxr-xr-x  5 root  root  160 Jun  4 22:56 internal-dev
+    drwxr-xr-x  5 root  root  160 Jun  4 22:56 internal-minimal
+    drwxr-xr-x  5 root  root  160 Jun  4 22:56 internal-qa
+
+To create a new variant, run the "create-variant.sh" script. Note that
+variant names use the "internal" and "external" prefixes:
+
+    $ ./scripts/create-variant.sh internal-dcenter
+
+### Step 2. Create and populate the ansible role directory
+
+The new variant that was just created may require some customizations applied
+to it, that are specific to this new variant. In order to apply these
+customization, a new Ansible role can be created and used. All existing Ansible
+roles that are used to customize existing build variants are located in the
+"live-build/misc/ansible-roles" directory, and follow the standard Ansible role
+directory structure (i.e. tasks, handlers, files, etc.)
+
+To create a new Ansible role, a new Ansible role directory will need to be
+created. In this example, the name of our Ansible role will be
+"appliance-build.dcenter" to match the name of our variant:
+
+    $ mkdir -p live-build/misc/ansible-roles/appliance-build.dcenter/tasks
+
+There are many roles that exist in "live-build/misc/ansible-roles" which can
+serve as good examples.
+
+### Step 3. Update the variant configuration
+
+With the role[s] created, you can populate the "playbook.yml" file in the
+new variant's "ansible" directory:
+
+    $ ls -l live-build/variants/internal-dcenter/ansible
+    total 8
+    -rw-r--r--  1 root  root  891 Jun 29 14:43 playbook.yml
+    lrwxr-xr-x  1 root  root   27 Jun 19 19:17 roles -> ../../../misc/ansible-roles
+
+For this example, we add our new role to the playboodk as shown below:
+
+    ---
+    - hosts: all
+      connection: chroot
+      gather_facts: no
+      vars:
+        ansible_python_interpreter: /usr/bin/python3
+      roles:
+        - appliance-build.minimal-common
+        - appliance-build.minimal-internal
+        - appliance-build.dcenter
+
+### Step 4. Build the variant
+
+See the instructions [above](#step-4-run-live-build) to setup your build
+environment and kick off the build:
+
+    $ ./scripts/docker-run.sh make internal-dcenter
--- a/live-build/base/config/package-lists/minimal.list.chroot
+++ b/live-build/base/config/package-lists/minimal.list.chroot
@@ -35,10 +35,10 @@ delphix-zfs
 # The following packages provide the minimal set of packages that are
 # necessary to create a working system (e.g. networking, users, etc).
 #
+auditd
 cloud-init
 debsums
 net-tools
 open-vm-tools
 openssh-server
 ubuntu-minimal
-auditd
--- a/live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
@@ -1,0 +1,90 @@
+#
+# Copyright 2018 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+---
+
+- apt:
+    name: "{{ item }}"
+    state: present
+  with_items:
+    - bind9
+    - dnsutils
+    - git
+    - isc-dhcp-server
+    - libldap2-dev
+    - libsasl2-dev
+    - nfs-kernel-server
+    - python-ldap
+    - python-paramiko
+    - python-pip
+    - python-requests
+    - python2.7
+    - telnet
+  register: result
+  until: result is not failed
+  retries: 3
+  delay: 60
+
+- pip:
+    name: pysphere
+    extra_args: --index-url http://artifactory.delphix.com/artifactory/api/pypi/delphix-virtual-pypi/simple/ --trusted-host artifactory.delphix.com --no-cache-dir
+    version: 0.1.8
+    executable: pip2
+
+- git:
+    repo: 'https://gitlab.delphix.com/devops/dcenter-gate.git'
+    version: projects/dx4linux
+    dest: /opt/dcenter/lib/dcenter-gate
+    accept_hostkey: yes
+    update: no
+
+- copy:
+    dest: "/etc/dcenter.conf"
+    content: |
+      [[ -z "$DC_ROOT" ]] && export DC_ROOT="dcenter"
+
+#
+# By default, ubuntu restricts directories where dhcpd and named
+# can operate. For dcenter, we maintain the dhcp configuration
+# in /tmp so we need to add those paths into the apparmor configuration.
+#
+- copy:
+    dest: "/etc/apparmor.d/local/usr.sbin.dhcpd"
+    content: |
+      /tmp/dcenter_dhcp_config/ r,
+      /tmp/dcenter_dhcp_config/** r,
+      /tmp/dcenter_dhcp_config/dhcpd{,6}.conf r,
+      /tmp/dcenter_dhcp_config/dhcpd{,6}_ldap.conf r,
+      /tmp/dcenter_dhcp_config/dhcpd{,6}.leases* lrw,
+      /tmp/dcenter_dhcp_config/dhcpd{,6}.pid rw,
+
+- copy:
+    dest: "/etc/apparmor.d/local/usr.sbin.named"
+    content: |
+      /tmp/dcenter_dhcp_config/named.conf r,
+      /tmp/dcenter_dhcp_config/named.zone r,
+      /tmp/dcenter_dhcp_config/named.pid w,
+
+#
+# Dcenter systems use static addresses so modify cloud.cfg to preserve
+# their hostname.
+#
+- lineinfile:
+    path: /etc/cloud/cloud.cfg
+    regexp: '^preserve_hostname: false'
+    line: 'preserve_hostname: true'
+
+- command: systemctl disable bind9.service
--- a/live-build/variants/internal-dcenter/ansible/playbook.yml
+++ b/live-build/variants/internal-dcenter/ansible/playbook.yml
@@ -1,0 +1,27 @@
+#
+# Copyright 2018 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+---
+- hosts: all
+  connection: chroot
+  gather_facts: no
+  vars:
+    ansible_python_interpreter: /usr/bin/python3
+  roles:
+    - appliance-build.minimal-common
+    - appliance-build.minimal-internal
+    - appliance-build.delphix-ldap
+    - appliance-build.dcenter
--- a/scripts/create-variant.sh
+++ b/scripts/create-variant.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+#
+# Copyright 2018 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#
+# Default variant's playbook to use when creating the new variant
+#
+EXAMPLE_VARIANT=internal-minimal
+
+TOP=$(git rev-parse --show-toplevel 2>/dev/null)
+
+if [[ -z "$TOP" ]]; then
+	echo "Must be run inside the git repsitory."
+	exit 1
+fi
+
+if [[ $# -ne 1 ]]; then
+	echo "Must specify a variant name to create." 1>&2
+	exit 1
+fi
+
+if [[ "$1" == "internal-"[a-z]* || "$1" == "external-"[a-z]* ]]; then
+
+	if [[ -d "$TOP/live-build/variants/$1" ]]; then
+		echo "variant specified ($1) already exists" 1>&2
+		exit 1
+	fi
+else
+	echo "Invalid variant naming convention: $1" 1>&2
+	echo "Variants must start with either 'internal-<name>'" \
+		"or 'external-<name>'" 1>&2
+	exit 1
+fi
+
+set -o errexit
+set -o xtrace
+
+mkdir -p "$TOP/live-build/variants/$1"
+mkdir -p "$TOP/live-build/variants/$1/ansible"
+mkdir -p "$TOP/live-build/variants/$1/config"
+
+ln -s ../../../misc/ansible-roles "$TOP/live-build/variants/$1/ansible/roles"
+cp "$TOP/live-build/variants/$EXAMPLE_VARIANT/.gitignore" \
+	"$TOP/live-build/variants/$1/"
+cp "$TOP/live-build/variants/$EXAMPLE_VARIANT/ansible/playbook.yml" \
+	"$TOP/live-build/variants/$1/ansible"
+
+ln -s ../../../misc/live-build-hooks \
+	"$TOP/live-build/variants/$1/config/hooks"
--- a/live-build/misc/ansible-roles/appliance-build.delphix-ldap/files/etc/idmapd.conf
+++ b/live-build/misc/ansible-roles/appliance-build.delphix-ldap/files/etc/idmapd.conf
@@ -1,0 +1,27 @@
+#
+# Copyright 2018 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+[General]
+Verbosity = 1
+Pipefs-Directory = /run/rpc_pipefs
+Domain = delphix.com
+
+[Mapping]
+Nobody-User = nobody
+Nobody-Group = nogroup
+
+[Translation]
+Method = nsswitch
--- a/live-build/misc/ansible-roles/appliance-build.delphix-ldap/files/etc/ldap.conf
+++ b/live-build/misc/ansible-roles/appliance-build.delphix-ldap/files/etc/ldap.conf
@@ -1,0 +1,21 @@
+#
+# Copyright 2018 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+base dc=delphix,dc=com
+uri ldap://ldap.delphix.com
+ldap_version 3
+rootbinddn cn=admin,dc=delphix,dc=com
+pam_password md5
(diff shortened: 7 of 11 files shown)


live-build/variants/internal-dcenter/ansible/playbook.yml
```diff
@@ -0,0 +1,35 @@
...
+- hosts: all
+  connection: chroot
+  gather_facts: no
+  vars:
+    ansible_python_interpreter: /usr/bin/python2
+  roles:
+    - appliance-build.minimal-common
+    - appliance-build.minimal-internal
+    - appliance-build.install-delphix-zfs-packages
+    - appliance-build.zfsonlinux-development
```
pzakha: We don't need this role

live-build/variants/internal-dcenter/ansible/playbook.yml
```diff
@@ -0,0 +1,35 @@
...
+    - appliance-build.masking-common
+    - appliance-build.masking-development
+    - appliance-build.virtualization-common
+    - appliance-build.virtualization-internal
+    - appliance-build.virtualization-standard
+    - appliance-build.virtualization-development
+    - appliance-build.zfsonlinux-development
```
pzakha: Duplicate. (we also don't need it)
Regarding the various other development roles, de we really need them?

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
```diff
@@ -0,0 +1,99 @@
...
+    name: "{{ item }}"
+    state: present
+  with_items:
+    - python-pip
+    - libldap2-dev
+    - libsasl2-dev
+    - isc-dhcp-server
+    - bind9
+    - dnsutils
+    - telnet
```
pzakha: I'm curious, what needs telnet?

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
```diff
@@ -0,0 +1,99 @@
...
+      [[ -z "$DC_ROOT" ]] && export DC_ROOT="dcenter"
+
+- copy:
+    dest: "/etc/apparmor.d/local/usr.sbin.dhcpd"
+    content: |
+      /tmp/dcenter_dhcp_config/ r,
+      /tmp/dcenter_dhcp_config/** r,
+      /tmp/dcenter_dhcp_config/dhcpd{,6}.conf r,
+      /tmp/dcenter_dhcp_config/dhcpd{,6}_ldap.conf r,
+
```
pzakha: Formatting looks off here

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
```diff
@@ -0,0 +1,99 @@
...
+        version: projects/dx4linux,
+        dest: dcenter-gate }
+
+- copy:
+    dest: "/etc/dcenter.conf"
+    content: |
+      [[ -z "$DC_ROOT" ]] && export DC_ROOT="dcenter"
+
+- copy:
+    dest: "/etc/apparmor.d/local/usr.sbin.dhcpd"
```
pzakha: Why do we need to play with apparmor? (Could you also add some comments)

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
```diff
@@ -0,0 +1,99 @@
...
+    content: |
+      /tmp/dcenter_dhcp_config/named.conf r,
+      /tmp/dcenter_dhcp_config/named.zone r,
+      /tmp/dcenter_dhcp_config/named.pid w,
+
+- file:
+    path: /etc/resolv.conf
+    state: absent
+
+- lineinfile:
```
pzakha: could you add some comments?

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
prakashsurya: iirc, "dc console" telnets from the dcenter host to the esx host.

pzakha: Ah right, I forgot about that one!

live-build/base/config/package-lists/minimal.list.chroot
```diff
@@ -43,3 +43,4 @@ open-vm-tools
 openssh-server
 ubuntu-minimal
 auditd
+python2.7
```
prakashsurya: can you sort this alphabetically, please. "auditd" should be sorted also, but I guess that was missed when it was added.

basil: I don't feel strongly about this, but I'm mildly concerned about adding Python 2.7 to all images given that [Python 2.7 will not be maintained in 1 year and 6 months from now](https://pythonclock.org/). If the Python 2.7 package is installed in all variants, it will be easier for people to start depending on it by accident, which will make it more difficult to rip out these dependencies in a release or two when Python 2.7 is EOL'd. I have a slight preference for limiting `python2.7` to the DCenter variant (if possible), though I don't feel strongly about this.

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
```diff
@@ -0,0 +1,99 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+---
+- include: ldap.yml
```
prakashsurya: can you move the ldap configuration into a new role, e.g. "appliance-build.delphix-ldap", so that we can then use it to configure ldap on our internal delphix engines too?

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
```diff
@@ -0,0 +1,99 @@
...
+
+- apt:
+    name: "{{ item }}"
+    state: present
+  with_items:
```
prakashsurya: unless there's a reason not to, can you sort this list alphabetically?

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
basil: LDAP support is something that may be generally useful to have in the `dev` variant. We currently have it in the `dlpx-trunk` images based on illumos. I don't know that we've intentionally decided _not_ to have it in the Linux `dev` variant, just that it hasn't been implemented yet. I'd suggest the team come to a decision about whether or not we want to include LDAP support in the `dev` variant. Key to this discussion is the question of whether we can grant `sudo` permissions to LDAP users while not granting it to non-LDAP users (something that wasn't possible in illumos). LDAP support without `sudo` support is relatively useless for end users, but giving all users `sudo` access is dangerous in that it is an unsupported product configuration (though the risk here might be somewhat mitigated by the existence of the new `qa` variant). If we decide that we do want LDAP support in the `dev` variant, the LDAP logic probably be better split out into a separate Ansible role rather than being part of the DCenter Ansible role.

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
```diff
@@ -0,0 +1,99 @@
...
+    - telnet
+  register: result
+  until: result is not failed
+  retries: 3
+  delay: 60
+
+
+- pip:
+    name: pysphere
+    extra_args: --trusted-host devpi.delphix.com -i http://devpi.delphix.com:3141/root/public/+simple/
```
basil: `devpi.delphix.com` is deprecated in favor of `artifactory.delphix.com`. We're trying to shut down `devpi`. To use Artifactory, the syntax would be something like this:

```
--index-url http://artifactory.delphix.com/artifactory/api/pypi/delphix-virtual-pypi/simple/ --trusted-host artifactory.delphix.com --no-cache-dir
```

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/ldap.yml
```diff
@@ -0,0 +1,70 @@
...
+# limitations under the License.
+#
+
+---
+
+# Adapted from:
+#   * https://help.ubuntu.com/community/AutofsLDAP
+#   * https://www.digitalocean.com/community/tutorials/how-to-authenticate-client-computers-using-ldap-on-an-ubuntu-12-04-vps
+
+- name: Ubuntu Linux LDAP Access | Install LDAP OS Packages
```
prakashsurya: please remove the "name" from here, and below tasks to be consistent with our other roles/tasks.

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/ldap.yml
```diff
@@ -0,0 +1,70 @@
...
+  apt:
+    name={{ item }}
```
prakashsurya: please reformat module fields here and below to use `:` instead of `=`.. e.g.
```
apt:
  name: "{{ item }}"
  state: present
```
this way we have a consistent style across all the ansible roles.

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/ldap.yml
```diff
@@ -0,0 +1,70 @@
...
+    line="{{ item.line }}"
+  with_items:
+    - { regexp: "^automount:.*$", line: "automount:\tfiles ldap" }
+    - { regexp: "^group:.*$", line: "group:\tcompat ldap" }
+    - { regexp: "^passwd:.*$", line: "passwd:\tcompat ldap" }
+    - { regexp: "^shadow:.*$", line: "shadow:\tcompat ldap" }
+
+- name: Ubuntu Linux LDAP Access | Ensure ldap-home exists
+  file:
+    path="{{ dlpx_ldap_home_path | default('/home') }}"
```
prakashsurya: I don't see "dlpx_ldap_home_path" configured with a value anywhere, so let's remove this reference (and below) since it'll always end up using the default value anyways.

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
```diff
@@ -0,0 +1,99 @@
...
+    version: 0.1.8
+
+- pip:
+    name: "{{ item }}"
+    state: present
+  with_items:
+    - paramiko
```
basil: For my own education, is there any particular reason we're installing these dependencies with `pip` rather than the Ubuntu packages `python-paramiko`, `python-ldap`, and `python-requests`? I am not familiar enough with the advantages and disadvantages of each choice to know what is best (especially with regard to upgrade), but I want to make sure we've thought out the long-term consequences of each approach. One disadvantage of the current approach is that the build will fail if PyPi is down (and this has happened before), though this particular disadvantage can be removed by using the abovementioned Artifactory PyPi cache.

live-build/variants/internal-dcenter/ansible/playbook.yml
```diff
@@ -0,0 +1,35 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+---
+- hosts: all
+  connection: chroot
+  gather_facts: no
+  vars:
+    ansible_python_interpreter: /usr/bin/python2
```
prakashsurya: I think it'd be worth a comment somewhere, stating that dcenter requires python2, and we use python2 to run ansible so that when it runs "pip" it will install the python 2 version of those modules.

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
```diff
@@ -0,0 +1,99 @@
...
+  delay: 60
+
+- git:
+    repo: "{{ item.repo }}"
+    dest:
+      "/opt/dcenter/lib/{{ item.dest }}"
+    version: "{{ item.version }}"
+    accept_hostkey: yes
+    update: no
+  with_items:
```
basil: [nit] `with_items` is typically used when there are multiple items. In this case, there's a single item (and I don't expect we'll have multiple items in the future), so it seems more readable to get rid of the `with_items` block and inline the relevant values.

live-build/variants/internal-dcenter/ansible/playbook.yml
```diff
@@ -0,0 +1,35 @@
...
+  roles:
+    - appliance-build.minimal-common
+    - appliance-build.minimal-internal
+    - appliance-build.install-delphix-zfs-packages
+    - appliance-build.zfsonlinux-development
+    - appliance-build.masking-common
```
prakashsurya: why do we need masking on this system?

live-build/variants/internal-dcenter/ansible/playbook.yml
```diff
@@ -0,0 +1,35 @@
...
+    - appliance-build.masking-development
+    - appliance-build.virtualization-common
```
prakashsurya: why do we need virtualization on this system?

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
```diff
@@ -0,0 +1,99 @@
...
+- file:
+    path: /etc/resolv.conf
+    state: absent
+
+- lineinfile:
+    path: /etc/cloud/cloud.cfg
+    regexp: '^preserve_hostname: false'
+    line: 'preserve_hostname: true'
+
+- shell: |
```
basil: For my own education, why use `shell` rather than the more simple `command` module? Is there a specific reason we need a shell here? A comment might help clarify.

README.md
```diff
@@ -149,3 +149,103 @@ console and perform any post-boot verification that's required (e.g.
...
+
+    $ ls -l live-build/misc/ansible-roles/appliance-build.dcenter
+    total 0
+    drwxr-xr-x  3 root  root   96 Jun 19 19:17 files
+    drwxr-xr-x  4 root  root  128 Jun 19 19:17 tasks
+
+There are many roles that exist in "live-build/misc/ansible-roles" and can
+serve as good examples.
+
+### Step 2. Create the new variant directory structure
```
prakashsurya: These docs are great, but I do wonder if it'd be better to encapsulate all this into a script that we could call to create the new variant directory structure, and base ansible playbook/configuration. Then we could reference that script in these docs.

live-build/variants/internal-dcenter/ansible/playbook.yml
basil: For my own education, why do we need to use the deprecated Python 2 runtime as the Ansible Python interpreter? We use Python 3 in all the other Ansible playbooks in this repository, so it would make more sense to me to be consistent and to use a non-deprecated version of Python. If there's a special reason we need to be using Python 2, perhaps a comment might clarify.

live-build/variants/internal-dcenter/ansible/playbook.yml
```diff
@@ -0,0 +1,35 @@
...
+  gather_facts: no
+  vars:
+    ansible_python_interpreter: /usr/bin/python2
+  roles:
+    - appliance-build.minimal-common
+    - appliance-build.minimal-internal
+    - appliance-build.install-delphix-zfs-packages
+    - appliance-build.zfsonlinux-development
+    - appliance-build.masking-common
+    - appliance-build.masking-development
```
basil: For my own education, why do we need the `appliance-build.masking-development` role? Same comment applies to `appliance-build.virtualization-development`? If we can eliminate these roles, that could reduce the size of the image, which would be a net gain in my opinion.

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/ldap.yml
basil: [nit] Here and elsewhere, you use the `key=value` style of writing Ansible rather than the `key: value` style. I don't have a strong preference here, but I would prefer that we try to be consistent if possible. Since virtually all of the existing Ansible code in this repository uses the `key: value` style, I have a slight preference for keeping with that.

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/ldap.yml
basil: [nit] The other Ansible roles in this repository don't have names for each Ansible task, so to be consistent with them we might want to remove the names from the LDAP Ansible code as well.

pzakha: I am against removing the name here as it provides useful comments. It will be harder for whoever decides to update the names if they are not familiar with the code, so I feel like we should just leave the names here and update it elsewhere whenever we have time for it. I don't feel like consistency is as important (also debugging without names right now is definitely more tedious).

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
```diff
@@ -0,0 +1,99 @@
...
+    version: "{{ item.version }}"
+    accept_hostkey: yes
+    update: no
+  with_items:
+    - { repo: 'https://gitlab.delphix.com/devops/dcenter-gate.git',
+        version: projects/dx4linux,
+        dest: dcenter-gate }
+
+- copy:
+    dest: "/etc/dcenter.conf"
```
basil: This (and much of the other Ansible code in this role, such as the dependencies on third-party software) seems to me like it more properly belongs in a dedicated Ubuntu package for the DCenter application rather than in the Ansible role. When I first started using Ansible, I recall @tehranian mentioning that "Ansible is not a package manager" and that some jobs are better for Ansible while others are better for package managers. This is just a suggestion, and I don't think it has to be implemented in this change. Perhaps future work could create a proper package for DCenter and move much of this logic into the package management layer.

live-build/variants/internal-dcenter/ansible/playbook.yml
basil: > dcenter requires python2, and we use python2 to run ansible so that when it runs "pip" it will install the python 2 version of those modules.

If that's the case, I have a slight preference that we install these Python modules through Apt instead and run all Ansible playbooks through Python 3 for consistency. Either that, or create a `.deb` package for DCenter that uses `pip` as part of a separate build system to fetch these modules.

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/ldap.yml
prakashsurya: If we want to use names, then please add a name to all existing tasks and update the ansiblelint configuration to enforce the use of names moving forward (perhaps in a separate/isolated commit). It'll be good to come up with the convention we want to use for the names; personally, I find the `foo bar | blah blah` convention weird, since the output of ansible already tells us the name of the role for that task being run (the `foo bar` part just seems like needless duplication.

With that said, I rather just not use names altogether, and get ansiblelint to complain when a name is used. I've never found them useful.

live-build/base/config/package-lists/minimal.list.chroot
prakashsurya: In general I think I agree.. but.. I asked George to explicitly add it here, since it was being pulled in already (likely from a dependency of one of these other packages, but I haven't verified which), and George was already using it without really knowing how it was getting pulled in.

So, if we really don't want to add python2.7 to the base configuration, we need to figure out which other package in the base configuration depends on it, and remove that package(s) too.

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/ldap.yml
pzakha: Well, looks like maybe it's just me who finds them useful. I find they provide quick comments on what's being done without needing explicit comments. That said, if we are moving away from using names all-together then it'd make sense not to have names here. It's also possible that the way that we were using some roles made it harder to follow what's failing (e.g. the install-delphix-s3-packages role that is reference by several other roles), but since we are getting rid of it maybe it's fine.

live-build/variants/internal-dcenter/ansible/playbook.yml
prakashsurya: I asked for a comment as well.. python2 is needed here, so that when "pip" is used in the role, it installs the python2 module rather than the python3 module (assuming I understand correctly based on my prior discussions with george).

If we can use the "apt" packages rather than "pip" packages, this might no longer be necessary (and doing that would be better, imo).

live-build/variants/internal-dcenter/ansible/playbook.yml
grwilson: fixed

live-build/variants/internal-dcenter/ansible/playbook.yml
grwilson: fixed

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
grwilson: fixed

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
prakashsurya: I agree that creating a dcenter package, much like what we do for virtualization and masking, would be a good change.

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
grwilson: fixed

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
pzakha: Yup I think once we have infrastructure set-up for hosting misc delphix packages we will be able to convert quite a few things into packages.

live-build/base/config/package-lists/minimal.list.chroot
basil: FYI, these are the packages we use in the standard image that currently depend on Python 2.7:

```
bpfcc-tools
cppcheck
dstat
fio
python-bpfcc
python-netaddr
python-rtslib-fb
zfs-test
zfsutils-linux
```

We should probably work on getting the ZFS stuff to stop depending on Python 2.7.

I think I still have a slight preference for keeping the explicit installation of Python 2.7 to the `dcenter` variant. That way, once the above utilities stop depending on Python 2.7, we'll stop pulling it in (but will keep doing so for the `dcenter` variant, as long as it needs Python 2.7).

prakashsurya: That reasonable to me; I'm OK moving the explicit python2.7 dependency into the dcenter Ansible role.

The catch being, the dcenter ansible playbook then should use python3 as the python interpreter (like the other playbooks), since we can't assume python2 will be there prior to running the role. That'll require other changes, e.g. to how "pip" is used in that role; but if we can just stop using pip in favor of the python packages (I think we should, if we can), this might not be a problem.

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
basil: > once we have infrastructure set-up for hosting misc delphix packages

We already have the infrastructure set up for hosting custom _first-party_ packages (those built from Delphix source repositories). This infrastructure is being used for the Virtualization Engine and Masking Engine packages.

We don't yet have the infrastructure set up for hosting custom _third-party_ packages (those built from non-Delphix source repositories).

pzakha: Right, so ideally I think we'd want to have something more similar to pkg-build-gate for both first and third party packages. Right now we have a separate job and a separate S3 prefix location for each one of the Delphix packages (ZFS, masking, virtualization). This is fine for those larger projects but not very scalable in terms of development effort to add small new packages.

basil: That's an interesting idea, but this work isn't currently planned. I'd like to talk more offline to understand the use case you have in mind better.

grwilson:
I have tested this by deploying the OVA onto a test dcenter image and creating VMs using `dc clone-latest`

README.md
```diff
@@ -207,7 +207,7 @@ For this example, we add our new role to the playboodk as shown below:
       connection: chroot
       gather_facts: no
       vars:
-        ansible_python_interpreter: /usr/bin/python2
+        ansible_python_interpreter: /usr/bin/python3
```
pzakha: nit: is this still required? Wouldn't that be the default?

scripts/create-variant.sh
```diff
@@ -0,0 +1,66 @@
...
+PLAYBOOK_VARIANT=internal-minimal
+
+TOP=$(git rev-parse --show-toplevel 2>/dev/null)
+
+if [[ -z "$TOP" ]]; then
+	echo "Must be run inside the git repsitory."
+	exit 1
+fi
+
+#if [[ $EUID -ne 0 ]]; then
```
pzakha: Remove commented out code

README.md
grwilson: Yes, that the default and I just reverted back to the default. All the playbooks have this line.

scripts/create-variant.sh
grwilson: fixed

scripts/create-variant.sh
```diff
@@ -0,0 +1,61 @@
...
+	exit 1
+fi
+
+set -o errexit
+
+mkdir -p "$TOP/live-build/variants/$1"
+mkdir -p "$TOP/live-build/variants/$1/ansible"
+mkdir -p "$TOP/live-build/variants/$1/config"
+
+cd "$TOP/live-build/variants/$1/ansible"
```
prakashsurya: I don't think this `cd` is necessary.. `ln` should work if we do something like:
```
ln -s ../../../misc/ansible-roles "$TOP/live-build/variants/$1/ansible/roles"
```

scripts/create-variant.sh
```diff
@@ -0,0 +1,61 @@
...
+ln -s ../../../misc/ansible-roles roles
+cp "$TOP/live-build/variants/$PLAYBOOK_VARIANT/ansible/playbook.yml" .
+
+cd "$TOP/live-build/variants/$1/config"
```
prakashsurya: I don't think this `cd` is necessary.. `ln` should work if we do something like:
```
ln -s ../../../misc/live-build-hooks "$TOP/live-build/variants/$1/config/hooks"
```

scripts/create-variant.sh
```diff
@@ -0,0 +1,61 @@
...
+ln -s ../../../misc/live-build-hooks hooks
+
+git add "$TOP/live-build/variants/$1"
```
prakashsurya: Personally, I don't like having scripts mucking with my git index, but if folks think this is useful, I'm OK with it.

scripts/create-variant.sh
```diff
@@ -0,0 +1,61 @@
...
+		exit 1
+	fi
+else
+	echo "Invalid variant naming convention: $1" 1>&2
+	echo "Variants must start with either 'internal-<name>'" \
+		"or 'external-<name>'" 1>&2
+	exit 1
+fi
+
+set -o errexit
```
prakashsurya: should we also `set -o xtrace` too?

prakashsurya:
Also, please run `git-ab-pre-push` with your final version before pushing, and verify that succeeds. I'm a little worried that this will cause ENOSPC issues, since the amount of storage required by the bootstrap system will grow.

prakashsurya:
bors delegate+

PR #65, opened 2018-07-06
LX-960 Enable kernel crash dump support

jgallag88:
This adds and configures `kdump` so that we can get crash dumps, and adds `crash` so that we can examine them.

Note that this adds some time to the build, because downloading debug kernel from the Ubuntu repository for debug packages is quite slow (it takes ~15min to download <600MB). Once we start mirroring all of the upstream Ubuntu packages, this should no longer be an issue.

This change requires expanding the size of the disk on the bootstrap vm, which is blocked until the fix for DOAM-786 is deployed. However, I've tested building a single variant, and used qemu to verify that crash dumps are gathered successfully. I'll run `git-ab-pre-push` after the bootstrap golden image is updated.

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
@@ -323,3 +323,41 @@
   with_items:
     - { key: 'kernel.core_pattern', value: '/var/crash/core.%e.%p.%t' }
     - { key: 'vm.overcommit_memory', value: '1' }
+
+#
+# Add the apt repos that have the debug versions of packages, so that we
+# can download a debug kernel to use with crash.
+#
+- apt_repository:
+    repo: "{{ item }}"
+  with_items:
+    - deb http://ddebs.ubuntu.com bionic main restricted universe multiverse
+    - deb http://ddebs.ubuntu.com bionic-updates main restricted universe multiverse
+    - deb http://ddebs.ubuntu.com bionic-proposed main restricted universe multiverse
+  retries: 3
+  delay: 30
+  register: result
+  until: result is succeeded
+
+- shell: ls /boot/vmlinuz-* | sed 's|/boot/vmlinuz-||'
+  register: kernel_version
+
+- apt:
+    name: 'linux-image-{{ item }}-dbgsym'
+  with_items:
+    - '{{ kernel_version.stdout_lines }}'
+  retries: 3
+  delay: 30
+  register: result
+  until: result is succeeded
+
+#
+# Increase the amount of memory reserved for the crash kernel.
+# Empirically, it seems we need about 512M to get it to boot. If the
+# system has less than 1G, disable crash dumps rather than reserving
+# half of memory for the crash kernel.
+#
+- lineinfile:
+    path: /etc/default/grub.d/kdump-tools.cfg
+    regexp: '^GRUB_CMDLINE_LINUX_DEFAULT='
+    line: 'GRUB_CMDLINE_LINUX_DEFAULT="$GRUB_CMDLINE_LINUX_DEFAULT crashkernel=1024M-:512M"'
--- a/live-build/base/config/package-lists/minimal.list.chroot
+++ b/live-build/base/config/package-lists/minimal.list.chroot
@@ -42,3 +42,9 @@ net-tools
 open-vm-tools
 openssh-server
 ubuntu-minimal
+
+#
+# The following package contains the GPG keys which allow us to download
+# from the repositories which contain packages containing debug symbols.
+#
+ubuntu-dbgsym-keyring
--- a/live-build/base/config/package-lists/tools.list.chroot
+++ b/live-build/base/config/package-lists/tools.list.chroot
@@ -22,6 +22,7 @@
 
 atop
 bpfcc-tools
+crash
 dnsutils
 dstat
 emacs
@@ -29,6 +30,7 @@ gdb
 glances
 iftop
 inotify-tools
+kdump-tools
 procinfo
 sg3-utils
 strace


live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -323,3 +323,37 @@
...
+- shell: ls /boot/vmlinuz-* | sed 's|/boot/vmlinuz-||' | sort -V | tail -n1
+  register: kernel_version
+
+- apt:
+    name: 'linux-image-{{ kernel_version.stdout }}-dbgsym'
+
+#
+# Increase the amount of memory reserved for the crash kernel.
+# Empirically, it seems we need about 512M to get it to boot. If the
+# system has less than 1G, disable crash dumps rather reserving half of
```
ahrens: should be `rather *than* reserving`

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -323,3 +323,37 @@
...
+#
+# Add the apt repos that have the debug versions of packages, so that we
+# can download a debug kernel to use with crash.
+#
+- apt:
+    name: 'ubuntu-dbgsym-keyring'
+- apt_repository:
+    repo: "{{ item }}"
+  with_items:
+    - deb http://ddebs.ubuntu.com bionic main restricted universe multiverse
```
prakashsurya: This isn’t wrong, but I think it might be more consistent to add these in a file in “live-build/base/config/archives” .. then we could add the “ubuntu-dbgsym-keyring” package to the minimal package list file, right?

Would that work?

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -323,3 +323,37 @@
...
+#
+# Increase the amount of memory reserved for the crash kernel.
+# Empirically, it seems we need about 512M to get it to boot. If the
+# system has less than 1G, disable crash dumps rather reserving half of
+# memory for the crash kernel.
+#
+- lineinfile:
+    path: /etc/default/grub.d/kdump-tools.cfg
+    regexp: '^GRUB_CMDLINE_LINUX_DEFAULT='
+    line: 'GRUB_CMDLINE_LINUX_DEFAULT="$GRUB_CMDLINE_LINUX_DEFAULT crashkernel=1024M-:512M"'
```
prakashsurya: Why doesn’t the default values work? What are the defaults?

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -323,3 +323,37 @@
...
+  with_items:
+    - deb http://ddebs.ubuntu.com bionic main restricted universe multiverse
+    - deb http://ddebs.ubuntu.com bionic-updates main restricted universe multiverse
+    - deb http://ddebs.ubuntu.com bionic-proposed main restricted universe multiverse
+
+#
+# Assume that the newest kernel is the one we are going to boot, and
+# fetch the symbols for that one.
+#
+- shell: ls /boot/vmlinuz-* | sed 's|/boot/vmlinuz-||' | sort -V | tail -n1
```
prakashsurya: I wonder if we could remove the “tail” call, and then process the output using “with_items” below.. so we could handle the case where multiple kernels are found.

I could test this out if it’s something we want to support.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
jgallag88: We need `ubuntu-dbgsym-keyring` installed before we add these repositories, otherwise we will get an error when we try to fetch the list of packages from the repositories:
```
GPG error: http://ddebs.ubuntu.com bionic Release: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY C8CAB6595FDFF622
```
I think this will prevent us from adding these repositories in “live-build/base/config/archives”. We could probably move `ubuntu-dbgsym-keyring` into the minimal package list if we wanted though.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
jgallag88: The default is 128M, but with that amount of memory the kernel panics before it even starts the first userspace process because it runs out of memory. This doesn't appear to be related to any configuration that is specific to Delphix, because it happens on a standard install of Ubuntu too. I haven't looked too closely at where exactly all the memory is going. I'm planning on filing an upstream bug to see whether there is any way to reduce the amount of memory needed, or at least to update the default value.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
jgallag88: My thinking was that we would need a zfs kernel module for each kernel that we actually want to use, and since we only install one zfs module as part of the appliance build, we are only going to have one usable kernel, so we only need the debug symbols for that kernel.

However, instead of just choosing the newest kernel, it's probably better to either
 - Fail if the assumption that there is only one kernel installed doesn't hold, or
 - Just install symbols for every version like you mention, and let the other parts of the build worry about which and how many kernels to install.

I don't I really have a preference for one over the other.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
prakashsurya: Would that work? adding `ubuntu-dbgsym-keyring` to the minimal package list work?

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
prakashsurya: Thanks. This looks good, was just curious.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
prakashsurya: If the implementation for the latter doesn't make things more complicated, then I'd opt for that.

Although, doing what you have here should be fine.. if needed, we should have a check someplace else to fail when more than a single kernel is found.

Did you consider using `dpkg-query -Wf '${Version}' linux-image-generic` to generate the version, rather than `ls` and `sed`? I'm fine with your version, but am curious if `dpkg-query` didn't work or just wasn't tried.

prakashsurya: > Did you consider using dpkg-query -Wf '${Version}' linux-image-generic
Nvm.. Now after looking at the value returned by that command, it doesn't map directly to the package names, so we'd still have to massage it's output into the package that we want to install.

I'm good with what you have.

jgallag88: The only reason I didn't use that was because it required a bit more parsing/formatting to get the output we need. It outputs something like `4.15.0.23.25`, and the string we need is of the form `4.15.0-23`. That shouldn't be too hard to do though, so I can change it if using `dpkg-query` is more correct.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
prakashsurya: I’d like to do it that way, since that’ll be note consistent with how we’re adding other APT sources, if it works. Can you try that and let me know?

Otherwise this change looks good.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -323,3 +323,33 @@
...
+    repo: "{{ item }}"
+  with_items:
+    - deb http://ddebs.ubuntu.com bionic main restricted universe multiverse
+    - deb http://ddebs.ubuntu.com bionic-updates main restricted universe multiverse
+    - deb http://ddebs.ubuntu.com bionic-proposed main restricted universe multiverse
+
+- shell: ls /boot/vmlinuz-* | sed 's|/boot/vmlinuz-||'
+  register: kernel_version
+
+- apt:
```
basil: I think it's likely that this task will be flaky. My experience with `ddebs.ubuntu.com` has been that it's slow and flaky as hell. It may even be faster to fetch from our Artifactory instance. In the short term I'd suggest doing a 3x retry here, in the long term I'd suggest we factor this into our existing plans to mirror the regular Ubuntu repositories.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -323,3 +323,33 @@
...
     - { key: 'vm.overcommit_memory', value: '1' }
+
+#
+# Add the apt repos that have the debug versions of packages, so that we
+# can download a debug kernel to use with crash.
+#
+- apt_repository:
+    repo: "{{ item }}"
+  with_items:
+    - deb http://ddebs.ubuntu.com bionic main restricted universe multiverse
```
basil: If we're going to be adding `ddebs.ubuntu.com` to our build, we might as well start adding some useful userland debug symbols in addition to the kernel debug symbols that you are adding in this change (e.g., debug symbols for the C library, Java and Python runtimes, etc.). I don't think this needs to be done as part of your current change, but your change definitely enables it.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
jgallag88: Great point. I've added the retries, and I'll file a task to tracking mirroring `ddebs.ubuntu.com`.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
grwilson: I know you've been talking to the maintainers of this package, have you talked to them about increasing the default memory when using zfs too? It would be nice to have the package just handle this correctly.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
prakashsurya: > in the long term I'd suggest we factor this into our existing plans to mirror the regular Ubuntu repositories.

I think we'll need/want to do this for all the repositories we pull from; i.e. postgres, ansible, regular ubuntu, and now ddebs.ubuntu.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
jgallag88: @grwilson I actually don't think this is specific to ZFS, the amount of memory needs to be increased even when using an ext4 root filesystem. I am planning on following up with the package maintainers though.

grwilson: Great thanks!

jgallag88:
[pre-push run](http://collins.d.delphix.com:32382/job/devops-gate/job/projects/job/dx4linux/job/appliance-build/job/master/job/pre-push/2/console) (passed)

prakashsurya:
bors delegate+

Please squash to a single commit before landing, and then add comment of “bors r+” when your ready for it to be merged.

PR #68, opened 2018-07-19
LX-1194 ZFS panics should result in kernel panics

jgallag88:
Assertion failures in zfs call `spl_panic`, which by default does not trigger a kernel panic. This change sets the `spl_panic_halt` parameter for the `spl` module so that we will get a kernel panic, and therefore a crash dump, on zfs assertion failures.

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
@@ -361,3 +361,12 @@
     path: /etc/default/grub.d/kdump-tools.cfg
     regexp: '^GRUB_CMDLINE_LINUX_DEFAULT='
     line: 'GRUB_CMDLINE_LINUX_DEFAULT="$GRUB_CMDLINE_LINUX_DEFAULT crashkernel=1024M-:512M"'
+
+# Panic on ZFS assertion failures
+- lineinfile:
+    path: /etc/modprobe.d/zfs.conf
+    create: yes
+    line: 'options spl spl_panic_halt=1'
+  # Because the spl module is loaded from the initramfs when using zfs on root,
+  # we need to rebuild the initramfs to account for the modprobe.d change.
+  notify: update initramfs
--- a/live-build/misc/ansible-roles/appliance-build.minimal-common/handlers/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.minimal-common/handlers/main.yml
@@ -1,0 +1,19 @@
+#
+# Copyright 2018 Delphix
+#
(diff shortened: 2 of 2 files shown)


live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -351,13 +351,18 @@
...
 
-#
-# Increase the amount of memory reserved for the crash kernel.
-# Empirically, it seems we need about 512M to get it to boot. If the
-# system has less than 1G, disable crash dumps rather than reserving
-# half of memory for the crash kernel.
-#
 - lineinfile:
-    path: /etc/default/grub.d/kdump-tools.cfg
+    path: "/etc/default/grub.d/{{ item.file }}.cfg"
```
ahrens: What are the tradeoffs between making the change here, vs in `/etc/modprobe.d/zfs.conf`?

jgallag88: I thought that that didn't work with zfs on root because zfs is loaded before the root filesystem was mounted. However, I just did a bit more research, and it looks like I just needed to regenerate the initramfs for the change to take effect. I'll update this to use `/etc/modprobe.d`.

jgallag88:
Switched from passing a kernel command line arg to adding an file in `/etc/modprobe.d`. Pre-push run is [here](http://ops.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build-orchestrator-pre-push/11/console).

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -361,3 +361,12 @@
...
     line: 'GRUB_CMDLINE_LINUX_DEFAULT="$GRUB_CMDLINE_LINUX_DEFAULT crashkernel=1024M-:512M"'
+
+# Panic on ZFS assertion failures
+- lineinfile:
+    path: /etc/modprobe.d/zfs.conf
+    create: yes
+    line: 'options spl spl_panic_halt=1'
+# Because the spl module is loaded from the initramfs when using zfs on root,
+# we need to rebuild the initramfs to account for the modprobe.d change.
+- shell: update-initramfs -u
```
prakashsurya: It'd be better to run this command as an Ansible "handler", such that it's only executed when the "zfs.conf" file changes. We don't have an example of a handler yet in this repository, but it's pretty simple; I can give you some help if you want to look into doing it that way.

With that said, if you'd rather just land this as is, I'm OK with that. This isn't quite as idempotent as it could be, but given our use case, that's not a practical concern.

PR #74, opened 2018-07-23
Use LDAP with "internal-dev" variant

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://ops.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build-orchestrator-pre-push/20/)

When that finishes (assuming it's successful), I'll test that LDAP actually works on the image it generates.

ahrens:
Does this mean we'll be able to log in as ourselves and get our home directories????? 🙏

prakashsurya:
I tried logging into the VM generated by the pre-push job, and it looks like LDAP still isn't working. There's probably some additional configuration required to get this working.

@grwilson do you apply additional configuration, in addition to the "delphix-ldap" role, to get LDAP working on dcenter?

prakashsurya:
Looking more closely, I wasn't correctly testing the right image when I commented above. Now using the correct image, LDAP appears to be working:
```
$ ssh psurya@ps-lx-ldap-1.dlpxdc.co
...
$ id
uid=1138(psurya) gid=10(uucp) groups=10(uucp)
$ dpkg -l | grep delphix-entire
ii  delphix-entire                                1.0.0-20180725T163056              all          Entirety of Delphix Appliance (internal-dev)
$ pwd
/home/psurya
```

PR #75, opened 2018-07-23
Fail build when 2 kernel versions are detected

### LINKED ISSUE #73: Fail build when 2 kernel versions are detected
When we build the appliance, we install the latest version of kernel available. When we build ZFS packages, we also build them against the latest version of the kernel available. However, if the ZFS packages haven't been rebuilt in a while and the latest kernel version changes, we'll end up installing ZFS modules that target an older version of the kernel.

This will result in the appliance having two kernel versions, one of which will not have our ZFS bits installed. The appliance will be broken overall, so we should fail the build when we get into this situation to avoid updating our internal dcenter and DCOA images.

pzakha:
This is intended to solve the case where the version of ZFS modules
installed does not match the version of the kernel that comes with
the package 'linux-generic', resulting in a broken appliance.

Pre-commit:
http://ops.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build-orchestrator-pre-push/26 (pass)

build with a different zfs kernel modules version (should fail):
http://ops.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build/job/master/job/pre-push/121/console (fail as expected)

Fixes #73

### DIFF
--- a/live-build/misc/live-build-hooks/81-upgrade-repository.binary
+++ b/live-build/misc/live-build-hooks/81-upgrade-repository.binary
@@ -27,6 +27,28 @@
 set -o pipefail
 
 #
+# Sanity check to make sure we haven't installed multiple kernel versions.
+# This could happen if the version of the ZFS modules we install is not the
+# same as the kernel version installed with the package 'linux-generic'.
+#
+# If the kernel version of the ZFS packages is lower than the one for
+# 'linux-generic' then we will boot into a kernel that doesn't have Delphix ZFS
+# modules and will use the built-in Ubuntu ones (while using Delphix userland
+# packages, which are not compatible). If on the other hand the kernel version
+# is higher, then we will boot into a kernel image that contains our ZFS
+# modules (ZFS modules depend on 'linux-image-KVERS', and will install it
+# along). However, the 'linux-generic' package has more dependencies than just
+# 'linux-image-KVERS', which will not be installed along with the ZFS modules
+# for the running kernel version, and which our appliance expects to have.
+#
+# shellcheck disable=SC2012
+if [[ $(ls binary/lib/modules | wc -l) -gt 1 ]]; then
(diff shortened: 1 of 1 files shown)


live-build/misc/live-build-hooks/81-upgrade-repository.binary
```diff
@@ -26,6 +26,18 @@
 
 set -o pipefail
 
+#
+# Sanity check to make sure we haven't installed multiple kernel versions.
+# This could happen if the version of the ZFS modules we install is not the
+# same as the kernel version installed with the package 'linux-generic'.
```
prakashsurya: This comment doesn't discuss why having multiple kernel versions installed is a bad thing. If we had multiple kernels installed, and we had the delphix zfs modules installed for each kernel version, then everything would work just fine. It might be worth expanding on this comment to explain the actual failure scenario better.

pzakha: I've updated the comment.

>  It might be worth expanding on this comment to explain the actual failure scenario better.

See 2nd link in description.

prakashsurya: I think this comment still leaves out the most important information, which is that we can wind up with an image that boots a kernel version that doesn't have the corresponding delphix zfs kernel module package installed, but _does_ have the delphix zfs userspace packages installed.

That is, unless you are trying to convey this in the sentence:

> there are other packages that come with 'linux-generic' that won't be installed for that particular kernel version

?

pzakha: I've updated comments with more details.

pzakha:
Please see https://github.com/delphix/appliance-build/issues/103 for why we would see this error message and for a workaround.

PR #77, opened 2018-07-24
LX-194 pam module for linux - phase I  (add dev libraries needed for building PAM module)

rasantel:


jgallag88:
I think the best way to do this long term would be to have a `delphix-virtualization-dev` package in the app-gate, which would get built alongside the `delphix-virtualization` package. It wouldn't have any content, but would just declare these dependencies. Then we could install the dev package as part of the virtualization-development role.

It would take some infrastructure changes to get this working though, so I think this change is very reasonable for now.

prakashsurya:
> I think the best way to do this long term would be to have a delphix-virtualization-dev package in the app-gate, which would get built alongside the delphix-virtualization package. It wouldn't have any content, but would just declare these dependencies. Then we could install the dev package as part of the virtualization-development role.

I agree in concept. I think a good place to put these dependencies would be in the "build dependencies" metadata for the package, which specify the packages that are required to turn a "source package" into the resultant "binary package".

I don't know how this would apply given how we're using gradle to build the virtualization package, but this is how this sort of thing is generally done for packages.

This would apply for our other packages too (e.g. masking, zfs, etc); we'd just need appliance-build to pull in the build dependencies for all of our delphix package for the "internal-dev" variant.

Until we get to that point, though, we're stuck with this solution.

rasantel:
Forgot to mention, yes, pre-push worked: http://ops.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build-orchestrator-pre-push/18

prakashsurya:
While I'm thinking about it, if we used a proper build system for building our packages, we'd build the binary package in an environment that was generated using the build dependencies (e.g. a chroot with only the build dependency packages installed); this way we could verify we don't leave a necessary build dependency out of the list. There's tools like "pbuilder" that help with this sort of thing.

PR #78, opened 2018-07-30
LX-1213 Install azure udev rules

shartse:
This change installs the Azure Linux agent by default as a core dependency but disables its service. We'll eventually want to re-enable the service for Azure VMs, though probably with reduced functionality (configured in `/etc/waagent.conf`). I decided to do this much now because we need the azure specific udev rules that come in this package and we'll need want the agent at some point as well. 

I tested this by running through the live build process and using qemu to check that the image booted properly, had the new udev rules and that `walinuxagent.service` was present but disabled.


prakashsurya:
bors delegate+

PR #81, opened 2018-08-01
LX-1217 Generate crash dump on NMI

jgallag88:
The appliance should panic and generate a crash dump when it receives an NMI, to allow debugging of engines which are in some bad state such that interactive debugging isn't possible.


jgallag88:
[ab-pre-push](http://ops.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build-orchestrator-pre-push/47/)

PR #82, opened 2018-08-06
Use "sysctl.d" for appliance customizations

prakashsurya:
This change modifies the "minimal-common" Ansible role to no longer modify
the "/etc/sysctl.conf" file, but instead use the "sysctl.d" directory. This
should result in no changes to the normal operation of the appliance, but
makes upgrade more resilient as our "local" changes will not be overwritten
by the maintainer supplied "/etc/sysctl.conf" file.

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
@@ -226,10 +226,6 @@
           lock_passwd: false
 
 #
-# We don't want to have our secondary addresses deleted when someone
-# deletes the primary IP address of a network interface, thus we enable
-# the promotion of secondary addresses to primary ones.
-#
 # As for the priority prefix for our config file (see what that means
 # in /etc/sysctl.d/README) we pick 50 which is higher than what other
 # Ubuntu packages will be installing but less than customization made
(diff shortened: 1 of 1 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://ops.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build-orchestrator-pre-push/53/)

prakashsurya:
Looking at the README for `/etc/sysctl.d`, it says:
```
This directory contains settings similar to those found in /etc/sysctl.conf.
In general, files in the 10-*.conf range come from the procps package and
serve as system defaults.  Other packages install their files in the
30-*.conf range, to override system defaults.  End-users can use 60-*.conf
and above, or use /etc/sysctl.conf directly, which overrides anything in
this directory.
```
I need to verify the settings we are changing aren't already contained in the default `/etc/sysctl.conf` file, otherwise it seems like our changes won't take affect.

prakashsurya:
Verified our custom settings are still applied with this change:
```
$ grep kernel.core_pattern /etc/sysctl.conf 
$ grep vm.overcommit_memory /etc/sysctl.conf
$ grep kernel.unknown_nmi_panic /etc/sysctl.conf
$ sysctl kernel.unknown_nmi_panic
kernel.unknown_nmi_panic = 1
$ sysctl vm.overcommit_memory
vm.overcommit_memory = 1
$ sysctl kernel.core_pattern
kernel.core_pattern = /var/crash/core.%e.%p.%t
```

PR #83, opened 2018-08-06
WIP: Install new "delphix-foundation" package

prakashsurya:
This change modifies the base live-build configuration, to install a new
"delphix-foundation" package which will depend on all of the packages
that constitute the "foundation" of the Delphix appliance. As a result,
the live-build configuration only needs to install this single package,
rather than listing all of the packages that constitute the "base".


prakashsurya:
This is intended to be used in conjunction with [this](https://gitlab.delphix.com/prakash.surya/delphix-foundation) new repository/package.

PR #85, opened 2018-08-13
bootstrap configuration should always use the latest apt index

shartse:
The `appliance-build.bootstrap apt` task has been failing when trying to install the latest packages. Basil suspects that the issue is that it's trying to install a package which is no longer available from Ubuntu because the `apt` index is old (it’s from when the `bootstrap-18-04` group was made). 

He suggested adding these commands to update the `apt` indices before performing the install so that the latest version is always pulled in.

Testing:
I ran `ansible-playbook bootstrap/playbook.yml` without the changes and saw the same failures we've been seeing in the Jenkins job. With the changes, `ansible-playbook bootstrap/playbook.yml` completes successfully.

### DIFF
--- a/bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
+++ b/bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
@@ -19,6 +19,16 @@
   register: toplevel
 
 - apt:
+    name: aptitude
+    state: present
+
+- apt:
+    update_cache: yes
+
+- apt:
+    upgrade: yes
+
+- apt:
     name: "{{ item }}"
     state: present
   with_items:


bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
```diff
@@ -18,6 +18,16 @@
 - shell: git rev-parse --show-toplevel 2>/dev/null
   register: toplevel
 
+- apt:
+    name: aptitude
+    state: latest
+    update_cache: yes
+  become: true
```
basil: Here and below, the `become: true` is unnecessary because the whole playbook is executed with `become: true` (see `bootstrap/playbook.yml`).

basil:
bors delegate+

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
```diff
@@ -18,6 +18,16 @@
 - shell: git rev-parse --show-toplevel 2>/dev/null
   register: toplevel
 
+- apt:
+    name: aptitude
+    state: present
+
+- apt:
```
prakashsurya: Is there a reason we use multiple “apt” invocations? I thought they could be pulled into a single one, but until I can look at the docs, idk for sure. 

I’m fine with this either way.

shartse: I'm going to integrate this now so we can get the builds going again (I just got a passing `git-ab-pre-push` run) but I'll look at the docs and see if it can be simplified.

PR #87, opened 2018-08-16
LX-198 delphix nsswitch functionality (nss and passwordless auth config)

rasantel:
### Change
This companion change of LX-198 configures the CLI NSS module in virtualization appliances. This includes updates to the NSS configuration to add the new module and the NSCD configuration to prevent our module from being skipped by the `passwd` NSCD cache. For development variants, it also adds the `libnss3` development libraries.

It also takes charge of creating the `cli` user used as a placeholder of the uid 65434 assumed by all cli users. This user is no longer created by app gate code because `refresh_databases.sh` is going away (whose real purpose was to update `/etc/passwd` and `/etc/shadow` with native engine users, which we don't want to do anymore).

Finally, this change fixes a minor issue in configuring sshd and, more importantly, fixes the LDAP role which in its current form removes any modules from the NSS `passwd` database other than `files` (i.e. the `systemd` module and the new  `delphix` NSS module).

### Testing
The appliance build process completes successfully for the `internal-dev` variant: http://selfservice.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build-orchestrator-pre-push/41/flowGraphTable/
Cloned a VM from the resulting group in DCoA and verified it contains all the configuration changes and including the new NSS module deployed to it (the latter is thanks to the app-gate solution to be reviewed separately).

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.delphix-ldap/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.delphix-ldap/tasks/main.yml
@@ -44,13 +44,18 @@
 
 - lineinfile:
     dest: /etc/nsswitch.conf
-    regexp: "{{ item.regexp }}"
-    line: "{{ item.line }}"
+    regexp: "^({{ item.db }}:(?!.*\\sldap\\s).*$)"
+    line: "\\1 ldap"
+    backrefs: yes
   with_items:
-    - { regexp: "^automount:.*$", line: "automount:\tfiles ldap" }
-    - { regexp: "^group:.*$", line: "group:\tcompat ldap" }
-    - { regexp: "^passwd:.*$", line: "passwd:\tcompat ldap" }
-    - { regexp: "^shadow:.*$", line: "shadow:\tcompat ldap" }
+    - { db: "passwd" }
+    - { db: "group" }
+    - { db: "shadow" }
+
+- lineinfile:
+    dest: /etc/nsswitch.conf
+    regexp: "^automount:.*$"
+    line: "automount:\tfiles ldap"
 
 - file:
     path: /home
--- a/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
@@ -30,6 +30,12 @@
   when: lookup('env', 'APPLIANCE_USERNAME') == ''
 
 #
+# Apt cache needs to be up to date.
+#
+- apt:
+    update_cache: yes
+
+#
 # To adhere to the priciple of least surprise, we disable automatic
 # upgrades here; this package can always be added on a specific system
 # when it's desired.
@@ -374,8 +380,8 @@
 
 - lineinfile:
     path: /etc/ssh/sshd_config
-    regexp: "^#?{{ item.key }}="
-    line: "{{ item.key }}={{ item.value }}"
+    regexp: "^#?{{ item.key }} "
+    line: "{{ item.key }} {{ item.value }}"
   with_items:
     #
     # Configure SSH to allow PAM "conversations" (interactions with the user).
--- a/live-build/misc/ansible-roles/appliance-build.virtualization-development/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.virtualization-development/tasks/main.yml
@@ -36,6 +36,9 @@
     - ant
     - git
     - libcurl4-openssl-dev
+    - libnss3-dev
+    - libnss3-dbg
+    - libnss3-tools
     - libpam0g-dev
     - libssl-dev
   register: result


bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
```diff
@@ -18,6 +18,10 @@
 - shell: git rev-parse --show-toplevel 2>/dev/null
   register: toplevel
 
+- name: Ensure that the apt cache is up to date
```
basil: This repository generally doesn't make use of `name:`, and in previous reviews @prakashsurya and I both expressed our opinion that we shouldn't be using `name:` in new tasks, for a few reasons:

- Existing tasks don't use `name:`, and we would like to be consistent within the same repository.
- The names were only ever truly useful in order to help people map an error in a production run to a specific source file and line number; however, in the latest version of Ansible, source files and line numbers can be printed for tasks without having to use `name:`.

As such, I would recommend removing the use of `name:` both here and elsewhere.

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
```diff
@@ -18,6 +18,10 @@
...
+  apt:
+    update_cache: yes
```
basil: A few lines down, we have an identical `apt` task that updates the cache. So what is the reason for adding this additional cache-updating task?

live-build/misc/ansible-roles/appliance-build.delphix-ldap/tasks/main.yml
```diff
@@ -30,6 +30,13 @@
     - nfs-common
     - nscd
 
+- name: Prevent NSDC from ignoring the CLI NSS module
```
basil: I think you meant "NSCD" instead of NSDC. Also, I'm not sure if this should be capitalized; [the man page](http://manpages.ubuntu.com/manpages/bionic/man8/nscd.8.html) uses lowercase letters. I'd personally describe this program with the identifier `nscd(8)` (a Solaris convention, but a generally useful one in my opinion).

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -29,6 +29,12 @@
     msg: "Required environment variable 'APPLIANCE_USERNAME' is empty."
   when: lookup('env', 'APPLIANCE_USERNAME') == ''
 
+#
+# Apt cache needs to be up to date.
+#
+- apt:
+    update_cache: yes
```
basil: Could you explain why this change is necessary? What problem does this solve?

live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
```diff
@@ -45,3 +45,34 @@
       domain=
       dns1=
       dns2=
+
+- name: Add the CLI user with home directory
+  user:
+    name: cli
+    comment: Delphix CLI User
+    uid: 65434
+    non_unique: yes
```
basil: [nit] This is a fairly unusual option to enable, but I think I understand why it's necessary (because all CLI users share the same UID). We might want to consider adding a comment clarifying this.

live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
```diff
@@ -45,3 +45,34 @@
...
+    group: staff
+    home: /export/home/cli
+    create_home: yes
+
+- name: Create directory where authorized keys will be specified for all CLI users
+  become_user: cli
+  file:
+    path: /export/home/cli/.ssh
+    state: directory
```
basil: Should we ensure that the directory has 0700 permissions? I seem to recall that these permissions are necessary for `sshd(8)` to be able to read the keys.

live-build/misc/ansible-roles/appliance-build.virtualization-development/tasks/main.yml
```diff
@@ -87,3 +87,11 @@
     password:
       "{{ lookup('env', 'APPLIANCE_PASSWORD') | password_hash('sha512') }}"
   no_log: true
+
+- apt:
+    name: "libnss3*"
```
basil: What are the tradeoffs of installing the `libnss3*` packages here rather than listing these packages as dependencies of the Virtualization Engine package?

prakashsurya: +1 .. I think this is the wrong place to do this, but perhaps I'm overlooking something.

live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
```diff
@@ -45,3 +45,34 @@
...
+
+- name: Configure authorized-keys file names for passwordless SSH into CLI
+  lineinfile:
+    path: /etc/ssh/sshd_config
+    regexp: "^#?{{ item.key }} "
+    line: "{{ item.key }} {{ item.value }}"
+  with_items:
+    - { key: "AuthorizedKeysFile", value: ".ssh/authorized_keys /export/home/cli/.ssh/authorized_keys.%u" }
```
prakashsurya: should `/export/home/cli/.ssh/authorized_keys.%u` actually be `/export/home/%u/.ssh/authorized_keys.%u`? or even `.ssh/authorized_keys.%u`? perhaps I misunderstand how this value is being used; if so, a comment would be helpful.

live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
```diff
@@ -45,3 +45,34 @@
       domain=
       dns1=
       dns2=
+
+- name: Add the CLI user with home directory
```
prakashsurya: I think all of the changes to this Ansible role should actually live in the app-gate. Creating the CLI user should live in the post-install hook for the app-gate package, and the sshd configuration should live in the delphix-config role.

Why did you choose this approach vs. what I'm suggesting?

live-build/misc/ansible-roles/appliance-build.virtualization-development/tasks/main.yml
prakashsurya: Now I see this is being done in the development role.. if we don't want these packages on the qa and/or customer image, then I take back my prior statement; this is probably the right place to install the packages.

If that's the case, why install with a glob of packages, rather than being more explicit about the packages we require?

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
rasantel: An apt task would fail next during my appliance-build orchestrator run. It would fail to find a dependency on some repository (don't recall which one it was exactly).

live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
rasantel: that's right. Although I'll move this to the post-install hook of the debian package, as per @prakashsurya's comment below.

live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
rasantel: Those restrictive permissions would be needed if we had an ssh key pair for the `cli`. In this case, we never really log in as `cli`. All we care about are the the `authorized_keys.*` files. I can try making this 0700 anyway, if it doesn't interfere with our management of the `authorized_keys.*` files.

live-build/misc/ansible-roles/appliance-build.virtualization-development/tasks/main.yml
rasantel: I see that using the wildcard is confusing. This is meant to install the -dev and -dbg packages, which are only needed for development. I'll make this explicit.

live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
rasantel: We don't really need a unique home directory per user. We have always used the `cli` home as a neutral location for authorized-keys files.

live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
rasantel: I thought of this location as part of the appliance preparation, but the post-install hook as you suggest might be a better place (more robust if we ever install the package without going through this role).

rasantel:
All comments are addressed in the updated pull request.

live-build/misc/ansible-roles/appliance-build.virtualization-development/tasks/main.yml
```diff
@@ -87,3 +87,15 @@
     password:
       "{{ lookup('env', 'APPLIANCE_PASSWORD') | password_hash('sha512') }}"
   no_log: true
+
+- apt:
+    name: "{{ item }}"
+    state: present
+  with_items:
+    - libnss3-dev
```
prakashsurya: why not combine this list of packages with the list above, on line 32?

live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
```diff
@@ -15,9 +15,16 @@
...
+#
+# Install the virtualization package. This package needs to configure nscd,
+# which might not be installed yet, so we make sure nscd is installed first.
+#
 - apt:
-    name: delphix-virtualization
+    name: "{{ item }}"
     state: present
+  with_items:
+    - nscd
```
prakashsurya: Why do you list `nscd` here, rather than make it a dependency of the `delphix-virtualization` package?

rasantel:
Test run with the latest app gate change under review in http://reviews.delphix.com/r/42625/diff/1:
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/dx-integration-tests/9025/flowGraphTable/

rasantel:
Latest test run: http://selfservice.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build-orchestrator-pre-push/68

basil:
bors delegate+

rasantel:
bors +

PR #89, opened 2018-08-21
Add delphix-upgrade package to base configuration

prakashsurya:
This change adds the "delphix-upgrade" package to the base system
configuration, as well as extends the build-ancillary-repository.sh
script to download this new package from AWS S3.

### DIFF
--- a/live-build/base/config/package-lists/minimal.list.chroot
+++ b/live-build/base/config/package-lists/minimal.list.chroot
@@ -38,6 +38,7 @@ delphix-zfs
 auditd
 cloud-init
 debsums
+delphix-upgrade
 net-tools
 open-vm-tools
 openssh-server
--- a/scripts/build-ancillary-repository.sh
+++ b/scripts/build-ancillary-repository.sh
@@ -168,15 +168,20 @@ function build_ancillary_repository() {
 #    environment variables, and the script will work as expected.
 #
 
-AWS_S3_URI_VIRTUALIZATION=$(resolve_s3_uri \
-	"$AWS_S3_URI_VIRTUALIZATION" \
-	"$AWS_S3_PREFIX_VIRTUALIZATION" \
-	"dlpx-app-gate/master/build-package/post-push/latest")
-
 AWS_S3_URI_MASKING=$(resolve_s3_uri \
 	"$AWS_S3_URI_MASKING" \
 	"$AWS_S3_PREFIX_MASKING" \
 	"dms-core-gate/master/build-package/post-push/latest")
+
+AWS_S3_URI_UPGRADE=$(resolve_s3_uri \
+	"$AWS_S3_URI_UPGRADE" \
+	"$AWS_S3_PREFIX_UPGRADE" \
+	"appliance-upgrade/master/build-package/post-push/latest")
+
+AWS_S3_URI_VIRTUALIZATION=$(resolve_s3_uri \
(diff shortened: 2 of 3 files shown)


prakashsurya:
This is dependent on the devops-gate change out for review [here](http://reviews.delphix.com/r/42674/).

prakashsurya:
cc: @basil @grwilson @grodr @pzakha

pzakha:
How hard would it be to get that package generated along with the virtualization package, from the app-gate repository? I'm thinking about the cost of maintaining another repository, i.e having to provide another link to appliance-build, flag days for syncing up with app-gate, etc. Are we creating a new repository out of convenience or because we feel like upgrade logic should not be tied so closely to app-gate anymore?

prakashsurya:
> How hard would it be to get that package generated along with the virtualization package, from the app-gate repository?

I didn't try, but my guess is it would be more work than it was to create this repository/package, due to how the app-gate build works.

> Are we creating a new repository out of convenience or because we feel like upgrade logic should not be tied so closely to app-gate anymore?

Both. It was easier for me to do it this way, but I also think this logic should not be tightly coupled with our other repositories; i.e. the app-gate, nor the dms-gate. Further, it'd be good (imo) to enable upgrades of our dcenter appliance using this new package; that appliance doesn't have the app-gate nor dms-gate installed (please, correct me if I'm wrong).

> I'm thinking about the cost of maintaining another repository, i.e having to provide another link to appliance-build, flag days for syncing up with app-gate, etc.

While I might agree with your concern, we've already committed to this approach. If we want to challenge this design, I'd rather do that with a broader view, than focusing only on this one new package/repository.

pzakha:
> While I might agree with your concern, we've already committed to this approach. If we want to challenge this design, I'd rather do that with a broader view, than focusing only on this one new package/repository.

I'm not sure to which extent we have committed to that approach. It's true we currently do that for appliance/masking/ZFS, although those are the main components of our product. In any case, I guess the real issue is that we do not currently have infrastructure in place for packages outside of those main 3. Once we have that figured out we should integrate this new upgrade-repository in there, perhaps something similar to our current pkg-build-gate.

The code LGTM.

prakashsurya:
> Once we have that figured out we should integrate this new upgrade-repository in there, perhaps something similar to our current pkg-build-gate.

I'd love to do it this way, but my expectation has been (set by prior discussion on this topic) this is how "first party" packages were designed to work; and what comes out of our support for custom "third party" packages would not work for things like this. i.e. if we maintain a git repository with changes to a project (like this new delphix-upgrade project), that would not be a candidate for what's planned to come out of our "port of pkg-build-gate to linux".

I could be mistaken, though.

pzakha:
I think it would be terrible if were to require a separate S3 link for each first-party package. This would make the cost of adding first-party packages high enough to deter us from using them.

prakashsurya:
I agree.

gvrodr:
Is it probable that we will add additional first party packages?

prakashsurya:
> Is it probable that we will add additional first party packages?

I think so, yes. For example, @brad-lewis is planning to create a new repository with the code out for review [here](http://reviews.delphix.com/r/42541).

prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build-orchestrator-pre-push/54/)

PR #90, opened 2018-08-23
Add upgrade-scripts to upgrade image contents

prakashsurya:


live-build/misc/upgrade-scripts/verify
```diff
@@ -0,0 +1,128 @@
...
+	# the verification container. This is intended to make it easier
+	# to diagnose and debug verification failures on developer
+	# machines.
+	#
+	if [[ $rc -ne 0 ]] && [[ -n "$DLPX_DEBUG" ]] && $DLPX_DEBUG; then
+		report_progress_inc 80 "Skipping upgrade verification cleanup"
+	else
+		report_progress_inc 80 "Performing upgrade verification cleanup"
+
+		"${BASH_SOURCE%/*}/verify-container" stop "$CONTAINER"
```
pzakha: nit: here and elsewhere, perhaps replace `${BASH_SOURCE%/*}` by a variable defined in `common.sh`?

live-build/misc/upgrade-scripts/verify
```diff
@@ -0,0 +1,128 @@
...
+	f | l | o) eval "opt_$c=$OPTARG" ;;
+	d) eval "opt_$c=true" ;;
+	*) usage "illegal option -- $OPTARG" ;;
+	esac
+done
+
+[[ "$EUID" -ne 0 ]] && die "must be run as root"
+[[ -n "$DLPX_VERSION" ]] || usage
+
+trap cleanup EXIT
```
pzakha: That's cool.

live-build/misc/upgrade-scripts/verify-container
```diff
@@ -0,0 +1,274 @@
...
+	# verification software running in the container to expose their status
+	# and/or results to the host.
+	#
+	# We always want to enable the results of any such verifcation to be
+	# communicated back to the host; thus, we always want to ensure this
+	# directory exists. Otherwise, if the directory didn't exist and we
+	# still attempted to bind mount it into the container, the container
+	# would fail to start.
+	#
+	if [[ ! -d "$DROPBOX_DIR" ]]; then
```
pzakha: Sounds like this is a temporary workaround. I'm assuming that if the dropbox directory isn't there that'd be due to some bug that it would be good to catch.

live-build/misc/upgrade-scripts/verify-container
```diff
@@ -0,0 +1,274 @@
...
+	;;
+stop)
+	[[ $# -le 1 ]] && usage "too few arguments specified"
+	CONTAINER="$2"
+	stop
+	;;
+destroy)
+	[[ $# -le 1 ]] && usage "too few arguments specified"
+	CONTAINER="$2"
+	destroy
```
pzakha: It might be good to have `verify-container destroy <container>` return a non-zero exit code if the destroy actually fails.

live-build/misc/live-build-hooks/81-upgrade-repository.binary
```diff
@@ -126,6 +126,6 @@ aptly repo add upgrade-repository ./*.deb
 
 aptly publish repo -skip-signing upgrade-repository
 
-tar -cf "$APPLIANCE_VARIANT.repo" -C ~/.aptly .
+tar -czf "$APPLIANCE_VARIANT.tar.gz" -C upgrade-scripts . -C ~/.aptly .
```
pzakha: perhaps we'd want a more meaningful name than just `$APPLIANCE_VARIANT.tar.gz`? e.g. `$APPLIANCE_VARIANT-upgrade.tar.gz` or `$APPLIANCE_VARIANT.upgrade.tar.gz`

prakashsurya:
> The git-ab-pre-push link above points to this review instead of a pre-push run.
Sorry, I've updated that comment/link.

> We should probably run some manual testing before integrating this.
Yes, I'm doing this as I test the app-gate changes that'll consume this image. I'll include the manual testing in the app-gate review. I can also put that testing here, but wasn't planning to; let me know if you think I should.

live-build/misc/upgrade-scripts/common.sh
```diff
@@ -0,0 +1,52 @@
...
+
+function report() {
+	echo -e "$(date +%T:%N:%z): $1"
+}
+
+function report_progress_inc() {
+	echo "Progress increment: $(date +%T:%N%z), $1, $2"
+}
+
+function configure_apt_sources() {
```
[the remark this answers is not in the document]
prakashsurya: Alternatively, I can just remove this function, since it's now only used in a single script.

I'd rather do that, then add comments/documentation, since it only has a single consumer.

live-build/misc/upgrade-scripts/verify
prakashsurya: We'd still have to use `${BASH_SOURCE%/*}` when sourcing `common.sh`.. as a result, I think it's fine to just use it throughout, but I don't feel strongly.. I can make that change if you think it'd make things more readable.

It's worth noting though, `${BASH_SOURCE%/*}` isn't necessarily the same when used in `common.sh`. Practically it is, since `common.sh` and the scripts are in the same directory, but if we later were to move `common.sh` that'd break.

live-build/misc/upgrade-scripts/verify-container
prakashsurya: No, this isn't a woraround.

Currently, `DROPBOX_DIR` is created by the verification JAR. We can't do that, since we not run this JAR in the container, so we need to create it outside of that JAR, which is why I do it here.

See [here](http://reviews.delphix.com/r/42619/) for more context.

live-build/misc/upgrade-scripts/verify
pzakha: You are right, it's probably better to have something like:
```
TOP="${BASH_SOURCE%/*}"
. "${TOP}/common.sh"
```

live-build/misc/upgrade-scripts/verify-container
prakashsurya: I'll look into this.

live-build/misc/upgrade-scripts/verify-container
pzakha: I see. This looks yucky. I feel like we should create DROPBOX_DIR as part of appliance-build.

live-build/misc/live-build-hooks/81-upgrade-repository.binary
prakashsurya: I'm OK with whatever makes sense. Let me know what name you'd like to use, and I'll change it.

FWIW, we easily change this later, as the consumers of this image (e.g. `dx_unpack`) don't rely on the filename.

pzakha: > as the consumers of this image (e.g. dx_unpack) don't rely on the filename.

Ah, I wasn't aware of this. 
I think `.upgrade.tar.gz` would be good.

live-build/misc/upgrade-scripts/verify-container
prakashsurya: In the `minimal-common` ansible role? or where are you thinking?

pzakha: I think that would be a reasonable place to do it.

prakashsurya:
`git-ab-pre-push` for this next iteration is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build-orchestrator-pre-push/59/)

prakashsurya:
`git-ab-pre-push` for this next iteration is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build-orchestrator-pre-push/61/)

prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build-orchestrator-pre-push/65/)

PR #91, opened 2018-08-24
Increase ZFS LUA memory limit to 512MB

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build-orchestrator-pre-push/69/)

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -363,14 +363,28 @@
...
+    # we're running a rootfs on ZFS.
+    #
+    - "options spl spl_panic_halt=1"
+    #
+    # We increase the max memory limit for ZFS channel programs to 512MB
+    # to give us a wide buffer for development of channel programs. We
+    # were using upwards of 70MB during internal testing, which is too
+    # much with the default limit of 100MB.
+    #
+    - "options zfs zfs_lua_max_memlimit=536870912"
```
ahrens: We can change tunables either by modifying the conf file (as you're doing here) or by changing the source code (of github/delphix/zfs).  We've done the latter for some other tunables.  How should we determine which mechanism to use?

prakashsurya: I’m OK with either approach. I asked on Slack, and was directed to this approach. 

@ahrens let me know if you’d rather us make this change in the ZFS code, instead.

prakashsurya: cc @gwilson

ahrens: We've made these tunable changes in delphix/zfs that are not present upstream (zfsonlinux/zfs).  We should make all the tunable changes in one place, or have some criteria for determining where to make them.

```
+int zfs_dmu_offset_next_sync = 1
+int zfs_mg_noalloc_threshold = 5;
+int spa_asize_inflation = SPA_DVAS_PER_BP + 1;
+uint32_t zfs_vdev_sync_read_min_active = 35;
+uint32_t zfs_vdev_sync_read_max_active = 35;
+uint32_t zfs_vdev_sync_write_min_active = 35;
+uint32_t zfs_vdev_sync_write_max_active = 35;
+uint32_t zfs_vdev_async_read_max_active = 10;
+uint32_t zfs_vdev_async_write_min_active = 1;
+int zfs_vdev_aggregation_limit = SPA_OLD_MAXBLOCKSIZE;
```

grwilson: I also started to make these change in `delphix-config` so we should figure out where the final home for these changes should be.

prakashsurya: Are you suggesting we should opt to go the route of using this config file as a general rule of thumb?

If so, I can open another PR to pull in those settings as well. And then another PR to zfs to revert back to defaults.

Personally, I like this config file approach since it makes it clear which tunables we change, but I’d be OK going the route of making the change in our ZFS too.

Which approach do you prefer? I can’t quite tell so far..

prakashsurya: @gwilson the problem with doing this in delphix-config, is it then requires a reboot (and rebuilding the initramfs) to take effect.

prakashsurya:
Verified this change got applied from the `git-ab-pre-push` run above:
```
$ cat /etc/modprobe.d/zfs.conf
options spl spl_panic_halt=1
options zfs zfs_lua_max_memlimit=536870912
$ sudo cat /sys/module/zfs/parameters/zfs_lua_max_memlimit
536870912
```

prakashsurya:
@ahrens @grwilson are you two OK with this approach, or should I close it and make this change in the ZFS codebase?

ahrens:
I'm OK with either approach, I'd just like us to standardize on one.

prakashsurya:
I've opened LX-1301 to standardize on using the configuration file. We can use that issue to convert the existing tunable configurations to use this file, so we're consistent across our tunable changes.

Is there a good place to document this decision? This way, we'll have docs to point folks at, that explain how to make tunable changes (i.e. edit this config file, rather than editing the zfs code)?

prakashsurya:
This last commit does not bring in the following tunables into the configuration file:
```
+int spa_asize_inflation = SPA_DVAS_PER_BP + 1;
+int zfs_vdev_aggregation_limit = SPA_OLD_MAXBLOCKSIZE;
```
Since these are based on macros, do we want to keep them tuned in the ZFS code base? That way, if the macro values change, so will the tunable value?

If not, I'll determine the current value for these and pull them into the modprobe configuration file, just like the others.

grwilson:
I think we should have a tunable file (i.e. 10-zfs.conf) that provides our customized changes (including macros). The file should include a comment indicating that support should create a secondary file (20-zfs.conf) that would run after the default one for any changes they need to override.

prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build-orchestrator-pre-push/94/)

prakashsurya:
I cloned the image produced by `git-ab-pre-push` and verified the file `/etc/modprobe.d/10-zfs.conf` was as expected and the SPL/ZFS kernel module parameter values from that file were in effect.

@ahrens @grwilson Could you take another look at this. I think it's ready to land. Once we integrate this, I can open a PR on the ZFS project to revert our tunable changes done in the code, back to the default values.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -369,13 +369,57 @@
...
+      # On Linux, SEEK_FILE and SEEK_HOLE are inaccurate if the file is
+      # dirty; in which case, they say that there are no holes. To get the
+      # same behavior as illumos, we need to configure this tunable.
+      #
+      options zfs zfs_dmu_offset_next_sync=1
+
+      #
+      # The following changes are made to match the values for these same
+      # tunables, that we use on our illumos-based DelphixOS.
+      #
```
ahrens: We'll probably want to improve this comment at some point, since it basically is just saying "this is the way we've always done it".  But this is fine for now.

ahrens:
Thanks for doing this, and (in advance) for reverting the changes from the delphix/zfs repo.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
prakashsurya: Yea. I just pulled this comment from the commit message on the zfsonlinux repository that made the change.

Since I'm already here, I'll do a little more legwork and look at the commit message and/or bug for our illumos repository, and see if I can add some more details here.

PR #98, opened 2018-08-30
Add passwordless sudo for LDAP users on internal-dev appliances

jgallag88:
This works on Illumos appliances, so developers will expect it to work
on Linux as well. Here we do something slightly better than what we do
on Illumos by restricting passwordless sudo to LDAP users, which makes
it less likely that users will overlook bugs where we accidentally rely
on these extra privileges which don't exist on the version we ship.

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.delphix-ldap/files/80-internal-ldap
+++ b/live-build/misc/ansible-roles/appliance-build.delphix-ldap/files/80-internal-ldap
@@ -1,0 +1,10 @@
+#
+# For convenience, allow passwordless sudo for all ldap users on internal
+# development appliances. Delphix ldap users are in a group with gid 10 because
+# on Illumos gid 10 is assigned to the group staff. On Linux, gid 10 is assigned
+# to the group uucp, so a side effect of this entry is that passwordless sudo is
+# allowed for uucp. We can fix this eventually if we create a new ldap group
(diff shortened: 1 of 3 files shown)


jgallag88:
ab-pre-push: [http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/dx-integration-tests/9249/](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/dx-integration-tests/9249/)

basil:
bors delegate+

prakashsurya:
@jgallag88 is there anything holding this up, other than time/priorities? just wondering if you tested and found this not to work.

jgallag88:
@prakashsurya I refactored the change per your suggestion, and everything seems to be working as expected. I was just waiting on LDAP to be fixed to do the final sanity check and update the PR.

jgallag88:
Per @prakashsurya's suggestion, I've re-arranged things so that this logic only applies to the `internal-dev` variant, instead of all internal variants.

Testing:
- ab-pre-push:  http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/5/
 - logged into an internal-dev vm built with this change and made sure that passwordless sudo worked for my ldap user.
 - logged into an internal-qa vm built with this change and made sure that the `80-internal-ldap` file was not present in `/etc/sudoers.d/`

PR #104, opened 2018-09-11
LX-1220 CLI PAM module fails to use Bcrypt when stack is down (bcrypt dev package)

rasantel:
We need to fix authentication in the CLI in Linux when the management stack is down. This requires using bcrypt to hash and compare the password with what's stored in the enabled users cache.

The `libcrypt-blowfish` and `libcrypt-blowfish-dev` packages are already published in artifactory.

This review adds our artifactory Debian repo to the list of apt sources. It also ensures the dev package (the header files) are available in virtualization development appliances.

Test run: http://selfservice.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build-orchestrator-pre-push/125
(there are unexpected Linux test results but this verifies that the Debian repo is added and that the dev package gets installed)


rasantel:
Addressed @prakashsurya 's comment.
New test run: http://selfservice.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build-orchestrator-pre-push/131

PR #105, opened 2018-09-11
Temporarily install build dependencies in Virtualization Engine product variants

basil:
As the comment states, we would like to use the `internal-qa` variant for pre-push testing; however, in order to do that, we would currently need to revamp our entire pre-push testing workflow. As a workaround, we temporarily include the build dependencies in the `internal-qa` variant to avoid having to revamp the pre-push testing workflow. Once the pre-push testing workflow is redone, we can move the build dependencies back into the `internal-dev` variant.


basil:
`git ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/projects/job/dx4linux/job/appliance-build-orchestrator-pre-push/126/).

PR #113, opened 2018-09-18
Set tunable to enable accurate zfs send size estimates

pcd1193182:
This change can safely land before the corresponding change to zfs, since the property will be silently ignore if it isn't valid.


live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -451,6 +451,14 @@
...
       #
       options zfs zfs_mg_noalloc_threshold=5
+
+      #
+      # Our data blocks are of size 8k, but the filesystems usually have a
+      # recordsize of 128k. That causes incorrect size estimates in zfs
+      # send. Overriding the property here fixes that.
+      #
+      options zfs zfs_override_estimate_recordsize=8192
+
```
prakashsurya: Looks like the linter doesn't like this newline.

PR #119, opened 2018-09-26
LX-199 linux CRA implementation (appliance-build part)

rasantel:
Install CRA PAM module. By default, it is not enabled. The corresponding app gate change will include an updated script to allow support to enable and disabling CRA, just like in illumos.

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
@@ -544,3 +544,10 @@
     state: directory
     mode: 0755
     recurse: yes
+
+#
+# Install, but do not enable, the challenge-response PAM module for support
+# logins with the 'delphix' user.
+#
+- apt:
+    name: 'pam-challenge-response'


rasantel:
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/39/flowGraphTable/
The integration test failure was due to LX-1350. The blackbox failure seems to be an infrastructure issue. Re-running all Linux tests after rebasing on top of the LX-1350 fix: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/dx-integration-tests/9799

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -544,3 +544,10 @@
     state: directory
     mode: 0755
     recurse: yes
+
+#
+# Install, but do not enable, the challenge-response PAM module for support
+# logins with the 'delphix' user.
+#
+- apt:
+    name: 'pam-challenge-response'
```
prakashsurya: This is only required by virtualization, right? i.e. if virtualization isn’t installed, do we still want this package?

If so, let’s put it in the “minimal” package list. If not, let’s put it as a virtualization package dependency.

rasantel: It's not only for virtualization. CRA is for all engines -- for the support (`delphix`) user.

prakashsurya: Sure.. But it can only be enabled and used with the accompanying logic from the virtualization package, right?

That’s why I think this should be listed as a dependency of the virtualization package. 

If I’m wrong, then let’s add this to the minimal package list (i.e. [here](https://github.com/delphix/appliance-build/blob/master/live-build/base/config/package-lists/minimal.list.chroot)), rather than installed here via ansible.

rasantel: You are right. We need CRA to be available for all engines though. Having CRA in masking-only engine was not an issue in our illumos distribution since the related app gate scripts are present there, but I realize that this won't be the case for Linux. The real solution here is to package and deploy those scripts separately from virtualization.

rasantel: I will make this change in my `linux-dlpx-pkgs` and `dlpx-app-gate` reviews.

prakashsurya: > Having CRA in masking-only engine was not an issue in our illumos distribution since the related app gate scripts are present there, but I realize that this won't be the case for Linux.

This will be true for Linux too; we'll always ship the virtualization package on the appliance along with the masking package. Even for "masking only" customers, the virtualization application is still required (e.g. for initial setup).

rasantel:
Closed. Will add CRA PAM as a dependency of virtualization instead, as @prakashsurya suggested.

PR #121, opened 2018-10-01
LX-1427 linux engines should use tmpfs

grwilson:


live-build/misc/live-build-hooks/90-raw-disk-image.binary
```diff
@@ -123,6 +123,11 @@ zfs create \
 #
 rsync --info=stats3 -Wa binary/* /mnt/
 
+#
+# Setup tmpfs for /tmp
+#
+echo "tmpfs /tmp tmpfs nodev,nosuid,size=4G 0 0" >> /mnt/etc/fstab
```
prakashsurya: Why'd you choose to make this change here? Would it be more appropriate, and perhaps more elegant/maintainable, to do this configuration from the `appliance-build.minimal-common` role? you could use the ansible [mount module](https://docs.ansible.com/ansible/2.6/modules/mount_module.html) to do it that way.

grwilson: Originally, this is were we did the swap addition so I just added on to that. I'll look at the new location. Thanks for the input.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -544,3 +544,13 @@
     state: directory
     mode: 0755
     recurse: yes
+
+#
+# Setup tmpfs for /tmp
```
prakashsurya: this comment isn't helpful; we should either remove it, or have it complement the code by explaining _why_ we want `/tmp` to be `tmpfs`.

prakashsurya:
bors delegate+

PR #125, opened 2018-10-05
Use "-t" when creating each variant's rpool

prakashsurya:
Currently we can't reliably make use of the "-j" option to "make", to
run multiple variant builds concurrently. This is because when rpool is
created for a given variant, the rpool for another variant can't be
imported at the same time. This is because both pools will share the
same pool name, and ZFS prevents two pools with the same name from being
imported simultaneously.

To fix this, this change attempts to use the "-t" option when creating
the variant rpool, such that we can have multiple variant rpools
imported at the same time. This way, one can use "-j" with "make" to
concurrently build multiple variants, improving overall build latency.


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/98/)

PR #130, opened 2018-10-12
Install new version of delphix-entire on upgrade

prakashsurya:
The purpose of this change is to replace the use of "dist-upgrade" on
upgrade, with installing the specific new version of "delphix-entire".
This way, we're more sure of exactly what packages (and the specific
versions of those packages) we'll end up with after the upgrade.

### DIFF
--- a/live-build/misc/upgrade-scripts/execute
+++ b/live-build/misc/upgrade-scripts/execute
@@ -17,6 +17,8 @@
 
 . "${BASH_SOURCE%/*}/common.sh"
 
+set -o pipefail
+
 function usage() {
 	echo "$(basename "$0"): $*" >&2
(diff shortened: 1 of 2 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/121/)

I also manually tested this change by:
1. creating a `dlpx-linux-trunk` VM
2. downloading and unpacking the latest upgrade image
3. manually copying the `execute` script from this change
4. running this new `execute` script to upgrade my VM

prakashsurya:
cc: @grwilson @pzakha @dtzeng

PR #132, opened 2018-10-19
LX-1418 Bind mount /var/delphix/server/upgrade-verify to save logs fr…

pandar00:
…om upgrade verify

**Problem**
* upgrade-verify.jar logs are saved within the container at `/var/delphix/server/upgrade-verify/`. However, because this directory is created within the container and purged along with container cleanup code, we cannot inspect it for debugging.

**Possible solutions**
1. Expose `/var/delphix/server/upgrade-verify` from the container's host such that files are not purged when container gets destroyed.
2. Disable container cleanup code if it fails. Currently we only keep the failed container only if `$DLPX_DEBUG` option is set.

**Fix**
* Went with option 1) since it's the easiest solution. Also, option 2) might give more space overhead on customer's environment.

**Tests**
* Log files are now saved in the container's host

```
delphix@harry-dlpx-linux-trunk:/var/delphix/server/upgrade-verify$ ls -la
total 14
drwxr-xr-x 2 root root    8 Oct 19 00:51 .
drwxr-xr-x 8 root root   10 Oct 19 00:47 ..
-rw-r--r-- 1 root root 7573 Oct 19 00:51 debug.log
-rw-r--r-- 1 root root    0 Oct 19 00:51 error.log
-rw-r--r-- 1 root root    0 Oct 19 00:51 info.log
-rw-r--r-- 1 root root    0 Oct 19 00:51 opensource.log
-rw-r--r-- 1 root root    0 Oct 19 00:51 sshj.log
-rw-r--r-- 1 root root    0 Oct 19 00:51 trace.log
```

### DIFF
--- a/live-build/misc/upgrade-scripts/verify
+++ b/live-build/misc/upgrade-scripts/verify
@@ -18,6 +18,8 @@
 TOP="${BASH_SOURCE%/*}"
 . "$TOP/common.sh"
 
+UPGRADE_VERIFY_CONTAINER_LOG_DIR=/var/delphix/server/upgrade-verify
+
 function usage() {
 	echo "$(basename "$0"): $*" >&2
 	echo "Usage: $(basename "$0") -v <version>"
@@ -26,6 +28,23 @@ function usage() {
 
 function report_progress_inc() {
 	echo "Progress increment: $(date +%T:%N%z), $1, $2"
+}
+
+function collect_logs() {
+	verify_logs_dest="$DROPBOX_DIR/verify_logs"
+
+	if [[ -d "$verify_logs_dest" ]]; then
+		rm -rf "$verify_logs_dest"
+	else
+		mkdir "$verify_logs_dest"
+	fi
+
+	for log_dir in "$@"; do
+		if [[ -d "$log_dir" ]]; then
+			cp -r "$log_dir" "$verify_logs_dest"
+		fi
+	done
+
 }
 
 function cleanup() {
@@ -42,6 +61,15 @@ function cleanup() {
 	# further cleanup for us to do here.
 	#
 	[[ -z "$CONTAINER" ]] && return
+
+	#
+	# Before cleaning the container, we need to copy logs to the
+	# container's host for debugging purpose. Otherwise the logs will
+	# be cleaned up as part of container deletion.
+	#
+	report_progress_inc 70 "Collecting verification logs"
+	local logs_dirs_to_copy="$UPGRADE_VERIFY_CONTAINER_LOG_DIR"
+	collect_logs $logs_dirs_to_copy
 
 	#
 	# On failure, and when DLPX_DEBUG is true, we avoid cleaning up
@@ -105,7 +133,7 @@ fi
 
 systemd-run --machine="$CONTAINER" --quiet --pipe --wait -- \
 	/usr/bin/java \
-	-Dlog.dir=/var/delphix/server/upgrade-verify \
+	-Dlog.dir="$UPGRADE_VERIFY_CONTAINER_LOG_DIR" \
 	-Dmdsverify=true \
 	$VERIFY_DEBUG_OPT \
 	-jar /opt/delphix/server/lib/exec/upgrade-verify/upgrade-verify.jar \


pandar00:
cc: @prakashsurya , @dtzeng

pandar00:
I realized after finishing is that shouldn't we save logs from other services as well?
e.g.) Postgres service log or /var/delphix/server/log if we start testing stack startup

prakashsurya:
The code looks good, but I think we could improve on things by exposing a single directory, and then having the Java class handle copying everything we want over to that directory. This way, we don’t have to have multiple bond mounts for all the things we may want to save.

I added the dropbox mount simply because that was easiest at the time, and didn’t require application changes. Now that we’re making application changes, I think it’d be best to modify the application to work better in this verify environment.

What do you think?

pandar00:
Sounds good to me. Let me scope out the required changes.

pandar00:
@prakashsurya , @dtzeng 

I think it's more appropriate to rephrase the bug as "verification container logs are not saved once container is purged". It turns out that there are more logs we want to collect from the verification container in addition to `upgrade-verify.jar` such as Postgres logs, java stack logs (once we start adding post upgrade startup verification) or any other necessary logs to help debugging if verification fails. (Btw. There's a separate task deciding what logs other than `upgrade-verify.jar` so this review only includes `upgrade-verify.jar` log.)

With that said, I reverted the original change and added log collection logic in the `verify` script as part of cleanup.

There are few things I discussed with @dtzeng before making the change:

1. Do we want to collect verify logs per container? Current implementation only saves the last executed container.
   * Collecting logs per container implies we should implement some kind of cleanup mechanism. One that @dtzeng  suggested was implementing some kind of monitor that purges log directories based on some condition such as disk space or timestamp. 
2. What if log files are large? Is compression an option?
   * Compression will be needed if log file sizes are large. Upon testing against longevity machine, we found out that `/var/delphix/server/log` takes `466M`, compressed one takes `252M` and compression time takes about 3 minutes.
   * We actually might not want to compress since support bundle is already compressed and includes the contents of `DROPBOX_DIR`.
   * Ideally, we should only collect logs generated from the start of the container, excluding already existing ones. Doing so in Java stack is pretty straightforward by deleting the logs directory and then starting the Java stack, but not sure how it'd behave if we decided to collect system logs (`/var/logs`). By the way, the system logs takes `460M` on longevity machine. Anyhow, more investigation is needed to achieve this.
3. Instead of log collection, how about we leave the container running?
   * This is most likely infeasible in customer environment in terms of both disk space and compute resource (CPU/RAM).
       * Container creates a new snapshot of the root dataset and the longer we keep that snapshot the more it diverges from the parent and takes more space. (Please correct me if I'm wrong here)
       * For obvious reasons, running container takes compute resources.

Due to following reasons, I decided collecting _one set_ of verification container logs and add the collection logic to cleanup script is probably the simplest solution for now.

Also here are few extra things I though worth sharing.
* The `DROPBOX_DIR` is used by `WindowsDebugService` class in the app stack to save some diagnostic data.
* The contents of `DROPBOX_DIR` is copied to support bundle in `support_info.sh` script and the entire directory is **deleted and recreated**.

What do you think about the new changes? I've update the code.

pandar00:
The current support bundle collection script (i.e. `support_info.sh`) sort of already does that. User can specify one of these options as an API parameter:
> "PHONEHOME", "MDS", "OS", "CORE", "LOG", "DROPBOX", "STORAGE_TEST", "MASKING", "ALL"

The verification logs would fall into `DROPBOX`. 

However, in order for the script to collect verification logs, those logs first need to be saved in the container's host environment, which is what the current review addresses. All verification logs will be located in `UPGRADE_VERIFY_CONTAINER_LOG_DIR=/var/delphix/server/upgrade-verify`.

prakashsurya:
What I am thinking, is that we could run `support_info.sh` from inside the upgrade/verification container, as a way to "collect all the information we may need on failure". The tarball generated by that script, could then be copied to someplace that'll preserve it after the container is destroyed.

This way, we have a single script that is used to "collect information on failure", that works for both the "regular system" and the "verification container". This would solve the issue of us needed to collect things like postgres logs from the verification container on failure, and anything else we may need.

`support_info.sh` could also be modified to perform conditional logic based on if it's running from inside the container or not, using `systemd-detect-virt`, if we need to do this.

I'm just thinking that having a single way to collect information on failure would be good, and using `support_info.sh` for inside and outside of the container would be a good way to do this. When run inside of the container, the generated tarball could be stored in a bind mounted directory (e.g. like `DROPBOX_DIR`), and then when it's run outside of the container, it'd collect these container tarballs.

Does this make sense?

I'm OK moving forward with the current changes, if this idea is better left for a different PR.

pandar00:
I see your point now, but I'd prefer deferring that task to a future PR since it needs more investigation. 

Few things to consider for future reference:
* Is `support_info.sh` written in a fault-tolerant way? (e.g. if a stack is down but the log collector requires stack API, will it move on to a next log collection mechanism?)
* `support_info.sh` should _not_ purge `<DROPBOX_DIR>` since it's used by container host as well.
* Calling `support_info.sh` within container means will collect redundant logs since the container will carry over existing logs from the host. This may be OK if we only save one set of verification container logs, but what if we start supporting multiple?

I filed a tracking ticket here: https://jira.delphix.com/browse/LX-1541

live-build/misc/upgrade-scripts/verify
```diff
@@ -105,7 +133,7 @@ fi
 
 systemd-run --machine="$CONTAINER" --quiet --pipe --wait -- \
 	/usr/bin/java \
-	-Dlog.dir=/var/delphix/server/upgrade-verify \
+	-Dlog.dir="$UPGRADE_VERIFY_CONTAINER_LOG_DIR" \
```
prakashsurya: Rather than having to add the additional "copy" step to the "cleanup" function, can we instead just set:
```
-Dlog.dir="$DROPBOX_DIR/verify_logs
```
?

If so, that seems like it'd be more simple of a solution, and we wouldn't have to perform any copy operations on cleanup.

pandar00: I added the "copy" logic to easily add more log directories to copy if we needed. Users can add additional directories to `logs_dirs_to_copy` variable. Postgres log, for example, is one of them but I didn't include it to this review since I needed more info on what exact logs we need to copy.

Also, I think the "copy" logic is necessary if we do a verification logic that doesn't allow specifying log output directory but want to save those for debugging purpose. For example, `dpkg.log` maybe?

I think the logic is fairly clean as it is: there is one place to store logs and one place to copy.

Just for additional info, I introduced `$DROPBOX_DIR/verify_logs` to isolate verification logs from other irrelevant files in `$DROPBOX_DIR` directory because it's used by other services as well (`WindowsDebugService` for example), but unfortunately `support_info.sh` removes and recreates the root of `$DROPBOX_DIR` which doesn't include creating nested directories (This makes sense (kind of..) since there is no way to find out container structure of `$DROPBOX` dir), so I create it just before the copy logic such that precondition is guaranteed to satisfy. (Just for additional info, `log.dir` is used by Log4J but I don't think it creates the directory automatically if I'm not wrong).

If you still strongly feel it's not necessary, then I can make the changes.

prakashsurya:
@pandar00 When you're ready for this to land, leave a comment of "bors r+" and the bot will do the work of merging this. Please get a successful run of `git-ab-pre-push` before doing so (that doesn't really test this change, but still a good habit to do this). You'll also need to accept/sign the CLA.

pandar00:
[`git-ab-pre-push` link is here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/150/console)

pandar00:
Closing the PR for now. Looking into leveraging `support_info` script to collect logs within verify container as suggested during the review.

PR #134, opened 2018-10-23
LX-1415 Investigate automatic stack restart after deferred upgrade

pandar00:
**Changes**

* Adds a logic to automatically restart `delphix-platform` service after upgrading packages. (I.e. deferred upgrade)
* In case of full-upgrade, we do **not** want to restart `delphix-platform`. As of now, full-upgrade is not supported and will be addressed with [LX-224](https://jira.delphix.com/browse/LX-224) once full-upgrade is implemented.

**Testing**

* [`git-ab-pre-push`](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/152/)
* Modified `execute` script of an upgrade image and ran it.
    * Checked that `dpkg -l` correctly shows that packages are updated
    * Checked that `systemctl status delphix-mgmt` correctly shows that the service is restarted.

### DIFF
--- a/live-build/misc/upgrade-scripts/execute
+++ b/live-build/misc/upgrade-scripts/execute
@@ -71,4 +71,8 @@ CANDIDATE_VERSION=$(pkg_version_candidate "delphix-entire")
 apt_get install -y "delphix-entire=$CANDIDATE_VERSION" ||
 	die "upgrade failed; from '$INSTALLED_VERSION' to '$CANDIDATE_VERSION'"
 
+# We treat all upgrades as deferred for now. LX-224 addresses reboot upgrade.
+systemctl restart delphix-platform ||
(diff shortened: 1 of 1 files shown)


prakashsurya:
> Just a side thought that I had, what if we just had the Java stack do the work of restarting delphix-platform or the VM after this script is run, instead of this script doing the work?

I thought this was going to be the approach taken (doing the restart in Java), but I'm fine either way. Both approach seem like equally valid ways to solve the problem, so I don't have a preference (at least not yet).

pandar00:
OK, I'll look into Java approach.

pandar00:
:open_mouth: Didn't know `SystemService` existed! I'm more than happy to move the restart logic into Java stack. 

Closing the PR.

PR #135, opened 2018-10-24
Allow building with LDAP users

jgallag88:
Our LDAP users are now members of the staff group. The files in the git
repos cloned on the internal-dev variant are already owned by the staff
group, but they need to be group writable in order to allow LDAP users
to push changes.


jgallag88:
### Testing
ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/153/flowGraphTable/

Successfully ran both `git-appliance-update` and `git-masking-update` using my LDAP user on the appliance produced by ab-pre-push.

PR #136, opened 2018-10-29
Use cloud-init to set APT sources for dev variant

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/159/)

live-build/misc/ansible-roles/appliance-build.minimal-development/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
...
+# limitations under the License.
+#
+
+---
+- copy:
+    dest: /etc/cloud/cloud.cfg.d/99-delphix-development.cfg
+    mode: 0644
+    content: |
+      apt:
+        preserve_sources_list: false
```
pandar00: Just a side note: Looks like this is set to `false` by default.

prakashsurya: right.. so, in my platform change [here](https://github.com/delphix/delphix-platform/pull/11), we'll start setting that to true by default (i.e. we don't want use ubuntu's repos for customer installations or our QA variant), and then we'll override that setting for only the developer variant (via this change).

PR #139, opened 2018-10-30
Add "not-in-place" option on upgrade container creation

prakashsurya:
This change modifies the script that's used to create the container used
for upgrade verification, adding a new "not-in-place" mode. Thus, now
there's two modes: "in-place" and "not-in-place".

When creating the container "in-place", the container will be created from
a snapshot of the current root filesystem. This mode is the same as what
was previously supported. The result of this, will be a new container with
all of the same software installed as the current root filesystem.

When creating the container "not-in-place", the container will be created
from scratch by using debootstrap. The result of this, will be a new
container will only a very minimal set of software installed, such that the
container can be started and the "delphix-entire" package can be installed.

While the "not-in-place" mode is not currently used on upgrade (we still
only use the "in-place" container creation support), the intention is to
build on this, and eventually allow upgrading to a new root filesystem that
is not a clone of the current root filesystem.


prakashsurya:
cc: @dtzeng @pandar00

`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/173/)

prakashsurya:
As verification, I ran the following using the latest upgrade image + the changes in this PR:
```
$ sudo /var/dlpx-update/latest/upgrade-container create not-in-place latest
...
not-in-place.d4YOt3Q
$ sudo /var/dlpx-update/latest/upgrade-container start not-in-place.d4YOt3Q
$ sudo systemd-run --machine not-in-place.d4YOt3Q --quiet --pipe --wait -- /var/dlpx-update/latest/execute latest
```

live-build/misc/upgrade-scripts/upgrade-container
```diff
@@ -164,6 +143,70 @@ function create() {
...
+	# name and then later start/stop/destroy the container. Thus, we
+	# have to redirect the output from debootstrap away from stdout.
+	#
+	# Also, we need to include the "systemd-container" package when
+	# installing the base systemd with debootstrap so that starting
+	# the container will work properly. Otherwise, after starting
+	# the container, we won't be able to communicate and later run
+	# commands in the container with "systemd-run".
+	#
+	debootstrap --no-check-gpg \
```
pzakha: Have you looked at `--variant`? Do we want to use `minbase`?

prakashsurya: I did, but now I don't recall the details. I could be mistaken, but I think the problem with that, is it doesn't create a system that can be booted with `systemd-nspawn`.

So, we could do that, then `chroot` in and install `delphix-entire`, without using `systemd-nspawn`, but I ran into issues where `delphix-entire` would fail to install in the `chroot` (i.e. dependent packages would fail to install).. this could have been due to not having things like `/sys`, `/proc`, etc. bind mounted, but I didn't try to root case that.

I opted for this approach since it integrates well into our existing code for creating the verification container using `systemd-nspawn`.

prakashsurya:
> Can we use a debootstrap package from an earlier LTS version to bootstrap a later LTS version? (e.g. can bootstrap from 16.04 create an 18.04 image?)

Yes, I didn't test, but this should definitely be possible; `debootstrap` should be able to be used from a non-ubuntu system too (that's what their docs say).

> What happens with the /boot directory? Since the grub config is written there, I'm still not sure how does grub determine which dataset to read grub.cfg from.

This change doesn't do anything to grub, so we won't boot the new dataset as-is. I'm still investigating this part, and that'll come in another change. I have verified that after I create the dataset, I can manually edit things like `mountpoint` and `grub` and get the new dataset to boot, though; I just need to figure out how best to do that from the scripts (e.g. use bootfs, or chroot + update-grub + grub-install, or something else).

PR #142, opened 2018-11-01
Infer version from path of script being executed

prakashsurya:
This changes the interface of the upgrade scripts such that the version
being used for the upgrade or verification does not have to be specified
as an option/parameter to the script. Rather, we now infer the version
from the path of the script being executed, since all of these scripts
will reside in a version specific subdirectory of "/var/dlpx-update".

This eliminates the inconsistency between the scripts, where some would
require the version to be specified using "-v", and others would require
the version to be passed as a positional parameter.

### DIFF
--- a/live-build/misc/upgrade-scripts/common.sh
+++ b/live-build/misc/upgrade-scripts/common.sh
@@ -22,3 +22,11 @@ function die() {
 	echo "$(basename "$0"): $*" >&2
 	exit 1
 }
+
+function get_image_path() {
+	readlink -f "${BASH_SOURCE%/*}"
+}
+
+function get_image_version() {
+	basename "$(get_image_path)"
+}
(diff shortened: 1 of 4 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/184/)

Ran the following commands with this changes in place:
```
$ sudo /var/dlpx-update/latest/verify
Progress increment: 19:12:44:442056061+0000, 0, Creating upgrade verification container
Progress increment: 19:12:44:551966534+0000, 20, Starting upgrade verification container
Progress increment: 19:13:31:954089246+0000, 40, Performing package upgrade verification
...
$ sudo /var/dlpx-update/latest/execute
...
```

prakashsurya:
cc: @dtzeng @pandar00 @grwilson @pzakha

PR #143, opened 2018-11-01
Add "upgrade" script that can upgrade an appliance

prakashsurya:
This change adds a new "upgrade" script to the set of upgrade-scripts. This
new script is intended to be used to upgrade an existing appliance to the
version contained in the upgrade image; it script supports both,
"in-place" and "not-in-place" upgrades.

### DIFF
--- a/live-build/misc/upgrade-scripts/upgrade
+++ b/live-build/misc/upgrade-scripts/upgrade
@@ -1,0 +1,198 @@
+#!/bin/bash
+#
+# Copyright 2018 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+. "${BASH_SOURCE%/*}/common.sh"
+
+IMAGE_PATH=$(get_image_path)
+[[ -n "$IMAGE_PATH" ]] || die "failed to determine image path"
+
+#
+# This variable is used to determine if the application verification
+# checks should be run as part of the upgrade. The default value can be
+# overridden by setting this environment variable to "true", and/or
+# passing the "-v" option to this script (see the scripts usage). If
+# both, the environment variable and the "-v" option are specified, the
+# "-v" option takes precedence.
+#
+DLPX_UPGRADE_SKIP_VERIFY=${DLPX_UPGRADE_SKIP_VERIFY:-"false"}
+
+function usage() {
+	echo "$(basename "$0"): $*" >&2
+
+	PREFIX_STRING="Usage: $(basename "$0")"
+	PREFIX_NCHARS=$(echo -n "$PREFIX_STRING" | wc -c)
+	PREFIX_SPACES=$(printf "%.s " $(seq "$PREFIX_NCHARS"))
+
+	echo "$PREFIX_STRING [-v] in-place"
+	echo "$PREFIX_SPACES [-v] not-in-place"
+
+	exit 2
+}
+
+function cleanup_container() {
+	#
+	# If the CONTAINER variable is empty, this means container
+	# creation failed. In that case, the container creation process
+	# will have already cleaned up after itself, and there's no
+	# further cleanup for us to do here.
+	#
+	[[ -z "$CONTAINER" ]] && return
+
+	"$IMAGE_PATH/upgrade-container" stop "$CONTAINER" ||
+		warn "failed to stop '$CONTAINER'"
+
+	"$IMAGE_PATH/upgrade-container" destroy "$CONTAINER" ||
+		warn "failed to destroy '$CONTAINER'"
+}
+
+function cleanup_in_place_upgrade() {
+	#
+	# Capture the exit code here, and use that to determine if
+	# upgrade verification was successful or not.
+	#
+	local rc="$?"
+
+	if [[ "$rc" == "0" || "$DLPX_DEBUG" != "true" ]]; then
+		#
+		# On success, or on failure when DLPX_DEBUG is not true,
+		# we clean up the container previously created. When
+		# DLPX_DEBUG is true, we leave the container around on
+		# failure to aid debugging efforts.
+		#
+		cleanup_container
+	fi
+
+	return "$rc"
+}
+
+function upgrade_in_place() {
+	trap cleanup_in_place_upgrade EXIT
+
+	CONTAINER=$("$IMAGE_PATH/upgrade-container" create in-place)
+	[[ -n "$CONTAINER" ]] || die "failed to create container"
+
+	"$IMAGE_PATH/upgrade-container" start "$CONTAINER" ||
+		die "failed to start '$CONTAINER'"
+
+	"$IMAGE_PATH/upgrade-container" run "$CONTAINER" \
+		"$IMAGE_PATH/execute" ||
+		die "'$IMAGE_PATH/execute' failed in '$CONTAINER'"
+
+	"$IMAGE_PATH/upgrade-container" run "$CONTAINER" \
+		/bin/systemctl start delphix-platform ||
+		die "'systemctl start delphix-platform' failed in '$CONTAINER'"
+
+	if [[ "$DLPX_UPGRADE_SKIP_VERIFY" != "true" ]]; then
+		"$IMAGE_PATH/upgrade-container" run "$CONTAINER" \
+			"$IMAGE_PATH/verify-impl" "$@" ||
+			die "'$IMAGE_PATH/verify-impl' '$*' failed in '$CONTAINER'"
+	fi
+
+	"$IMAGE_PATH/execute" ||
+		die "'$IMAGE_PATH/execute' failed in running appliance."
+}
+
+function cleanup_not_in_place_upgrade() {
+	#
+	# Capture the exit code here, and use that to determine if
+	# upgrade verification was successful or not.
+	#
+	local rc="$?"
+
+	if [[ "$rc" != "0" && "$DLPX_DEBUG" != "true" ]]; then
+		#
+		# On failure and when DLPX_DEBUG is not true, we clean
+		# up the container previously created.
+		#
+		# Unlike in-place upgrades, the container is not cleaned
+		# up on success; instead the container will have been
+		# already stopped and converted to be used as the root
+		# filesystem for the appliance after the next reboot.
+		#
+		cleanup_container
+	fi
+
+	return "$rc"
+}
+
+function upgrade_not_in_place() {
+	trap cleanup_not_in_place_upgrade EXIT
+
+	CONTAINER=$("$IMAGE_PATH/upgrade-container" create not-in-place)
+	[[ -n "$CONTAINER" ]] || die "failed to create upgrade container"
+
+	"$IMAGE_PATH/upgrade-container" start "$CONTAINER" ||
+		die "failed to start upgrade '$CONTAINER'"
+
+	"$IMAGE_PATH/upgrade-container" run "$CONTAINER" \
+		"$IMAGE_PATH/execute" ||
+		die "'$IMAGE_PATH/execute' failed in '$CONTAINER'"
+
+	"$IMAGE_PATH/upgrade-container" run "$CONTAINER" \
+		/bin/systemctl start delphix-platform ||
+		die "'systemctl start delphix-platform' failed in '$CONTAINER'"
+
+	if [[ "$DLPX_UPGRADE_SKIP_VERIFY" != "true" ]]; then
+		"$IMAGE_PATH/upgrade-container" run "$CONTAINER" \
+			"$IMAGE_PATH/verify-impl" "$@" ||
+			die "'$IMAGE_PATH/verify-impl' '$*' failed in '$CONTAINER'"
+	fi
+
+	"$IMAGE_PATH/upgrade-container" stop "$CONTAINER" ||
+		die "failed to stop '$CONTAINER'"
+
+	#
+	# After this point, we no longer want to execute the normal
+	# cleanup handler on any failure. The following command will
+	# convert the container to be used as the next boot/root
+	# filesystem, and thus the "stop" and "destroy" container logic
+	# will no longer work. So, if the "convert-to-bootfs" script
+	# fails, we'll need to manually rectify the situation.
+	#
+	trap - EXIT
+
+	"$IMAGE_PATH/upgrade-container" convert-to-bootfs "$CONTAINER" ||
+		die "failed to convert-to-bootfs '$CONTAINER'"
+}
+
+[[ "$EUID" -ne 0 ]] && die "must be run as root"
+
+while getopts ':v' c; do
+	case "$c" in
+	v)
+		DLPX_UPGRADE_SKIP_VERIFY="true"
+		;;
+	*)
+		usage "invalid option -- '$c'"
+		;;
+	esac
+done
+shift $((OPTIND - 1))
+
+case "$1" in
+in-place)
+	shift 1
+	upgrade_in_place "$@"
+	;;
+not-in-place)
+	shift 1
+	upgrade_not_in_place "$@"
+	;;
+*)
+	usage "invalid option -- '$1'"
+	;;
+esac
--- a/live-build/misc/upgrade-scripts/upgrade-container
+++ b/live-build/misc/upgrade-scripts/upgrade-container
@@ -146,6 +146,16 @@ function create_upgrade_container_common() {
 	echo "$CONTAINER"
 }
 
+function get_mounted_rootfs_name() {
+	local rootfs
+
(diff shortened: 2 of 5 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/185/)

This is dependent on #142.

I verified this by running the following with these changes:
```
$ sudo /var/dlpx-update/latest/upgrade in-place
...
$ sudo /var/dlpx-update/latest/upgrade not-in-place
...
```

prakashsurya:
cc: @dtzeng @pandar00 @pzakha @grwilson

live-build/misc/upgrade-scripts/upgrade
```diff
@@ -0,0 +1,167 @@
...
+	"$IMAGE_PATH/upgrade-container" start "$CONTAINER" ||
+		die "failed to start upgrade container '$CONTAINER'"
+
+	"$IMAGE_PATH/upgrade-container" run "$CONTAINER" \
+		"$IMAGE_PATH/execute" ||
+		die "'$IMAGE_PATH/execute' failed in container '$CONTAINER'"
+
+	"$IMAGE_PATH/upgrade-container" run "$CONTAINER" \
+		"$IMAGE_PATH/verify" "$@" ||
+		die "'$IMAGE_PATH/verify' '$*' failed in container '$CONTAINER'"
```
dtzeng: I don't think we would want to split the code for `verify` like this, the app-gate code would break
The app-gate wants to be able to just simply call a `verify` script that will take care of setting up all the container stuff and running the checks.  It looks like with this change we no longer have that capability anymore.

How about doing something like renaming your new `verify` to a different script?  And then use that as a building block for the old `verify` script and your new `upgrade` script.  That way we can keep this refactoring while maintaining the old functionality?

prakashsurya: Sure. We could do something like that, but we also need to think through how this will work for “not-in-place” upgrades and verification. 

I can definitely do what you suggest to avoid breaking the current app-gate code, but let’s discuss (perhaps at the next upgrade meeting) how we want this to work long term, considering both “in-place” and “not-in-place” upgrades; I think we want to be able to run the same verification logic is both cases.

For now, I’ll do as you suggest.

live-build/misc/upgrade-scripts/verify
```diff
@@ -20,18 +20,7 @@
...
-[[ -n "$IMAGE_VERSION" ]] || die "failed to determine image version"
-
-function usage() {
-	echo "$(basename "$0"): $*" >&2
-	echo "Usage: $(basename "$0") -v <version>"
-	exit 2
-}
-
-function report_progress_inc() {
-	echo "Progress increment: $(date +%T:%N%z), $1, $2"
```
prakashsurya: @dtzeng Also note, this change is removing the progress reporting output lines. I'm not convinced the current reporting is sufficient for what we'll eventually need on Linux, so I've opted to just remove it altogether for now, and plan to circle back to this problem later.

For example, there's certain commands that can take a long time to complete (e.g. `apt-get install`, `debootstrap`), so I'm not sure if the current approach of simply emitting progress lines in between each of the commands we run will be enough.

Let me know if you're OK removing these for now, or if you'd rather me add them back (even if they may not be sufficient when this is all done).

live-build/misc/upgrade-scripts/upgrade
prakashsurya: OK, I've created a new `verify-impl` script that's used by the new `upgrade` script; `verify` will create/destroy the verification container as needed, just like prior to this change. Let me know what you think when you get a chance to give this another look.

live-build/misc/upgrade-scripts/verify
dtzeng: I think we should actually put these back.  We don't really have any short-term plans to re-vamp any of the progress checks (at least for L.0).  And having some progress is better than nothing

Otherwise, this review looks good to me

live-build/misc/upgrade-scripts/upgrade-container
```diff
@@ -273,12 +273,79 @@ function run() {
...
+		umount "/var/lib/machines/${CONTAINER}${dir}" ||
+			warn "failed to umount '$dir' from '$CONTAINER'"
+	done
+}
+
+function convert_to_bootfs() {
+	trap convert_to_bootfs_cleanup EXIT
+
+	for dir in /dev /proc /run /sys; do
+		mount --bind "$dir" "/var/lib/machines/${CONTAINER}${dir}" ||
```
pzakha: mount `--bind` bounds non-recursively, which can mess up some devices. I think we should use `--rbind` and then `--make-rslave` to make sure we do not unmount those devices from the root environement.

Also, do we actually need to mount all of those devices?

live-build/misc/upgrade-scripts/upgrade-container
```diff
@@ -273,12 +273,79 @@ function run() {
...
+}
+
+function convert_to_bootfs_cleanup() {
+	for dir in /dev /proc /run /sys; do
+		umount "/var/lib/machines/${CONTAINER}${dir}" ||
+			warn "failed to umount '$dir' from '$CONTAINER'"
+	done
+}
+
+function convert_to_bootfs() {
```
pzakha: Can you elaborate a bit on what is the effect of this?
It seems to be updating and installing new grub. Also, perhaps `update-grub` would be a better name?

prakashsurya: > Can you elaborate a bit on what is the effect of this?

It's used to convert a upgrade container, to be used as the boot filesystem. It does this by doing things like: 1. update grub to boot the specified container's filesystem on the next boot, 2. update the container's dataset's mountpoint, etc.

> It seems to be updating and installing new grub. Also, perhaps update-grub would be a better name?

It does more than just updating grub though, it also has to change the mountpoint for the container's filesystem.

I chose this name because it takes an "upgrade container" as input, and converts this to be used as the new boot filesystem. I thought simplying naming it "update-grub" wouldn't be a little misleading.

Also, when the actual script is called, it's called like this:
```
$ .../update-container convert-to-bootfs $CONTAINER
```
i.e. it reads like, "convert the specified container to be the new boot filesystem"

which, to me at least, is better than:
```
$ .../update-container update-grub $CONTAINER
```

With that said, I'm fine naming things to whatever makes the most sense to you and everybody else. If you'd rather the option be named "update-grub" then I can make that change.

live-build/misc/upgrade-scripts/verify
prakashsurya: OK. I'll add it back.

live-build/misc/upgrade-scripts/upgrade-container
prakashsurya: > I think we should use `--rbind` and then `--make-rslave` to make sure we do not unmount those devices from the root environment.

Can you give me the commands that I should use to mount, and then the commands I should use to unmount?

I'm willing to try it out, I'm just not sure how to do it cleanly without the ugly complexity that we do for the raw image creation (which I'd like to avoid if possible), [here](https://github.com/delphix/appliance-build/blob/master/live-build/misc/live-build-hooks/90-raw-disk-image.binary#L147) and [here](https://github.com/delphix/appliance-build/blob/master/live-build/misc/live-build-hooks/90-raw-disk-image.binary#L133).

pzakha: Hmm, so I think the way it's done in the links you provided might cause some issues as we didn't specify `--make-rslave`, and so unmounting the child fs in the bind-mount would also unmount the child fs in the original mount. 

Here's an example:
```
	for path in /proc /sys /dev /run; do
		mount --rbind ${path} ${NROOT}${path}
		mount --make-rslave ${NROOT}${path}
	done
```
```
	for path in /proc /sys /dev /run; do
		umount -R ${NROOT}${path}
	done
```
Note that even if we use bind instead of rbind, we might still need to use `--make-slave`.

live-build/misc/upgrade-scripts/upgrade-container
pzakha: Alright, I think your explanation makes sense.
It would be helpful to add a comment in the code though.

prakashsurya: :+1: Sure, I can add more comments.

live-build/misc/upgrade-scripts/upgrade-container
prakashsurya: Interesting.. no need to do this complicated `mount | grep | tac | awk | xargs` nonsense? That seems simple enough.. I'll give it a try.

prakashsurya:
@dtzeng @pzakha When y'all have a chance, can you give this another look? The latest revision should address all prior comments, and have minimally tested both "in-place" and "not-in-place" upgrades.

I've added some minimal migration logic when performing a "not-in-place" upgrade, such that ssh at least works after a reboot (i.e. migrating the passwords, and minimal cloud-init configuration), but this will need to be revisited later.

grwilson:
We should think about the naming convention of the two types of upgrades. For example,

Express upgrade = in-place
Full upgrade = not-in-place

This is just a suggestion but we will eventually need to expose these to customers and define the differences.

live-build/misc/upgrade-scripts/upgrade-container
```diff
@@ -273,12 +273,149 @@ function run() {
...
+	#
+	while read -r file; do
+		migrate_file "$file"
+	done <<-EOF
+		/etc/cloud/cloud.cfg.d/99-delphix-development.cfg
+		/etc/cloud/cloud.cfg.d/99-delphix-internal.cfg
+		/etc/issue
+		/etc/nftables.conf
+		/etc/ssh/sshd_config
+		/etc/systemd/system/delphix-masking.service.d/override.conf
```
grwilson: How do we deal with services which are enabled by the stack and not enabled by default?

For example, snmpd and fluentd

prakashsurya: What do you mean, how is that handled on initial install? Presumably by the delphix-platform service? I’d have to look, but it should work the same as initial install.

grwilson: If we're doing a not-in-palce upgrade and fluentd was enabled by the stack, it looks like we would miss that configuration file on the newly upgraded dataset. Am I missing something?

prakashsurya:
I didn’t think we were planning to expose the “in-place” and “not-in-place” upgrades to customers. Did I misunderstand?

live-build/misc/upgrade-scripts/upgrade-container
grwilson: For example, I believe that fluentd creates an override file but I don't see it in the list

grwilson: Maybe the other way to look at it is, if delphix-platform handles this why do we need to copy over the override for these services explicitly here?

prakashsurya: Ah, yes. Anything done by the stack is going to be lost. That’ll need to be preserved by whatever mechanism we use for the “configuration migration” part. I’ve only added this migration as a short term solution, so the system will boot and be roughly configured the same as initial install.

grwilson:
This was just about considering naming conventions that will map to what we expose to customers. For example, if we not-in-place is always used for reboot upgrades then we could map whatever we call that type of upgrade to not-in-place.

live-build/misc/upgrade-scripts/upgrade-container
prakashsurya: Just to be clear, the migration done here is purposefully incomplete. I only tried to add migration for things appliance-build currently does for the initial build/install so the system can boot properly after the upgrade. We’ll need to circle back and implement a more complete solution later.

grwilson: That addresses my concern. I was also wondering if we should use a migration.d directory to allow scripts to be placed in there to handle the migration of individual components. I know we've used this convention before so thought it might be useful here. It might be overkill too.

prakashsurya: Yea, this is basically what I was thinking as well. I haven’t thought of anything better. One drawback being, we can easily forget to add a migration when new code is added.

I plan to circle back to this later, and will seek design ideas from folks.

prakashsurya:
I’m definitely open to changing the naming convention. I couldn’t come up with anything better, considering we’ve already been using the term “in-place upgrades” in various discussions. I considered “clean upgrade” instead of “not in-place”, but opted against it.

We can certainly change the terms later, but that’ll be a flag day for any consumers of it (e.g. upgrade logic in virtualization).

pzakha:
I think since we haven't really decided when we'll performing the in-place vs not-in-place, we should probably leave the names as is for now, especially since those won't be seen by the customer.

live-build/misc/upgrade-scripts/upgrade
```diff
@@ -0,0 +1,198 @@
...
+
+	"$IMAGE_PATH/upgrade-container" start "$CONTAINER" ||
+		die "failed to start '$CONTAINER'"
+
+	"$IMAGE_PATH/upgrade-container" run "$CONTAINER" \
+		"$IMAGE_PATH/execute" ||
+		die "'$IMAGE_PATH/execute' failed in '$CONTAINER'"
+
+	"$IMAGE_PATH/upgrade-container" run "$CONTAINER" \
+		/usr/bin/env systemctl start delphix-platform ||
```
pzakha: nit: why do you need the `/usr/bin/env` here?

live-build/misc/upgrade-scripts/upgrade-container
```diff
@@ -273,12 +273,149 @@ function run() {
...
+	# LX-72 Until we have a proper solution for migrating the
+	# configuration of a given appliance to the new root filesystem
+	# (i.e. any configuration that isn't supplied by a package), we
+	# resort to explicitly migrating some of this configuration
+	# here. The files listed here are dynamically generated and/or
+	# modified by appliance-build.
+	#
+	while read -r file; do
+		migrate_file "$file"
+	done <<-EOF
```
pzakha: Did you forget `/etc/netplan/50-cloud-init.yaml` by any chance? Without it networking will be unconfigured.

live-build/misc/upgrade-scripts/upgrade-container
```diff
@@ -273,12 +273,149 @@ function run() {
...
+	for dev in $(get_bootloader_devices); do
+		[[ -e "/dev/$dev" ]] ||
+			die "bootloader device '/dev/$dev' not found"
+
+		[[ -b "/dev/$dev" ]] ||
+			die "bootloader device '/dev/$dev' not block device"
+
+		chroot "/var/lib/machines/$CONTAINER" \
+			grub-install "/dev/$dev" ||
+			die "'grub-install /dev/$dev' failed in '$CONTAINER'"
```
pzakha: Did you figure out how does grub determine which ZFS dataset to get the `/boot/` directory from? Since it needs to mount the right dataset to get `/boot/grub.cfg`, it can't be any entry in `grub.cfg` that defines which dataset needs to be mounted at first.

live-build/misc/upgrade-scripts/verify
```diff
@@ -48,94 +41,46 @@ function cleanup() {
...
-if [[ -n "$DLPX_DEBUG" ]] && $DLPX_DEBUG; then
-	VERIFY_DEBUG_OPT="-Ddelphix.debug=true"
-fi
-
-if $opt_d; then
-	VERIFY_LIVE_MDS_OPT="-disableConsistentMdsZfsDataUtil"
-fi
+	die "'$IMAGE_PATH/execute' failed in '$CONTAINER'"
 
+report_progress 40 "Performing application upgrade verification"
```
pzakha: Should that be `report_progress 60`?

live-build/misc/upgrade-scripts/upgrade
prakashsurya: `systemd-run` needs a full path to execute.. alternatively, I could have provided the full path to `systemctl`.. I can do that if you think it'd be better and/or more clear.

pzakha: probably that'd be more clear, and the path isn't likely to change.

live-build/misc/upgrade-scripts/upgrade-container
prakashsurya: No. I'm relying on the same mechanism we use on first boot, which is our cloud-init configuration files that we migrate.

As we talked, it's probably only working because of a cloud-init bug, but I wanted to be consistent with how we configure networking on the initial boot.

Once we implement configuration migration "properly", we'll probably copy that file.. or w/e file the virtualization engine creates when configuring the network.

live-build/misc/upgrade-scripts/upgrade-container
prakashsurya: I'm not sure, but I didn't verify running `update-grub` and `grub-install` from the `chroot` works (i.e. the system boots to w/e dataset I used to `chroot`).

The contents of `/boot/grub/grub.cfg` in the `chroot` dataset will be "correct", while the running system's `/boot/grub/grub.cfg` will be "stale".

I don't really know the details about how grub will determine which `grub.cfg` file to use, but in my testing, the filesystem I expected to boot (i.e. the new one) was booted after running these commands.

live-build/misc/upgrade-scripts/verify
prakashsurya: Ah, yes, thanks for catching that.

PR #144, opened 2018-11-02
Configure delphix user's password during build

prakashsurya:
Soon the delphix-platform will remove it's current logic that handles
configuration of the delphix user's password. Thus, we (now) need to
handle configuration of the password in this repository, as part of the
initial appliance build.

Additionally, we're removing the ability to change the appliance's
username, since this flexibility is unnecessary, and only serves to make
things more difficult on upgrade and/or migration.

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.masking-development/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.masking-development/tasks/main.yml
@@ -15,10 +15,6 @@
 #
 
 ---
-- fail:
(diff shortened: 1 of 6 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/188/)

see also: https://github.com/delphix/delphix-platform/pull/14

prakashsurya:
cc: @pzakha @dtzeng @pandar00 @grwilson

PR #145, opened 2018-11-02
Always use "delphix.XXXXXXX" as the rootfs name

prakashsurya:
With this change, the root filesystem(s) of the appliance will take on a
different naming scheme. Currently, the name will be one of: "ubuntu",
"in-place.XXXXXXX", or "not-in-place.XXXXXXX", depending on how the
filesystem was generated (e.g. on initial build, or upgrade). With this
change, now the name will always be "delphix.XXXXXXX" regardless of if
the filesystem was generated on initial appliance build, or on upgrade.


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/190/)

prakashsurya:
cc: @dtzeng @pandar00 @pzakha @grwilson

prakashsurya:
```
$ mount | awk '$3 == "/" {print $0}'
rpool/ROOT/delphix.meciCFN on / type zfs (rw,relatime,xattr,noacl)
```
Cloned a VM from the image produced by `git-ab-pre-push`, verified the root filesystem's name.

PR #148, opened 2018-11-06
Use legacy pool version when creating rpool

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/196/)

ahrens:
Can you provide some context here?  Which pool is this, and why do we want/need it to be version=28?

prakashsurya:
IIRC, we do (or did?) similar when creating our initial root pool on illumos; as you know, this just makes the pool "more compatible" with different versions of ZFS.

In this case, we run the build using Ubuntu's ZFS version. Thus, to try and prevent any potential compatibility issues between the pool created here using Ubuntu's ZFS, and when we boot with our Delphix ZFS version, I'm opting to use a legacy version with all features disabled.

If we want to use a more recent version and/or pool feature, I think it'd be better to enable these via the delphix-platform or delphix-mgmt services.

prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/242/)

PR #153, opened 2018-11-13
Add python-pip for zfstest log processing

jwk404:
The log processing script requires the ordereddict and junit modules.
It would be preferable to install these modules via apt as packages,
but the packages in the apt repo do not work correctly, so we install
pip and use it to install the needed modules.

Testing:
I built the qa variant and verified `pip` was available. pre-push build is
running [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build/job/master/job/pre-push/170/console)

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.qa-internal/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.qa-internal/tasks/main.yml
@@ -53,6 +53,7 @@
     - libnss3-tools
     - libpam0g-dev
     - libssl-dev
+    - python-pip
   register: result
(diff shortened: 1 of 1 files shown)


jwk404:
Is there a way to see the bors build failure? Clicking on the "Details" link leads to a page that just says "Permission denied.

prakashsurya:
I can’t seem to figure out how to get at the test log for bors. I remember there was a way to do it, but can’t find it from my phone. I thought the comment it posts contained a link, but that’s not the case here. I’m going to retry and see what happens...

bors r+

PR #158, opened 2018-11-28
Lx 1418 Add mechanism to save container logs to a host

pandar00:
This is a reattempt of a discarded PR https://github.com/delphix/appliance-build/pull/132

Just to summarize, we discussed that adding additional binding directory and cherry-picking logs to copy was not a clean solution and should be leveraging `support_info` script if possible.

This PR uses `support_info` script to collect logs within the container and saves to `/var/dlpx-update/verify-logs` directory.

There is one TODO as noted below but just wanted to get early feedback how do you guys think about the approach.

TODO:
* I need to disable progress reporting from `support_info` script since app-gate doesn't know which ones are coming from `verify` script.

Testing
* Bundle collected at expected location
```Complete! The compressed log bundle is at /var/dlpx-update/verify-logs/dlpx-support-423FB6DC-D163-8354-29BA-516F13D1FCCE-20181128-01-44-47.tar.gz```

* Expected contents are there
```delphix@harry-dlpxtrunk:/var/dlpx-update/verify-logs/archive$ ls -la
total 23
drwxr-xr-x 5 root root  5 Nov 28 01:51 .
drwxr-xr-x 3 root root  4 Nov 28 01:51 ..
drwxr-xr-x 2 root root  3 Nov 28 01:44 dropbox
drwxr-xr-x 5 root root  5 Nov 28 01:44 log
drwxr-xr-x 2 root root 18 Nov 28 01:44 misc
delphix@harry-dlpxtrunk:/var/dlpx-update/verify-logs/archive$ cd dropbox/
delphix@harry-dlpxtrunk:/var/dlpx-update/verify-logs/archive/dropbox$ ls
upgrade_verify_report.json
delphix@harry-dlpxtrunk:/var/dlpx-update/verify-logs/archive/dropbox$ cd ../log/
delphix@harry-dlpxtrunk:/var/dlpx-update/verify-logs/archive/log$ ls -la
total 8
drwxr-xr-x 5 root root  5 Nov 28 01:44 .
drwxr-xr-x 5 root root  5 Nov 28 01:51 ..
drwxr-xr-x 4 root root 12 Nov 28 01:44 mgmt_log
drwxr-xr-x 2 root root  4 Nov 28 01:44 pg_log
drwxr-xr-x 9 root root 30 Nov 28 01:44 sys_log
```

`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/262/)

### DIFF
--- a/live-build/misc/upgrade-scripts/common.sh
+++ b/live-build/misc/upgrade-scripts/common.sh
@@ -15,7 +15,10 @@
 # limitations under the License.
 #
 
+# shellcheck disable=SC2034
 UPDATE_DIR=${UPDATE_DIR:-"/var/dlpx-update"}
+UPDATE_VERIFY_LOG_DIR=${UPDATE_DIR}/verify-logs
+SUPPORT_INFO_SCRIPT=${SUPPORT_INFO_SCRIPT:-"/opt/delphix/server/bin/support_info"}
 DROPBOX_DIR=${DROPBOX_DIR:-"/var/delphix/dropbox"}
 
 function die() {
--- a/live-build/misc/upgrade-scripts/verify-impl
+++ b/live-build/misc/upgrade-scripts/verify-impl
@@ -26,6 +26,24 @@ function usage() {
 	exit 2
 }
 
+function cleanup() {
+	local rc="$?"
+
+	#
+	# Persist container logs by saving to the container's host.
+	# Delete and create verify log destination directory to save only one copy of log to limit
+	# space usage.
+	#
+	rm -rf "${UPDATE_VERIFY_LOG_DIR:?}" ||
+		die "Failed to remove verify log directory ${UPDATE_VERIFY_LOG_DIR}."
+	mkdir -p "${UPDATE_VERIFY_LOG_DIR}/${CONTAINER}" ||
+		die "Failed to create verify log directory ${UPDATE_VERIFY_LOG_DIR}/${CONTAINER}."
+	${SUPPORT_INFO_SCRIPT} -o "${UPDATE_VERIFY_LOG_DIR}/${CONTAINER}" -t "log dropbox" ||
+		die "Failed to collect verify logs."
+
+	return "$rc"
+}
+
 opt_d=false
 while getopts ':df:l:o:' c; do
 	case "$c" in
@@ -44,6 +62,8 @@ fi
 if $opt_d; then
 	VERIFY_LIVE_MDS_OPT="-disableConsistentMdsZfsDataUtil"
 fi
+
+trap cleanup EXIT
 
 /usr/bin/java \
 	-Dlog.dir=/var/delphix/server/upgrade-verify \
--- a/live-build/misc/upgrade-scripts/upgrade-container
+++ b/live-build/misc/upgrade-scripts/upgrade-container
@@ -282,7 +282,7 @@ function destroy() {
 }
 
 function run() {
-	systemd-run --machine="$CONTAINER" --quiet --wait --pipe -- "$@"
+	systemd-run --machine="$CONTAINER" --setenv=CONTAINER="$CONTAINER" --quiet --wait --pipe -- "$@"
 }
 
 function get_bootloader_devices() {


pandar00:
cc: @dtzeng @prakashsurya

live-build/misc/upgrade-scripts/verify
```diff
@@ -41,6 +41,20 @@ function cleanup() {
...
+	# is deleted. We leverage support_info script to do the collection
+	# and only collect "log" and "dropbox" types. User who wish to add
+	# more collection types from support_info here should check if
+	# they work in the container.
+	# We collect only one copy to limit space usage.
+	#
+	report_progress 70 "Collecting verification logs"
+	rm -rf "${UPDATE_VERIFY_LOG_DIR:?}/*"
+	mkdir -p "${UPDATE_VERIFY_LOG_DIR}"
+	${SUPPORT_INFO_SCRIPT} -o "${UPDATE_VERIFY_LOG_DIR}" -t "log dropbox"
```
prakashsurya: This is supposed to run inside of the upgrade container, right? If so, I don't think this is correct.. right here, we'll be running on the host, not the container.

prakashsurya: I think it might be more appropriate to add this logic to the `verify-impl` script, which is intended to run inside the container.

Putting it in that script, will also make sure we run this log collection for "inplace" upgrades, as well as "not-inplace" upgrades. That script is also verify "app-gate" specific, so we wouldn't be leaking "app-gate" abstractions any more than we already do (e.g. leaking app-gate specific code into the upgrade scripts, which should be able to be run without the app-gate being installed).

What do you think?

live-build/misc/upgrade-scripts/verify
```diff
@@ -41,6 +41,20 @@ function cleanup() {
...
+	# We need to collect logs within the container and move to
+	# container's host otherwise those will delete when container
+	# is deleted. We leverage support_info script to do the collection
+	# and only collect "log" and "dropbox" types. User who wish to add
+	# more collection types from support_info here should check if
+	# they work in the container.
+	# We collect only one copy to limit space usage.
+	#
+	report_progress 70 "Collecting verification logs"
+	rm -rf "${UPDATE_VERIFY_LOG_DIR:?}/*"
```
prakashsurya: Why not remove the entire directory with:
```
rm -rf "${UPDATE_VERIFY_LOG_DIR:?}"
```
?

Does the `/*` part buy us anything? It means dot-files will not be removed, but idk if that's a bug, or a feature?

live-build/misc/upgrade-scripts/common.sh
```diff
@@ -15,7 +15,10 @@
 # limitations under the License.
 #
 
+# shellcheck disable=SC2034
 UPDATE_DIR=${UPDATE_DIR:-"/var/dlpx-update"}
+UPDATE_VERIFY_LOG_DIR=${UPDATE_DIR}/verify-logs
```
prakashsurya: I'm not necessarily against putting these logs in `/var/dlpx-update`, but we need to just be aware that this directory is backed by the `rpool/update` dataset, which has a quota on it. Thus, if we fill this directory with logs (I don't know what the disk usage is of these logs), it may mean we can't hold as many upgrade images as we previously could.

I don't know if the upgrade images + logs will cause ENOSPC issues, but I want to make sure we think that through. If we can hold both in this dataset, I'm OK with this approach.

live-build/misc/upgrade-scripts/verify
```diff
@@ -41,6 +41,20 @@ function cleanup() {
...
+	mkdir -p "${UPDATE_VERIFY_LOG_DIR}"
```
prakashsurya: Instead of using a single "global" directory to store logs, how about we use a per-container directory? This way, we could potentially have multiple processes running upgrade verification at the same time, and they won't clobber each other when writing their logs. e.g. use a directory like this:
```
mkdir -p "${UPDATE_VERIFY_LOG_DIR}/${CONTAINER}"
${SUPPORT_INFO_SCRIPT} -o "${UPDATE_VERIFY_LOG_DIR}/${CONTAINER}" ...
```

While we might not run multiple upgrade verify's concurrently right now, since all the other code would allow it, it'd be a shame to break that just because of this log collection.

Also, if we do make the log collection use container specific directories, it means (theoretically) we could have more logs collected at any given time, which means that we might need more disk space to store them. Further, it means that we'd have to clean up the per-container directories at some point (e.g. the old logs that aren't needed anymore).

If this per-container log directory makes things more complicated for little benefit, then at least we could do something like this:
```
mkdir -p "${UPDATE_VERIFY_LOG_DIR}/${CONTAINER}"
${SUPPORT_INFO_SCRIPT} -o "${UPDATE_VERIFY_LOG_DIR}/${CONTAINER}" -t "log dropbox"
mv "${UPDATE_VERIFY_LOG_DIR}/${CONTAINER}" "${UPDATE_VERIFY_LOG_DIR}/latest"
```
This way, if multiple verify's were running concurrently, we'd at least atomically get the logs for one of the runs or the other.. rather than now, where we could get some logs from one run, and some logs from the other run.

What do you think?

live-build/misc/upgrade-scripts/common.sh
pandar00: Good point. I collected support bundle (just the `log` and `dropbox` type) from longevity setup and the compressed size is 508M. The logs are currently rotating so I presume it's near the upper bound of a bundle size for those two types. I think saving one is OK, although we have to rethink different place to store this if we decide to store more than one.

Do you have any good location in mind where to save these logs? How about `/var/delphix/server/log`?

live-build/misc/upgrade-scripts/verify-impl
```diff
@@ -26,6 +26,25 @@ function usage() {
 	exit 2
 }
 
+function get_container_name() {
```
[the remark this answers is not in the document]
prakashsurya: I'd be OK if it didn't work with `machinectl shell` since we can use `upgrade-container run` to execute commands in the container. If we wanted to, we could add a `upgrade-container run-interactive` (maybe with a better name) to get an interactive shell in the container, since `run` is meant to be run non-interactively.

This way, we could update the `upgrade-container` script to expose the environment variable, and we'd "hide" the technology (`systemd-nspawn` and/or `machinectl`) that's used to create the container from users (both from actual people, but also from scripts and other software using the container).

With that said, if we want to expose the fact that we're using `systemd-nspawn` for the container, and encourage folks to use things like `machinectl` and `systemd-run` directly, I'm OK with that too. It just means that we're more closely tied to `systemd-nspawn` and the tools it provides, and changing that would be much more difficult; forcing folks to use `upgrade-container run` gives us a layer of abstraction, so we can change the underlying tools without changing the users of the script.

So, I just wanted to bring up an alternative solution, in case it wasn't considered; ultimately, I'm OK going this route if we deem this best (i.e. encouraging the use of tools like `machinectl`, which implies we're exposing the fact that we're using a `systemd-nspawn` container to users).

live-build/misc/upgrade-scripts/verify-impl
```diff
@@ -26,6 +26,25 @@ function usage() {
...
+	return basename $(df -k / | tail -1 | awk '{print $1}')
```
prakashsurya: how about changing this to:
```
basename $(zfs list -Ho name /)
```
?

I've also never seen `return` used to return a string, but I'm not a bash expert. Normally I'd do something like:
```
basename $(zfs list -Ho name /)
```
(i.e. no `return`).. and then it'd be used like this:
```
mkdir -p "${UPDATE_VERIFY_LOG_DIR}/$(get_container_name)"
```
i.e. using `$(...)` instead of `${...}`.

I'm fine if it works your way; I'll have to read up on the docs since I might have just learned something new :)

live-build/misc/upgrade-scripts/verify-impl
```diff
@@ -26,6 +26,25 @@ function usage() {
...
+function cleanup() {
+	local rc="$?"
+
+	#
+	# Persist container logs by saving to the container's host.
+	# Delete and create verify log destination directory to save only one copy of log to limit
+	# space usage.
+	#
+	rm -rf "${UPDATE_VERIFY_LOG_DIR:?}"
+	mkdir -p "${UPDATE_VERIFY_LOG_DIR}/${get_container_name}"
```
prakashsurya: Now that we're storing the logs in a per-container directory, how will these directories get cleaned up? Will there be logic in the app-gate that does this? or should we do this in the upgrade-scripts (e.g. perhaps when the container is destroyed)?

prakashsurya: I think I found the answer to my question in the comment above. We "cleanup" the logs for old runs, by removing the entire directory (including any old logs) each time we run verification. That seems reasonable to me.

live-build/misc/upgrade-scripts/verify-impl
```diff
@@ -26,6 +26,25 @@ function usage() {
...
+	${SUPPORT_INFO_SCRIPT} -o "${UPDATE_VERIFY_LOG_DIR}" -t "log dropbox"
```
prakashsurya: I think we might want to use `|| die "failure message"` for each of these commands.. if any of them fails, we want the script to return an error, right? as it is, I think the script could return success if any of these failed.

live-build/misc/upgrade-scripts/verify-impl
pandar00: Sorry about that. Review didn't get update with latest change. `return` doesn't work with string like you said. Should've been `echo`. However, I'm going with your suggestion of exposing environment variable from `upgrade-container` level, so this function will get removed on the next revision.

pandar00:
@prakashsurya, @dtzeng 
Please take a look when you guys have a chance. I've updated the review and repeated the same testing.

Just a side note. There is a minor app-date change that needs to to be pushed before this since app-gate logic scans `/var/dlpx-update` and adds *all* directories. My change adds only the directories that matches version name pattern (http://reviews.delphix.com/r/45379/).

PR #160, opened 2018-11-28
Remove "ansible-done" file during upgrade

prakashsurya:
This change updates the upgrade logic such that we properly remove the
"delphix-platform" service's "ansible-done" file prior to restarting the
service. Otherwise, restarting the service may not do anything useful,
since the service will only run the new ansible logic if this file
doesn't exist; i.e. if the file exists, the service is basically a noop.


prakashsurya:
`git-ab-pre-push -v internal-dev` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/265/)

PR #161, opened 2018-11-29
Enable disk.EnableUUID in the OVA Template

don-brady:
This fixes the device ids and serial numbers for Linux based VMs on ESXi by providing the `disk.EnableUUID` key.

Confirmed by building a complete OVA archive using pre-push (build #266) and using that OVA to create a new VM.  The VM boots and has properly configured device UUIDs (as seen by `udevadm info`).  Also confirmed that the 'disk.EnableUUID' key was present in the VM's vmx file.


shartse:
This is great! Can you run some tests (maybe blackbox precheckin) on your VM to sanity check that the device ids are handled properly before integrating this?

prakashsurya:
`git-ab-pre-push` should work, right? it'll consume the new VMX file in this change.

don-brady:
This is now blocked behind JIRA TOOL-6885 which allows the parsing of this new arg when using the `ovftool`. Once that lands this can proceed.

don-brady:
The ovatool changes have landed so this can now proceed.

PR #166, opened 2018-12-06
Customize builds for different platforms

jgallag88:
This change provides the ability to build versions of the appliance
customized for different platforms (hypervisors and cloud providers).
This is done by installing a different versions of the delphix-platform
and delphix-kernel packages depending on which platform we are
building for. Since we only want to have a single upgrade image per
variant, this change also adds a second stage to the build which
combines the live-build output for multiple platform versions of the
same variant into a single upgrade tarball.

The live-build stage of the build is now run by invoking 'gradle' with a
a target which is a combination of variant and platform, e.g.
`gradle buildInternalDevEsx`.

The second stage of the build is run by invoking 'gradle' with a variant
as a target, e.g. `gradle buildUpgradeImageInternalDev`. When the second
stage is run, an environment variable 'AWS_S3_URI_LIVEBUILD_ARTIFACTS'
can be passed. If it is used, previously built live-build artifacts will
be downloaded from the provided S3 URIs, and placed in
`live-build/build/artifacts` as if they had been built locally. If it is
not used, live-build will be invoked for each of the hypervisors
specified in the 'DELPHIX_PLATFORMS' environment variable.

A couple notes about the implementation:

1. This change replaces the Make build with a Gradle one. The build logic
    needed for this change was difficult to express using Make and
    resulted in a Makefile which was very difficult to understand. The
    use of Gradle make handling this build logic more straightforward
    and also made it possible to add better support for incremental
    builds.
2. This change removes the idea of the 'base' live-build variant. The
    base variant contains the kernel, and because the kernel differs
    between hypervisors, it cannot be shared between different hypervisor
    builds. It would be possible to have a different version of the base variant
    per hypervisor, and share that between different variants built for
    the same hypervisor. However, this likely isn't worth the effort
    because it doesn't help in either of the two most common use cases:
      - Building via a jenkins job: when building via Jenkins, each
         variant will now be built via a sub-job running on its own build
         VM, so the base would be rebuilt for each sub-job anyway.
      - Developers iterating on changes on personal build VMs: in this
         case developers are most likely to be building a single variant,
         in which case the 'base' variant would be less likely to be
         re-used.
3. We no longer do the live-build in place (that is, directly in
    `live-build/variant/<variant>/`). Now that we have multiple builds per
    variant, we need to make sure that intermediate live-build
    artifacts from one build are not incorrectly re-used in the next
    build of the same variant, which might be for a different
    hypervisor. The simplest way to accomplish this is just to do the
    live-build in a throw-away directory.

In light of these two changes, some of current layout of the
repository no longer makes sense, so this change re-arranges a number
of files in the repo, particularly in the `live-build/` directory.


README.md
```diff
@@ -20,9 +20,9 @@ Log into that VM using the "ubuntu" user, and run these commands:
     $ git clone https://github.com/delphix/appliance-build.git
     $ cd appliance-build
     $ ansible-playbook bootstrap/playbook.yml
-    $ ./scripts/docker-run.sh make internal-minimal
+    $ HYPERVISORS=generic ./scripts/docker-run.sh make internal-minimal
```
prakashsurya: How much thought have we given to the term of "hypervisors" here? Personally, it seems a little odd since we're not really listing hypervisors; e.g. "azure" isn't really the hypervisor name vs. "hyperv".

We also have the same concept in the linux-pkg repository, and I think the term "platform" is used there.

It seems like ubuntu uses the term "kernel flavour" for this; see [here](https://wiki.ubuntu.com/Kernel/Dev/Flavours).. which is a little confusing, since it also has the term "ubuntu flavour" which is different; see [here](https://www.ubuntu.com/download/flavours)

I think it'd be good to collectively decide on a good term to use here, and be consistent across our various projects/repositories. While it might not seem important, correct/consistent terminology can make future discussion more clear, since the term will be well defined across teams/projects/repos.

Personally, I lean towards something like "flavour" since it's already used by ubuntu (although we'd be overloading it even more), and it doesn't come with any baggage inside delphix (nothing inside delphix is coined "flavour" AFAIK, vs. hypervisor, where folks might have a preconceived notion of what that means); "platform" might also be good, but that's already used in the "delphix dynamic data platform".

Using something more ambiguous like "flavour" vs "hypervisor" seems more appropriate when we'll likely create different images for things that'll use the same underlying hypervisor; e.g. google cloud which is kvm based, vs. other kvm based cloud providers (openstack?).

Thoughts?

prakashsurya: It's probably also worth noting that the app-gate's `server.conf` has a `platform` field, which I think is used to denote this type of thing; e.g. see [here](https://gitlab.delphix.com/app/dlpx-app-gate/blob/master/appliance/server/sh/boot/dxinit.sh#L334). So, perhaps "platform" would actually be a good choice of terminology since it's already used.

pzakha: I agree that we should try to come up with a common terminology across all the projects. In `linux-pkg` we are using `kernel` for a specific kernel version and `platform` for a kernel "flavour". I think there's a distinction between the "kernel flavour" and what we refer to as the "hypervisor" here, so we should probably not treat them as the same. We could have 2 different "hypervisors" use the same "kernel flavour", e.g. let's say we could have 2 different builds for `OpenStack` and `CloudStack`, which both use the `kvm` kernel.

As for the naming, I think we'd need to discuss further.

jgallag88: I thought that I replied to this before, but I must have never published my comment.

I'm definitely open to using another term. My feeling is that it is actually better to stay away from something vague like 'flavour', since that basically means the same thing as 'variant' so it isn't obvious how the two relate to each other. We could use something like different types of 'variant', e.g. 'product variant' and 'platform variant'. I'm not sure if that would do a better job at conveying what these two concepts are and that they orthogonal to each other.

I also like 'platform', since I think that does a good job of capturing the concept introduced here. I don't worry too much about confusion with 'delphix dynamic data platform', since I don't think that term is likely to crop up too often in this context. @pzakha, I think both `linux-pkg` and `appliance-build` face the issue you point out. If we ever do have different builds which use the same 'kernel flavor', we could need to make that distinction in `delphix-platform` if the two different builds include different packages. If we do want to use 'platform', we could just punt on that issue for now in both repositories.

Makefile
```diff
@@ -14,46 +14,135 @@
 # limitations under the License.
 #
 
+SHELL = /bin/bash
+
 ALL_VARIANTS = $(shell find live-build/variants -maxdepth 1 -mindepth 1 -exec basename {} \;)
 ALL_INTERNAL = $(shell find live-build/variants -maxdepth 1 -mindepth 1 -name 'internal-*' -exec basename {} \;)
 ALL_EXTERNAL = $(shell find live-build/variants -maxdepth 1 -mindepth 1 -name 'external-*' -exec basename {} \;)
 
+ALL_HYPERVISORS := generic aws azure gcp kvm
```
sebroy: I was confused by the term "generic" in the linux-pkg repo, and I think I'm still confused. What is a generic hypervisor? Where does VMware ESX fit in? We only need the VMware open-vm-tools package installed on that one, but we currently install it on all platforms as part of a `delphix-platform` dependency. That probably needs to be moved here as part of a VMware-specific build (?)...

sebroy: The `walinuxagent` package is similarly installed everywhere using a general dependency from `delphix-platform`, and it should only be installed on Azure.

README.md
pzakha: Okay, so let's agree on the terms we want to use for different things. I propose:
Product variant: `variant`, e.g. internal-dev, internal-qa
Cloud variant: `platform`, e.g. ESX, AWS, CloudStack
Kernel flavour: `kernel type`, e.g. generic, aws, kvm
Specific kernel version: `kernel version`, e.g. 4.15.0-18-generic

So we'd keep using `variant` for the product variant, we'd change `platform` to `kernel type` in `linux-pkg`, and we'd change `kernel` to `kernel version` in linux-pkg. Finally, we'd use `platform` for the target cloud platform (or hypervisor in the case of ESX). delphix-platform will be customizable for each `platform`.

We currently have `KVERS` and `PLATFORM` placeholders in a few packages. We can keep `KVERS` and change `PLATFORM` to `KTYPE`.

For the `kernel type`, we could also use `kernel flavour` or simply `kernel`.

prakashsurya: > Product variant: variant, e.g. internal-dev, internal-qa

This is awfully similar to "product types", and likely easy to mix up; e.g. standard, aws marketplace, etc.

pzakha: @prakashsurya So are you proposing we change the current term `variant` to refer to things like internal-dev?
Just in case it wasn't clear, the terms I'm proposing to use are:
`variant`, `platform`, `kernel type`, `kernel version`

prakashsurya: I'm commenting on "product variant", and how that sounds similar to "product type". I wouldn't be surprised if folks confuse the two due to their similarities (both in their name, but also in how the two are implemented in the build), that's all.

It'd be good to loop in the notion of our product type in this list, but that's kind of encoded in variant (at least right now, this may not be the case moving forward).

> So are you proposing we change the current term variant to refer to things like internal-dev?

I don't understand; isn't this how you defined it above? "product variant".. "e.g. internal-dev internal-qa".. I coined the term variant in the appliance-build repository to mean these different build variations, so I'm not implying we change that. Just that this term is distinct from our product types, and they warrant terms distinct enough to avoid accidental confusion.

We could have a "dev" variant of our "standard" product type (what internal-dev is), but we could also have a "dev" variant of our "aws marketplace" product type (which we don't currently build). We could also have an "external" variant of our "standard" product type (what external-standard is). Basically, as it stands now, we could potentially want to produce the cross product of build variants (which is "external", "dev", and "qa") and product types (which is "standard", "aws marketplace", etc).

My suggestion would be to use the terms "build variant" and "product type".

pzakha: Okay, so just to clarify, as it looks like I've caused some confusion, here are the terms I'm proposing:
`variant`: e.g. internal-dev, internal-qa
`platform`: e.g. ESX, AWS, CloudStack
`kernel type`: e.g. generic, aws, kvm
`kernel version`: e.g. 4.15.0-18-generic

If we want to subdivide `variant` into `build variant` and `product type`, I'm fine with that. I've mostly been focusing on the other 3 terms.

Makefile
sebroy: Ok, I guess I'm confused with the use of the term "generic" to describe a hypervisor (the variable is called `ALL_HYPERVISORS`) if there is no generic hypervisor. Anyway, my overall point/questio is: how do we build an ESX-specific variant (or whatever we decide to call that) with `open-vm-tools` installed?

pzakha: @sebroy Using the terminology I've presented in the post above, we'd build for the ESX `platform` by customizing the `delphix-platform` package. The idea is that there would be a different `delphix-platform` package for each `platform`. When appliance-build builds for a particular platform, it will select the appropriate `delphix-platform` package.

So in other words, most of that work would go into the `delphix-platform` project.

README.md
jgallag88: This terminology sounds good to me.

So in appliance-build, we would expose `platform` as a parameter that can be passed in, and then `kernel type` would be chosen automatically based on the `platform`, and `kernel version` would be chosen implicitly by specifying the version of `linux-pkg` to use.

How do we see this working in `linux-pkg` though? It seems like we would need to make `platform` a parameter. Would `kernel version` need to be configurable as well?

pzakha: The way I see it, `appliance-build` doesn't need to care at all about the kernel, only about the platform. `delphix-platform` is the one that will install the proper kernel via dependencies.

For `linux-pkg`, it doesn't need to know anything about platforms. It only cares about `kernel version`s and `kernel type`s. When linux-pkg builds delphix-platform, delphix-platform will be the one that is aware of all the platforms and will produce one deb for each.

jgallag88: `appliance-build` does need to care about to the kernel, because you have to specify the `kernel-flavour` when running live-build, otherwise you end up with the generic kernel in addition to the one specified by `delphix-platform`. That's pretty easy to handle though. We can either hardcode the mapping of `platform` to `kernel type`, or just get the `kernel type` out of the `delphix-platform` package's metadata.

I'm still not sure what `linux-pkg`'s behavior will be though. If someone runs `linux-pkg` with just `gcp` specified for `kernel type`, will `delphix-platform` decide it should only produce a `delphix-platform-google` package?

pzakha: > appliance-build does need to care about to the kernel, because you have to specify the kernel-flavour when running live-build, otherwise you end up with the generic kernel in addition to the one specified by delphix-platform
I'm not sure to understand, does appliance-build try to install a kernel manually somewhere? The idea is that appliance-build doesn't install the kernel itseld but rather relies on `delphix-platform` to pull it in.

> If someone runs linux-pkg with just gcp specified for kernel type, will delphix-platform decide it should only produce a delphix-platform-google package?
The `kernel type` field is mostly there so that we can override the default kernel versions. For normal runs of `linux-pkg` (i.e. those that will be consumed by appliance-build) it *has* to build packages for all supported kernel types.

jgallag88: > I'm not sure to understand, does appliance-build try to install a kernel manually somewhere?

Yeah, exactly: https://github.com/delphix/appliance-build/pull/166/files#diff-c6618c419db1b442c4c804d5cf7b36abR104. I don't think this is an issue though, `delphix-platform` can still be the source of truth for the `kernel type`.

> The kernel type field is mostly there so that we can override the default kernel versions.

So `linux-pkg` will need to build for every kernel type on every build, and when run manually, some of the kernel types may have their versions overridden? That seems reasonable to me.

jgallag88: Nice find! That doesn't seem to be documented in the man page. I'll try it out.

pzakha: Yeah, I'd recommend getting the source package for live-build on Ubuntu and then extracting it on your laptop. It helped me a lot when looking for some customizations of the build since the man pages are far from being comprehensive.

scripts/fetch-livebuild-artifacts.sh
```diff
@@ -0,0 +1,61 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#
+# This script script downloads previously built vm artifacts from S3
+# and places them in live-build/artifacts/ as if they had been built by
+# running by running 'make internal-dev-aws', etc. This allows us to
```
pzakha: typo: by running x2

scripts/fetch-livebuild-artifacts.sh
```diff
@@ -0,0 +1,61 @@
...
+	rm SHA256SUMS
+
+	mv ./* "$TOP/live-build/artifacts/"
+
+	popd &>/dev/null
+	rm -rf "$tmp_dir"
+}
+
+if [[ -z "$TOP" ]]; then
+	echo "Must be run inside the git repsitory." 2>&1
```
pzakha: typo: repsitory

upgrade/upgrade-scripts/execute
```diff
@@ -68,7 +68,7 @@ EOF
 
 apt_get update || die "failed to update apt sources"
 
-INSTALLED_VERSION=$(dpkg-query -Wf '${Version}' delphix-entire)
+INSTALLED_VERSION=$(dpkg-query -Wf '${Version}' delphix-entire-generic)
```
pzakha: I think this code is outdated. There is no `delphix-entire-generic` package anymore.
We should also check for currently installed flavor of `delphix-entire-PLATFORM` and then install the same flavor.
We can use: 
```
dpkg-query -Wf '${Package}' 'delphix-entire-*'
dpkg-query -Wf '${Version}' 'delphix-entire-*'
```

Makefile
```diff
@@ -14,46 +14,135 @@
 # limitations under the License.
 #
 
+SHELL = /bin/bash
```
pzakha: I feel like this Makefile is getting overly complex.
An idea would be to split it out into 2 scripts:
One `configure` script, that takes the proper parameters and then feeds them into a `Makefile.in` file to create a `Makefile`.
the `configure` script could take "variant", "platform", etc. as parameters, do the proper argument checking, and then run a simple `make all`.
We could look at other ways to simplify it, but this would be one way to keep using a Makefile.

As of right now, this is quite hard to read & understand.
I don't know how much work that would take to refactor it though.

live-build/config/hooks/82-upgrade-repository.binary
```diff
@@ -118,6 +117,7 @@ fi
 sed -i "s/@@DEPENDS@@/$DEPENDS/" delphix-entire.ctl
 sed -i "s/@@VARIANT@@/$APPLIANCE_VARIANT/" delphix-entire.ctl
 sed -i "s/@@VERSION@@/$VERSION/" delphix-entire.ctl
```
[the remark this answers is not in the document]
prakashsurya: I haven't really looked at this change yet, but.. for customer images (with the virtualization package installed), the version should be the same (since they'll all still use the virtualization package's version, and that package will be the same in all the artifacts), right?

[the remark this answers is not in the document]
pzakha: Yeah you are right. I didn't recall we added that logic there, although I remember we discussed about it. You can drop that comment then.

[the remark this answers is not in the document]
jgallag88: Yeah, the version will be the same for all variants that have virtualization installed (not just the customer images). I figured it probably doesn't matter too much for the other variants.

Makefile
jgallag88: I agree that this Makefile is pushing the boundaries of what seems reasonable. The idea of having a `configure` script is interesting. I'd have to play with it a bit to see how it looks.

I was also thinking about replacing `make` with `Gradle`, which I think is vastly better for writing build logic that falls outside the normal C/C++ model. I was a little reluctant to add a JRE as a build dependency, though. Also, while we use Gradle pretty widely at Delphix, it would be new in the Linux appliance build area.

pzakha: Yeah I'm not sure what are the requirements to run gradle. Maybe having a JRE dependency wouldn't be that bad if that's all it requires? As long as it's not picky on the JRE version, that we do not need to build the build scripts and we can easily add the JRE dependency to the Dockerfile, I'd not see much issues with that.

upgrade/upgrade-scripts/common.sh
```diff
@@ -38,8 +39,13 @@ function get_image_version() {
 	basename "$(get_image_path)"
 }
 
+function get_platform() {
+	dpkg-query -Wf '${Package}' 'delphix-entire-*' | \
```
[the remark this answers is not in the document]
jgallag88: That's a great point. Maybe we can look at `delphix-platform-*` instead of `delphix-entire-*`. Would there ever be a case when that isn't installed? I can't think of any off the top of my head.

upgrade/upgrade-scripts/common.sh
```diff
@@ -1,4 +1,3 @@
-#!/bin/bash
```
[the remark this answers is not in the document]
jgallag88: It looks like the file is already non-executable, which is good. My tendency, and what seems to be the tendency in other Delphix code, is to leave shebangs off of sourced scripts, since they get ignored by the interpreter. If it helps with some editors though, I don't have any objection to leaving it in.

live-build/config/hooks/90-linux-migration-artifact.binary
```diff
@@ -52,15 +52,19 @@ rm -rf "$MIGRATION_ARTIFACT" "$DEPOT_DIRECTORY"
...
 #
 # Not all variants include the virtualization application (e.g.
-# internal-minimal) so don't bother generating a migration image
-# for those. /opt/delphix is always created as part of the
-# appliance. Thus, checking if it exists is a quick way for us to
-# distinguish between variants that have the appliance from
-# the ones that don't.
+# internal-minimal). Since a migration image wouldn't be useful for such
+# variants, we skip creation of it to save time. Instead, we create an
+# empty, dummy tarball that will satify Make when doing incremental
```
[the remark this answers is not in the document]
jgallag88: Yeah, we shouldn't need this hack any more. I'll update this.

build.gradle
```diff
@@ -0,0 +1,86 @@
...
+
+for (variant in allVariants) {
+    def taskName = "buildUpgradeImage${toCamelCase(variant).capitalize()}"
+    tasks.create(taskName, Exec) { task ->
+        dependsOn mkArtifactsDir
+
+        if (System.getenv("AWS_S3_URI_LIVEBUILD_ARTIFACTS") != null) {
+            dependsOn ":live-build:fetchLiveBuildArtifacts"
+        } else {
+            for (platform in System.getenv("PLATFORMS")?.trim()?.split()) {
```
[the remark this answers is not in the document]
jgallag88: That seems reasonable to me. What should the default be? Maybe 'kvm', since that's what people would test in qemu?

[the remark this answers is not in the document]
pzakha: Yeah kvm is probably fine.

upgrade/upgrade-scripts/common.sh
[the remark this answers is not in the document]
pzakha: That would probably not be a big deal, since we still have the .sh denomination, however I think this would be preferable over specifying the interpreter manually for shellcheck. We also have https://github.com/delphix/appliance-build/issues/186 that may be leveraging this information although I'm not sure.

upgrade/upgrade-scripts/common.sh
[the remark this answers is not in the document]
pzakha: We could do that, however I'm no sure we should support upgrade at all when `delphix-entire` is missing. We may end up adding some logic to do additional checks, so I think for now just failing when delphix-entire is missing should be fine.

prakashsurya: Please verify this works for not-in-place upgrades; e.g.
```
$ sudo /var/dlpx-update/latest/upgrade -v not-in-place
```
when running `execute` for that upgrade type, `delphix-entire` isn't installed yet, so I don't think doing this `dpkg-query` will return anything useful.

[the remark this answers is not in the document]
jgallag88: I think I'd prefer to leave that policy to the higher level logic, and just have this little helper function be reliable.

upgrade/upgrade-scripts/common.sh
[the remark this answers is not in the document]
jgallag88: I don't feel very strongly about this, so I've reverted this back to the way it was.

upgrade/upgrade-scripts/common.sh
jgallag88: I've modified the `execute` script to have the platform passed in. That seems a lot easier than trying to determine the platform from withing the container. Let me know what you think.

prakashsurya: Yea, that seems reasonable to me. Thank you!

prakashsurya: Thinking about this a little more, do we need to do anything special when we call `debootstrap` w.r.t. the platform? e.g. I'm curious which kernel package it'll install by default, and/or if it'll install a kernel package at all...

jgallag88: I think `debootstrap` doesn't install any kernel packages (at least the way we are using it).

prakashsurya: Sweet. If it just works, that's great! Thanks.

live-build/build.gradle
```diff
@@ -0,0 +1,112 @@
...
+ */
+
+apply plugin: "base"
+
+apply from: "${rootProject.projectDir}/gradle-lib/util.gradle"
+
+task ancillaryRepository(type: Exec) {
+    inputs.file "${rootProject.projectDir}/scripts/build-ancillary-repository.sh"
+
+    for (envVar in ["AWS_S3_URI_VIRTUALIZATION",
```
prakashsurya: does this mean, we now have to updated two places when exposing environment variables to our scripts? i.e. the `docker-run.sh` file, and also the gradle file(s)?

upgrade/upgrade-scripts/execute
```diff
@@ -63,8 +68,23 @@ EOF
...
+# before installing delphix-entire, because delphix-entire depends on
+# delphix-virtualization which in turn depends on the virtual package
+# delphix-platform. If some version of delphix-platform is already
+# installed, that dependency is already satisfied. If it is not, apt
+# will choose one of the packages that satisfies the dependency, and it
+# might choose the wrong one. When this happens apt will end up trying
+# to install two packages which provide delphix-platform, which will
+# fail because the two packages conflict with each other.
+#
+apt_get install -y "delphix-platform-$platform" ||
```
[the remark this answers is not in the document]
pzakha: @jgallag88 I know we've talked about this and I agreed on the approach, I'd still really like to understand more why is apt getting confused by this. 
Since `delphix-entire-$platform` should depend on `delphix-platform-$platform` and `delphix-platform-$platform` already provides `delphix-platform`, I'm really surprised that apt doesn't take it into consideration when looking for the dependencies of `delphix-virtualization`.
Let me know if you have a system in that state, I'd really love to take a look at it and understand better what's going on.

scripts/build-upgrade-image.sh
```diff
@@ -0,0 +1,76 @@
...
+# here we just take the artifacts from the live-build stage for whatever
+# appliance versions were built (making sure that we built for at least
+# one platform), and build an upgrade image from that.
+#
+LIVE_BUILD_OUTPUT_DIR="$TOP/live-build/build/artifacts"
+if ! compgen -G "$LIVE_BUILD_OUTPUT_DIR/$APPLIANCE_VARIANT*.debs.tar.gz"; then
+	echo "No live-build artifacts found for this variant" >&2
+	exit 1
+fi
+for deb_tarball in "$LIVE_BUILD_OUTPUT_DIR/$APPLIANCE_VARIANT"*.debs.tar.gz; do
```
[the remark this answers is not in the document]
pzakha: So from my understanding of how apt works, if we have both `foo=1.0` and `foo=1.1`, then the `delphix-entire` that has a dependency on `foo=1.0` will fail. apt will see `foo` as a dependency and will install the latest available foo according to apt preferences (in this case, since both are in the same repository it will take the highest version). Then, after the dependency has been installed `delphix-entire` installation will fail since it will not find `foo` with version `1.0` installed.

scripts/build-upgrade-image.sh
```diff
@@ -0,0 +1,76 @@
...
+	tar xf "$deb_tarball" -C debs
+done
+
+# Generate an Aptly/APT repository
+aptly repo create -distribution=bionic -component=delphix upgrade-repository
+aptly repo add upgrade-repository debs
+aptly publish repo -skip-signing upgrade-repository
+
+# Include version information about this image.
+VERSION=$(dpkg -f "$(find debs/ -name 'delphix-entire-*' | head -n 1)" version)
```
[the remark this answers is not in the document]
jgallag88: The version of delphix-entire is typically taken from delphix-virtualization, and the version of delphix-virtualization will be the same in each case. If the delphix-entire packages are all built on the same machine, then they all use the same ancillary repository. If they were built via Jenkins in parallel, the Jenkins job ensures that the same version of delphix-virtualization is passed to each sub job.

If delphix-virtualization isn't installed, then the versions could potentially be different, but that only applies to the internal-minimal variant, which didn't seem like too much of an issue to me, since we aren't going to do any real upgrade testing with that variant.

I can add a comment to that effect if that all seems reasonable.

upgrade/upgrade-scripts/common.sh
```diff
@@ -36,8 +36,13 @@ function get_image_version() {
 	basename "$(get_image_path)"
 }
 
+function get_platform() {
+	dpkg-query -Wf '${Package}' 'delphix-platform-*' |
```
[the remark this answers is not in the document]
jgallag88: I had it that way originally, but Pavel pointed out that delphix-entire won't be installed on developer machines that have had packages updated manually, which could happen in normal development workflows. I'm not sure how upgrade should behave in that situation, but I figured it was better to have this helper function be reliable in that scenario, and let higher level upgrade logic decide how to handle the case where delphix-entire isn't installed.

upgrade/upgrade-scripts/execute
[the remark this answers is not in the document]
jgallag88: > If we leave this out, why doesn't installing delphix-entire-$platform install delphix-platform-$platform? doesn't delphix-entire-$platform depend on delphix-platform-$platform?

So it does install delphix-platform-$platform, but it installs delphix-platform-$someOtherPlatform too. Say we are upgrading a vm on AWS. I suspect what happens is that apt first sees the `delphix-virtualization` dependency on `delphix-platform`. It also sees that there are mulitple versions of `delphix-platform` available, none of which are yet installed. In this scenario, apt picks an arbitrary package that provides the virtual package to install, say `delphix-platform-kvm`. It later sees that it needs to install the concrete package `delphix-platform-aws` because that is a dependency of `delphix-entire-aws`. It isn't smart enough to go back and realize that it no longer needs `delphix-platform-kvm`, so it tries to install both, and they conflict with each other.

It's possible that `apt_get install -y "delphix-platform-$platform delphix-entire-$platform` would work, but I don't know how we could be sure whether it's guaranteed to work or is just working accidentally unless we understood all of the details about how apt resolves transitive dependencies where virtual packages are involved (for which I haven't found any documentation).

Another way to resolve this would be to remove the explicit dependency on `delphix-platform` from `delphix-virtualization`. What I like about the current strategy though is that it makes the order in which the packages are installed more similar to the order used in live-build in appliance-build. I think this is good from a consistency perspective, since in general installing the same packages in different orders can result in different end-states. However, there could be other ways to make appliance build more similar to a not-in-place upgrade, maybe by getting rid of live-build, as you mentioned elsewhere.

@pzakha I can get a VM set up shortly.

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
```diff
@@ -37,10 +37,12 @@
...
-- docker_image:
-    path: "{{ toplevel.stdout }}/docker"
-    name: appliance-build
-    force: true
+#
+# We can't use the docker_image module because it doesn't yet support passing
+# the 'network' parameter: https://github.com/ansible/ansible/pull/50313, which
+# we need to be able to fetch things from Artifactory.
+#
+- shell: docker build --network host --tag "appliance-build:latest" "{{ toplevel.stdout }}/docker"
```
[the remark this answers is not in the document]
jgallag88: We need it so that we can reach other machines on the Delphix VPN, in particular, Artifactory. Do you feel that the comment that's there is incomplete?

README.md
```diff
@@ -106,50 +106,92 @@ correcting any deficencies that may exist. This is easily done like so:
...
 To exit "qemu", one can use "Ctrl-A X".
 
+## Building an Upgrade Image
+
+An upgrade image for a particular variant can be built by running the
+'buildUpgradeImage\<Variant\>' tasks. For instance, the task to build
+an upgrade image for the internal-minimal variant is
+'buildUpgradeImageInternalMinimal':
+
+    $ PLATFORMS='kvm aws' ./scripts/docker-run.sh gradle buildUpgradeImageInternalMinimal
```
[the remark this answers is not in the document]
jgallag88: I can certainly add something to `version.info`. At that point the image will be unpacked though, so maybe we could also just query to see which versions of `delphix-entire-` are in the apt repo. Either way though, I agree it would be good to have an explicit check for this failure mode so that we can give users a good error message explaining why they have hit this error and how they should proceed.

> I presume that we do this to allow faster builds in some circumstances, but we'll have to weigh that convenience with the possibility that we might accidentally build images for the wrong platforms

Faster builds and smaller upgrade images. Yes, there certainly is a trade off, but I think it's necessary for now unless we want to more or less give up on running the build outside of the jenkins job; running live build five times in a row to build the uprade image for a single variant is just too painful, IMO.

scripts/build-upgrade-image.sh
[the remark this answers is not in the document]
jgallag88: I can add the verification that the versions are all the same too, that seems worthwhile.

scripts/build-upgrade-image.sh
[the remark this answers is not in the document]
jgallag88: Oh it could be that @pzakha is right. It seems strange to me that apt would install the latest version of a package instead of the version specified (otherwise what's the point of being able to specify the version?), but I'm not sure. Anyway, if that is a problem, it seems like it might be a variation of the problem where we can't downgrade packages?

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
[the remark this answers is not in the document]
prakashsurya: Oh, that's right.. the Dockerfile needs to reach out to Artifactory now, which is why we need the network parameter during the image build. Sorry; I just overlooked that detail, the comment is fine.

upgrade/upgrade-scripts/execute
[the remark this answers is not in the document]
prakashsurya: I'm not 100% against installing `delphix-platform-$platform` prior to installing `delphix-entire-$platform`, but it does set a precedence that I'd like to avoid if we can.. since, this means if we ever want to create some other platform specific package(s), upgrade will need explicit knowledge about them. e.g. if we create some new package(s) `foo-aws`, `foo-esx`, etc. upgrade will have to know how to install the correct one, right?

Ultimately, I'd like to see if we can understand this behavior prior to coming to a decision.

> What I like about the current strategy though is that it makes the order in which the packages are installed more similar to the order used in live-build in appliance-build.

Yea, I agree with this in concept, but I'd argue the build is worse about this than upgrade. I'd rather see us revamp the build, to avoid the step-by-step implementation that it has now. If we don't change how the build works, though, I think this is a good argument for taking this approach.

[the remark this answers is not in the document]
jgallag88: Based on our offline findings and discussion, I've expanded the comment here substantially to explain exactly what goes wrong if we install delphix-entire-<platform> directly.

PR #168, opened 2018-12-06
XXX Transition to Ubuntu 18.10 (Cosmic Cuttlefish)

prakashsurya:


prakashsurya:
I don't intend to land this, but wanted to open a PR so that code is available when we decide to regularly build and test using Ubuntu 18.10.

PR #174, opened 2018-12-10
Enable more files to be migrated on upgrade

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/308/)

live-build/misc/upgrade-scripts/upgrade-container
```diff
@@ -365,25 +378,73 @@ function convert_to_bootfs() {
...
+	# system to properly operate; these must be preserved acorss
+	# upgrades.
+	#
+	while read -r file; do
+		migrate_file "$file"
+	done <<-EOF
+		/etc/hostname
+		/etc/hosts
+		/etc/machine-id
+		/etc/resolv.conf
```
pzakha: On my DE system this is a symlink to `./run/systemd/resolve/resolv.conf`. Does `migrate_file` handle symlinks? I See you've added `-a` to cp for `migrate_directory`, perhaps you need something similar for `migrate_file`?

prakashsurya: it should probably be commented better, but I think we actually don't want to migrate this file if it's the symlink (which is the default).

it won't be a symlink if the app-gate has modified it, to use some custom DNS settings, and that's when we need to do the migration.

otherwise, leaving it as not migrated will enable it to use the default systemd-resolved behavior. since we do the `-f` test in `migrate_file`, I think it won't copy anything when it's a symlink.

pzakha: I definitely think you should definitely add a comment. This holds in the case of `/etc/resolv.conf`, but might not necessarily hold for other configuration files/links, so it'd be important for that behaviour to be very obvious.

prakashsurya:
@dtzeng can you give this a look, and help to try and make sure I've captured all the files that need to be migrated?

live-build/misc/upgrade-scripts/upgrade-container
```diff
@@ -365,25 +378,90 @@ function convert_to_bootfs() {
...
+		/etc/ssh/ssh_host_dsa_key.pub
+		/etc/ssh/ssh_host_ecdsa_key
+		/etc/ssh/ssh_host_ecdsa_key.pub
+		/etc/ssh/ssh_host_ed25519_key
+		/etc/ssh/ssh_host_ed25519_key.pub
+		/etc/ssh/ssh_host_rsa_key
+		/etc/ssh/ssh_host_rsa_key.pub
+		/etc/sudoers.d/90-cloud-init-users
+		/etc/netplan/50-cloud-init.yaml
+	EOF
```
pzakha: Might also want to add `/etc/hostid` and `/etc/zfs/zpool.cache`.

prakashsurya: what is `/etc/hostid`? I don't see it on my linux system, and a quick grep in the virtualization code leads me to think it's illumos specific..?

prakashsurya: nvm.. I think I remember, it's created by the zfs commands, right?

pzakha: It is used by ZFS to identify the machine you are on. This file is not there by default on Linux. It's unclear to me if we will need to use it or not, but right now we have to do a work around when importing rpool from Illumos during migration (we must call `zpool import` with `-f`). The hostid property is stored on the pool (you can run `zdb -CC rpool` on Illumos). Note that it's not there on pools created on Linux. The issue is that on Linux, if `/etc/hostid` is not there then the `hostid` command returns some synthetic value based on an IP on your system (for some reason it defaults to the loopback IP, i.e. `007f0101`).

prakashsurya:
I wasn't sure if `/etc/fluent/fluent.conf` is present on linux or not, but I'll add it to the list. I think I've captured all the other files we need on Linux.

@sebroy do you think you can give this a look too? and maybe double check the OS specific files, to help ensure I didn't missing anything..

PR #175, opened 2018-12-11
Reload and restart delphix-platform during upgrade

prakashsurya:
This updates the upgrade logic to account for some recent changes to the
delphix-platform service. Now, we need to reload the service to ensure
the service will do useful work the next time it runs. Additionally, we
need to restart the service rather than start the service; otherwise, if
the service is already running (it should be), starting the service is
basically a noop.


prakashsurya:
`git-ab-pre-push -v internal-dev` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/314/)

PR #176, opened 2018-12-12
Enable automounting of /net on the internal-dev variant

sebroy:
Testing:
ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/310/flowGraphTable/

I cloned the resulting dcoa group into a VM and verified that the /net/ autofs filesystem was mountable:
`delphix@ip-10-110-216-115:~$ ls /net/wirefox/data/seb
dlpx-app-gate  dlpx-os-gate  dlpx-pkg-gate  ESCL-2156  jenkins  pkg-build-gate`


live-build/misc/ansible-roles/appliance-build.delphix-autofs/files/net.autofs
```diff
@@ -0,0 +1 @@
+/net	/etc/auto.net
```
prakashsurya: A general convention in ansible roles is to structure the "files" directory similar to where the files will be placed on the root filesystem. E.g. in this example, we'd have:
```
appliance-build.delphix-autofs/files/etc/autofs.master.d/net.autofs
```
then we can do this in the `main.yml` file to copy all files in that "files" directory:
```
- copy:                                                                         
    src: etc/                                                                   
    dest: /etc/                                                                 
    owner: root                                                                 
    group: root                                                                 
    mode: 0444  
```

with that said, it doesn't _really_ matter (this works), and we're already breaking that convention in at least one other place in this repository, but I just wanted to mention it.

basil: [The `devops-gate` version](https://gitlab.delphix.com/devops/devops-gate/blob/master/ansible/roles-shared/dlpx.common-ubuntu/tasks/autofs.yml) has `/net    -hosts`. I'm curious what the difference is between that and the version you have here. Should we change the `devops-gate` version to use `/net /etc/auto.net`?

sebroy: Yes, both of these suggestions make sense to me, I'll update shortly.

sebroy: It looks like the `auto.net` version executes the shell script there to populate the map, and the `-hosts` version is built-in to the automount daemon. I'll switch this commit to use the `-hosts` version since that's the one that's documented in the `auto.master` man page.

sebroy: On the enabling of debug logging, if I recall, we did have an issue a few years back associated with auto-mounting home directories from pharos which was root-caused to a bug in the NFSv4 client on Illumos (not sure if it affected automounts on Linux images). Switching to NFSv3 mounts fixed that. I don't recall other issues that would have required client-side auto-mounter debugging. I'll leave this off for now unless we're actively chasing down some issue.

PR #180, opened 2018-12-13
Migrate configuration prior to running "verify-impl"

prakashsurya:
We need to ensure the correct configuration is in place, prior to
executing the upgrade verification logic (i.e. prior to running
"verify-impl"). This change updates the "upgrade" script to do this.
Additionally, this only applies to "not-in-place" upgrades, since we
don't do any configuration migration for "in-place" upgrades.


prakashsurya:
`git-ab-pre-push -v internal-dev` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/330/)

cc: @dtzeng

PR #182, opened 2018-12-17
LX-1337 Allow app-gate stack to run in a chroot (systemd-nspawn) environment

pandar00:
### Changes

Adds stack startup verification logic.

* The stack startup verification logic will run only if it's a reboot required upgrade (i.e. when current version is less than `MINIMUM_REBOOT_OPTIONAL_VERSION`. Once a container packages are upgraded to the latest version, we don't have a way to get the previous version to compare against `MINIMUM_REBOOT_OPTIONAL_VERSION`. So I simply save the pre-upgrade version to `${IMAGE_PATH}/${UPGRADE_VERSION_FILE_NAME}` before running `execute` script and read the content before verifying.

* In order to run stack within a container, we need have a copy of MDS instance. In order to do so, we leverage `dx_manage_pg.sh` script to make a copy of MDS, start and stop the instance. Once MDS clone is prepared, we need to create new links to `/var/delphix/server/db` and `/var/delphix/server/mds_external` since stack depends on these location. We are also leveraging `start_mgmt_server_jvm` script to run management stack. This script will return non-zero exit status if it fails to start.

### DIFF
--- a/upgrade/upgrade-scripts/upgrade
+++ b/upgrade/upgrade-scripts/upgrade
@@ -113,8 +113,11 @@ function upgrade_in_place() {
 
 	if [[ "$DLPX_UPGRADE_SKIP_VERIFY" != "true" ]]; then
 		"$IMAGE_PATH/upgrade-container" run "$CONTAINER" \
-			"$IMAGE_PATH/verify-impl" "$@" ||
-			die "'$IMAGE_PATH/verify-impl' '$*' failed in '$CONTAINER'"
+			"$IMAGE_PATH/verify-jar" "$@" ||
+			die "'$IMAGE_PATH/verify-jar' '$*' failed in '$CONTAINER'"
+		"$IMAGE_PATH/upgrade-container" run "$CONTAINER" \
+			"$IMAGE_PATH/verify-stack" "$@" ||
+			die "'$IMAGE_PATH/verify-stack' '$*' failed in '$CONTAINER'"
 	fi
 
 	"$IMAGE_PATH/execute" -p "$(get_platform)" ||
@@ -180,8 +183,8 @@ function upgrade_not_in_place() {
 
 	if [[ "$DLPX_UPGRADE_SKIP_VERIFY" != "true" ]]; then
 		"$IMAGE_PATH/upgrade-container" run "$CONTAINER" \
-			"$IMAGE_PATH/verify-impl" "$@" ||
-			die "'$IMAGE_PATH/verify-impl' '$*' failed in '$CONTAINER'"
+			"$IMAGE_PATH/verify-jar" "$@" ||
+			die "'$IMAGE_PATH/verify-jar' '$*' failed in '$CONTAINER'"
 	fi
 
 	"$IMAGE_PATH/upgrade-container" stop "$CONTAINER" ||
--- a/upgrade/upgrade-scripts/verify
+++ b/upgrade/upgrade-scripts/verify
@@ -41,6 +41,15 @@ function cleanup() {
 	#
 	[[ -z "$CONTAINER" ]] && return
 
+	#
+	# Collect logs within container if it's running.
+	#
+	if [[ "$(machinectl list | grep -q "${CONTAINER}")" -eq 0 ]]; then
+		report_progress 75 "Collecting container logs into '${LOG_DIRECTORY}/${CONTAINER}'."
+		"$IMAGE_PATH/upgrade-container" run "$CONTAINER" "/opt/delphix/server/bin/support_info" \
+			-o "${LOG_DIRECTORY}/${CONTAINER}" -t "log dropbox"
+	fi
+
 	if [[ "$rc" == "0" || "$DLPX_DEBUG" != "true" ]]; then
 		report_progress 80 "Performing upgrade verification cleanup"
 
@@ -68,6 +77,9 @@ function cleanup() {
 
 trap cleanup EXIT
 
+VERSION=$(get_installed_version)
+source_version_information
+
 report_progress 0 "Creating upgrade verification container"
 CONTAINER=$("$IMAGE_PATH/upgrade-container" create in-place)
 [[ -n "$CONTAINER" ]] || die "failed to create container"
@@ -83,8 +95,21 @@ report_progress 40 "Performing package upgrade verification"
 
 report_progress 60 "Performing application upgrade verification"
 "$IMAGE_PATH/upgrade-container" run "$CONTAINER" \
-	"$IMAGE_PATH/verify-impl" "$@" ||
-	die "'$IMAGE_PATH/verify-impl' '$*' failed in '$CONTAINER'"
+	"$IMAGE_PATH/verify-jar" "$@" ||
+	die "'$IMAGE_PATH/verify-jar' '$*' failed in '$CONTAINER'"
+
+#
+# We skip stack startup verification if upgrade is a reboot-required upgrade
+# (i.e. 'current version' > 'minimum reboot optional version') because reboot-required upgrade
+# implies that stack is dependent on bits that are available post-reboot, so starting a
+# stack pre-reboot may result in dependency failure.
+#
+if compare_versions "$VERSION" "ge" "${MINIMUM_REBOOT_OPTIONAL_VERSION}"; then
+	report_progress 70 "Performing post-upgrade application startup upgrade verification"
+	"$IMAGE_PATH/upgrade-container" run "$CONTAINER" \
+		"$IMAGE_PATH/verify-stack" "$@" ||
+		die "'$IMAGE_PATH/verify-stack' '$*' failed in '$CONTAINER'"
+fi
 
 #
 # The cleanup logic will be run on EXIT, so rather than reporting 100
--- a/upgrade/upgrade-scripts/verify-jar
+++ b/upgrade/upgrade-scripts/verify-jar
@@ -26,13 +26,24 @@ function usage() {
 	exit 2
 }
 
-function run_support_info() {
+function verify_jar_verify_cleanup() {
 	local rc="$?"
 
-	/opt/delphix/server/bin/support_info \
-		-o "${LOG_DIRECTORY}/${CONTAINER}" \
-		-t "log dropbox" || die "Failed to collect verify logs."
+	#
+	# This name is used by the "upgrade-verify.jar" just executed, so we
+	# cannot change this value without also modifying that JAR.
+	#
+	MDS_SNAPNAME="MDS-CLONE-upgradeverify"
 
+	if /opt/delphix/server/bin/dx_manage_pg isrunning -s ${MDS_SNAPNAME}; then
+		/opt/delphix/server/bin/dx_manage_pg stop -s "${MDS_SNAPNAME}" ||
+			die "failed to stop postgres for snapshot '${MDS_SNAPNAME}'."
+	fi
+
+	if zfs list "${MDS_SNAPNAME}" &>/dev/null; then
+		/opt/delphix/server/bin/dx_manage_pg cleanup -s "${MDS_SNAPNAME}" ||
+			die "failed to cleanup postgres for snapshot '${MDS_SNAPNAME}'."
+	fi
 	return "$rc"
 }
 
@@ -55,16 +66,7 @@ if $opt_d; then
 	VERIFY_LIVE_MDS_OPT="-disableConsistentMdsZfsDataUtil"
 fi
 
-#
-# We use this directory to store the upgrade verification logs of this
-# container. This directory is accessible from both the container, and
-# the host. This allows the software running on the host to easily
-# access the logs after the verification is performed.
-#
-mkdir -p "${LOG_DIRECTORY}/${CONTAINER}" ||
-	die "failed to create directory: '${LOG_DIRECTORY}/${CONTAINER}'"
-
-trap run_support_info EXIT
+trap verify_jar_verify_cleanup EXIT
 
 /usr/bin/java \
 	-Dlog.dir=/var/delphix/server/upgrade-verify \
@@ -78,17 +80,5 @@ trap run_support_info EXIT
 	-pl 60 -ph 80 \
 	$VERIFY_LIVE_MDS_OPT ||
 	die "'upgrade-verify.jar' failed in verification container"
-
-#
-# This name is used by the "upgrade-verify.jar" just executed, so we
-# cannot change this value without also modifying that JAR.
-#
-MDS_SNAPNAME="MDS-CLONE-upgradeverify"
-
-/opt/delphix/server/bin/dx_manage_pg stop -s "$MDS_SNAPNAME" ||
-	die "failed to stop postgres"
-
-/opt/delphix/server/bin/dx_manage_pg cleanup -s "$MDS_SNAPNAME" ||
-	die "failed to cleanup postgres"
 
 exit 0
--- a/upgrade/upgrade-scripts/verify-stack
+++ b/upgrade/upgrade-scripts/verify-stack
@@ -1,0 +1,64 @@
+#!/bin/bash
+#
+# Copyright 2018 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+. "${BASH_SOURCE%/*}/common.sh"
+
+STACK_STARTUP_MDS_SNAPNAME="MDS-CLONE-stack-startup-verify"
+
+function stack_startup_verify_cleanup() {
+	local rc="$?"
+
+	if /opt/delphix/server/bin/dx_manage_pg isrunning -s ${STACK_STARTUP_MDS_SNAPNAME}; then
+		/opt/delphix/server/bin/dx_manage_pg stop -s "${STACK_STARTUP_MDS_SNAPNAME}" ||
+			die "failed to stop postgres for snapshot '${STACK_STARTUP_MDS_SNAPNAME}'."
+	fi
+
+	if zfs list | grep -q ${STACK_STARTUP_MDS_SNAPNAME}; then
+		/opt/delphix/server/bin/dx_manage_pg cleanup -s "${STACK_STARTUP_MDS_SNAPNAME}" ||
+			die "failed to cleanup postgres for snapshot '${STACK_STARTUP_MDS_SNAPNAME}'."
+	fi
+
+	return "$rc"
+}
+
+trap stack_startup_verify_cleanup EXIT
+
+#
+# Make sure delphix-mgmt and delphix-postgres aren't running within container.
+#
+IS_ACTIVE_MGMT=$(systemctl is-active delphix-mgmt)
+[[ "${IS_ACTIVE_MGMT}" == "inactive" ]] ||
+	die "delphix-mgmt service expected to be 'inactive', found to be '${IS_ACTIVE_MGMT}'"
+IS_ACTIVE_POSTGRES=$(systemctl is-active delphix-postgres@default)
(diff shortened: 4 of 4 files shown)


pandar00:
cc: @prakashsurya @dtzeng

live-build/misc/upgrade-scripts/verify-impl
```diff
@@ -44,6 +44,81 @@ function cleanup() {
...
+
+	#
+	# Edit Postgres systemd.service file to run within container.
+	#
+	local delphix_postgres_service='/lib/systemd/system/delphix-postgres@.service'
+	sed -i 's/ConditionVirtualization/# &/' ${delphix_postgres_service}
+	systemctl daemon-reload
+
+	#
+	# We need to make a copy of MDS within container to avoid writing data to the host MDS.
```
prakashsurya: Why not create a clone of MDS, and use the clone, similar to what the upgrade-verify JAR does? Would that work? A clone would likely be more efficient to create, and clean up.

live-build/misc/upgrade-scripts/verify-impl
```diff
@@ -44,6 +44,81 @@ function cleanup() {
...
+	[[ $(systemctl is-active delphix-mgmt) == 'inactive' ]] ||
+		die "delphix-mgmt service is not inactive."
+	[[ $(systemctl is-active delphix-postgres@default.service) == 'inactive' ]] ||
+		die "delphix-postgres@default.service service is not inactive."
+
+	#
+	# Edit Postgres systemd.service file to run within container.
+	#
+	local delphix_postgres_service='/lib/systemd/system/delphix-postgres@.service'
+	sed -i 's/ConditionVirtualization/# &/' ${delphix_postgres_service}
```
prakashsurya: We should use `|| die` here.

Would it be possible to incorporate all of this MDS manipulation logic into the postgres service, itself? E.g. have the service detect that it's running in a container, then clone/copy the data, then start itself using that clone? This we could just remove this condition from the service's definition, and let it start up while in the container.

It just seems awkward to do this in the verification script, when it's dependent on the internal workings of the service, and is prone to break if/when the service changes.

I'm not necessarily against doing it this way, but I'd be in favor of coming up with a better solution that didn't expose so much of the app-gate service's internals. Ideally, I think we could just rely on the service starting correctly when it's in the container, and then pointing the app-gate stack at this "containerized" postgres instance.

If you'd rather go with this for now, I'm OK with that, but I'd like to at least open a bug to investigate if we could reduce the coupling between the service and this script.

Another thing that I'm worried about, is which port will this postgres instance use? Presumably it'll use the default port, which is the same as the "production" MDS instance? If so, I think this will work now, because the container is configured to have a separate/isolated network stack than the host. I'm not sure we want to do this long term, though, since I think the container will need to be able to reach hosts external to the DE (e.g. so it can reach target environments). So we might want to be proactive, and ensure this second MDS instance uses a non-default port, in anticipation of the container sharing the network with the host.

live-build/misc/upgrade-scripts/verify-impl
```diff
@@ -44,6 +44,81 @@ function cleanup() {
...
+	systemctl daemon-reload
```
prakashsurya: we should use `|| die` here.

live-build/misc/upgrade-scripts/verify-impl
```diff
@@ -44,6 +44,81 @@ function cleanup() {
...
+
+	#
+	# We need to make a copy of MDS within container to avoid writing data to the host MDS.
+	#
+	local source_db='/mds/db'
+	local source_mds_external='/mds/mds_external'
+	local destination='/var/delphix/server'
+	rm ${destination}/db ${destination}/mds_external
```
prakashsurya: we should use `|| die` here.

do we want/need to use `-r` or `-f` here?

live-build/misc/upgrade-scripts/verify-impl
```diff
@@ -44,6 +44,81 @@ function cleanup() {
...
+	cp -rp ${source_db} ${destination} || die "Failed to copy '${source_db}' to '${destination}'."
```
prakashsurya: should we clean these up after we're done with them? if we keep them around, and we don't end up using a zfs clone, would it be worth using `rsync` to re-use the files that haven't changed? How much data may we copy with this `cp` and the one below?

live-build/misc/upgrade-scripts/verify-impl
```diff
@@ -44,6 +44,81 @@ function cleanup() {
...
+	local delphix_postgres_service='/lib/systemd/system/delphix-postgres@.service'
+	sed -i 's/ConditionVirtualization/# &/' ${delphix_postgres_service}
+	systemctl daemon-reload
+
+	#
+	# We need to make a copy of MDS within container to avoid writing data to the host MDS.
+	#
+	local source_db='/mds/db'
+	local source_mds_external='/mds/mds_external'
+	local destination='/var/delphix/server'
```
prakashsurya: since the destination isn't specific to this verification invocation, I think concurrent verifies could corrupt each other, right? i.e. if we kick of two verifies concurrently via the DE or the CLI? is it possible to kick of concurrent verify's via the DE API?

prakashsurya: Although, now that I think about it, the `/var/delphix` directory here is in the container, and it'll already be specific to this verification invocation, right? so maybe we don't have to worry about concurrently? since each verify will have a unique `/var/delphix` directory?

live-build/misc/upgrade-scripts/verify-impl
pandar00: Good call. I've made change to leverage `dx_manage_pg.sh` script to start and stop a copy of MDS. The script creates a ZFS of snapshot and uses that to start the instance internally.

live-build/misc/upgrade-scripts/verify-impl
pandar00: > Would it be possible to incorporate all of this MDS manipulation logic into the postgres service, itself?
New revision leverages `dx_manage_pg.sh` script to manipulate a copy of MDS. This script internally creates snapshot, systemd service definition, starts, stops postgres internally.

> Another thing that I'm worried about, is which port will this postgres instance use?
Ah. I haven't though about this, but I'm hard-coding `dx_manage_pg.sh` script to use default port of 5432. I agree that it'll likely be desirable that we want to start app-stack with custom postgres port but It seems that it's set during the build phase. I've filed https://jira.delphix.com/browse/DLPX-62625 to track this work.

live-build/misc/upgrade-scripts/verify-impl
pandar00: New revision does not changes systemd service definitions directly, so the need to do `sed` or `systemctl daemon-reload` is eliminated.

live-build/misc/upgrade-scripts/verify-impl
pandar00: The app-stack allows running only one verification at a time, and I think you are right about contents of `/var/delphix` being unique to each verification invocation.

upgrade/upgrade-scripts/verify-impl
```diff
@@ -36,6 +37,94 @@ function run_support_info() {
...
+		"${VERIFY_LIVE_MDS_OPT}" ||
+		die "'upgrade-verify.jar' failed in verification container"
+
+	#
+	# This name is used by the "upgrade-verify.jar" just executed, so we
+	# cannot change this value without also modifying that JAR.
+	#
+	MDS_SNAPNAME="MDS-CLONE-upgradeverify"
+
+	# TODO : Make sure MDS clone gets cleaned up if verification fails.
```
pandar00: TODO

upgrade/upgrade-scripts/verify-impl
```diff
@@ -36,6 +37,94 @@ function run_support_info() {
...
+	#
+	# Start management stack up to post-upgrade logic.
+	#
+	/opt/delphix/server/bin/start_mgmt_server_jvm -f /var/run/delphix-mgmt.pid ||
+		die "Stack startup verification failed."
+
+	#
+	# Cleanup MDS clone.
+	#
+	# TODO : Make sure MDS clone gets cleaned up if verification fails.
```
pandar00: TODO

upgrade/upgrade-scripts/verify-impl
prakashsurya: I think we can accomplish the cleanup on failure using another "cleanup" function with `trap`.. e.g.

```
function verify_run_verify_jar_cleanup() {
    if clone_is_running; then
        /opt/delphix/server/bin/dx_manage_pg stop ...
    fi
    if clone_exists; then
        /opt/delphix/server/bin/dx_manage_pg cleanup ...
    fi
    /opt/delphix/server/bin/support_info ...
}

function verify_run_verify_jar {
    trap verify_run_verify_jar_cleanup EXIT
    ... # run JAR
    /opt/delphix/server/bin/dx_manage_pg stop ...
    /opt/delphix/server/bin/dx_manage_pg cleanup ...
    trap - EXIT
}
```

we'd then need similar "cleanup" logic for the stack startup function:
```
function verify_stack_startup_cleanup {
    if ...; then
        /opt/delphix/server/bin/dx_manage_pg stop ...
    fi
    if ...; then
        /opt/delphix/server/bin/dx_manage_pg cleanup ...
    fi
    /opt/delphix/server/bin/support_info ...
}

function verify_stack_startup {
    trap verify_stack_startup_cleanup EXIT
    ... # run stack startup logic
    /opt/delphix/server/bin/dx_manage_pg stop ...
    /opt/delphix/server/bin/dx_manage_pg cleanup ...
    trap - EXIT
}
```

And then we'd probably also remove the line:
```
trap run_support_info EXIT
```
and the `run_support_info` function since we'd be running the support info script from the other "cleanup" functions.

upgrade/upgrade-scripts/verify-impl
```diff
@@ -36,6 +37,94 @@ function run_support_info() {
...
+}
+
+function verify_stack_startup() {
+	echo "Verify - Run stack startup."
+
+	#
+	# Make sure delphix-mgmt and delphix-postgres aren't running within container.
+	#
+	[[ $(systemctl is-active delphix-mgmt) == 'inactive' ]] ||
+		die "delphix-mgmt service is not inactive."
```
prakashsurya: Would it be better to capture the result of `systemctl is-active` to a variable, so we can emit the value on failure? e.g.
```
IS_ACTIVE_MGMT=$(systemctl is-active delphix-mgmt)
[[ "$IS_ACTIVE_MGMT" == "inactive" ]] ||
    die "delphix-mgmt service expected to be 'inactive', found to be '$IS_ACTIVE_MGMT'"
```
this might help us better understand a failure more quickly when/if it fails..

same for the postgres check below..

upgrade/upgrade-scripts/verify-impl
```diff
@@ -36,6 +37,94 @@ function run_support_info() {
...
+	# Create a copy of MDS and start the postgres service.
+	#
+	STACK_STARTUP_MDS_SNAPNAME="MDS-CLONE-stack-startup-verify"
+	/opt/delphix/server/bin/dx_manage_pg clone -s "${STACK_STARTUP_MDS_SNAPNAME}" ||
+		die "failed to clone postgres"
+	/opt/delphix/server/bin/dx_manage_pg start -s "${STACK_STARTUP_MDS_SNAPNAME}" -p 5432 ||
+		die "failed to start postgres"
+
+	#
+	# Modify application stack files to point to copy of MDS.
```
prakashsurya: Can you also leave a comment about why we need to do this?

Could we encapsulate this logic into a script delivered by the app-gate? At first glance it seems like very specific to the app-gate internals, so it'd be nice if we can avoid leaking these details into this script.. at the same time, this script is already very specific to the app-gate, so it's not that big of a deal.

upgrade/upgrade-scripts/verify-impl
```diff
@@ -55,40 +144,12 @@ if $opt_d; then
...
-#
-MDS_SNAPNAME="MDS-CLONE-upgradeverify"
+verify_run_verify_jar
 
-/opt/delphix/server/bin/dx_manage_pg stop -s "$MDS_SNAPNAME" ||
-	die "failed to stop postgres"
-
-/opt/delphix/server/bin/dx_manage_pg cleanup -s "$MDS_SNAPNAME" ||
-	die "failed to cleanup postgres"
+if is_stack_only_upgrade; then
```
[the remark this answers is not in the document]
pandar00: If I recall correctly, reboot required upgrades have hard dependency between the app-stack and bits that are available once reboot happens, and because we are verifying before reboot has happened, we cannot verify stack stack in those cases.

upgrade/upgrade-scripts/verify-impl
pandar00: Sounds good to me. I'll move these logic into `dx_manage_pg.sh` script.

upgrade/upgrade-scripts/verify-impl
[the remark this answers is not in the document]
prakashsurya: Yea, I see what you mean.

Currently, the incompatibility you're worried about doesn't exist, but I guess it could exist at some point in the future.

If it was me, I'd opt for the most simple solution, and just run stack startup in all cases, since practically it should work for all cases right now. Then, if startup didn't work for the "reboot required" case later, we could do something like what you have here.

I just feel like since we don't actually have an upgrade scenario where this is a problem, we likely won't get good test coverage of this exception case, so I'm worried the additional complexity will come without any additional benefit. And the complexity can regress in a later change, and we might not notice, since we won't have this code path regularly tested.

But, if you think it'd be better to do the conditional now, I'm OK with that too.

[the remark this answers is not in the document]
prakashsurya: At the very least, if we keep the conditional, I think we should add a comment that adequately explains the situation.. i.e. that we can't reliably start the new stack on the current kernel configuration for "reboot required" upgrades, so we skip that step here.

upgrade/upgrade-scripts/verify-impl
pandar00: Logic is removed to `dx_manage_pg.sh` script. It'll now take a extra `-l` parameter with `clone` command to replace symlinks after it clones MDS.

live-build/misc/upgrade-scripts/verify-impl
pandar00: We no longer copies files directly. Marking as resolved.

upgrade/upgrade-scripts/verify
```diff
@@ -76,6 +94,12 @@ report_progress 20 "Starting upgrade verification container"
 "$IMAGE_PATH/upgrade-container" start "$CONTAINER" ||
 	die "failed to start '$CONTAINER'"
 
+#
+# Save pre-upgrade version so that verification logic can use this to determine
+# which verification(s) to run.
+#
+save_upgrade_from_version
```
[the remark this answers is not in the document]
prakashsurya: We also need to ensure we maintain verification in the `upgrade` script, which calls `verify-impl`. I haven't looked at what changes will need to be made there, but I can do that after we solidify the implementation here.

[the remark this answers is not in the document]
pandar00: Your suggestion looks great to me. I'll make changes

upgrade/upgrade-scripts/verify-impl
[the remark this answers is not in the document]
pandar00: This comment will be resolved as part of addressing comment about splitting `verify-impl` into `verify-jar` and `verify-stack`

upgrade/upgrade-scripts/verify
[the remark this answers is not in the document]
pandar00: Ok, I think I addressed all of your comments, major change being splitting the `verify-impl` script into two: `verify-stack` and `verify-jar`. 

At high-level, these are the responsibilities of each script:
* `verify`
    * Starts up the container and does upgrade verifications. The stack startup verification is conditionally selected based on the current version and target versions minimum reboot optional version/
    * Collects logs within container on exit only if container is running.
* `verify-jar`
    * Runs `upgrade-verify.jar`.
* `verify-stack`
    * Runs stack startup verifications.

Please take a look when you get chance. Nothing urgent.
@dtzeng @prakashsurya

upgrade/upgrade-scripts/upgrade
```diff
@@ -113,8 +113,11 @@ function upgrade_in_place() {
 
 	if [[ "$DLPX_UPGRADE_SKIP_VERIFY" != "true" ]]; then
 		"$IMAGE_PATH/upgrade-container" run "$CONTAINER" \
-			"$IMAGE_PATH/verify-impl" "$@" ||
-			die "'$IMAGE_PATH/verify-impl' '$*' failed in '$CONTAINER'"
+			"$IMAGE_PATH/verify-stack" "$@" ||
```
prakashsurya: I think this should be `verify-jar` rather than `verify-stack`? same for the error message one line below.

upgrade/upgrade-scripts/verify-jar
```diff
@@ -26,12 +26,22 @@ function usage() {
...
-	/opt/delphix/server/bin/support_info \
-		-o "${LOG_DIRECTORY}/${CONTAINER}" \
-		-t "log dropbox" || die "Failed to collect verify logs."
+	#
+	# This name is used by the "upgrade-verify.jar" just executed, so we
+	# cannot change this value without also modifying that JAR.
+	#
+	MDS_SNAPNAME="MDS-CLONE-upgradeverify"
+
+	[[ "$(/opt/delphix/server/bin/dx_manage_pg isrunning -s ${MDS_SNAPNAME})" -eq 0 ]] &&
```
prakashsurya: This is just a style comment, but the `[[` isn't necessary.. e.g.
```
/opt/delphix/server/bin/dx_manage_pg isrunning -s "$MDS_SNAPNAME" &&
```
should work the same.

For all of the `if` conditionals, I don't think you need to use `[[ ... -eq 0 ]]`, you can just use the command with `if` and/or `&&` directly.

I'm fine either way, I think I'm just accustomed to not seeing `[[` in lots of our other scripts (e.g. in the app-gate).

upgrade/upgrade-scripts/verify-jar
```diff
@@ -26,12 +26,22 @@ function usage() {
...
+		(/opt/delphix/server/bin/dx_manage_pg stop -s "${MDS_SNAPNAME}" ||
+			die "failed to stop postgres for snapshot '${MDS_SNAPNAME}'.")
+
+	[[ "$(zfs list | grep -q ${MDS_SNAPNAME})" -eq 0 ]] &&
```
prakashsurya: can we omit the `grep` and just use this:
```
zfs list "$MDS_SNAPNAME" &>/dev/null &&
```
?

upgrade/upgrade-scripts/verify-jar
```diff
@@ -26,12 +26,22 @@ function usage() {
...
-		-t "log dropbox" || die "Failed to collect verify logs."
+	#
+	# This name is used by the "upgrade-verify.jar" just executed, so we
+	# cannot change this value without also modifying that JAR.
+	#
+	MDS_SNAPNAME="MDS-CLONE-upgradeverify"
+
+	[[ "$(/opt/delphix/server/bin/dx_manage_pg isrunning -s ${MDS_SNAPNAME})" -eq 0 ]] &&
+		(/opt/delphix/server/bin/dx_manage_pg stop -s "${MDS_SNAPNAME}" ||
+			die "failed to stop postgres for snapshot '${MDS_SNAPNAME}'.")
```
prakashsurya: This is spawning a subshell, right? I don't think `die` will cause the calling script to abort when it's used in a subshell. IIRC, it'll abort the subshell, but the calling process will continue.

upgrade/upgrade-scripts/verify-jar
```diff
@@ -26,12 +26,22 @@ function usage() {
...
+
+	[[ "$(zfs list | grep -q ${MDS_SNAPNAME})" -eq 0 ]] &&
+		(/opt/delphix/server/bin/dx_manage_pg cleanup -s "${MDS_SNAPNAME}" ||
+			die "failed to cleanup postgres for snapshot '${MDS_SNAPNAME}'.")
```
prakashsurya: Similar to the above, I don't think `die` will abort the parent process, so I'm not sure if this will behave as you expect.

pandar00:
@prakashsurya , @dtzeng 
It would be nice if you guys could take a look at this one last time.

PR #167, opened 2018-12-06
LX-1551, LX-1553 - dx_apply & dx_execute 

sdimitro:
LX-1551 dx_apply for migration
    LX-1553 dx_execute for migration

    This patch introduces the next 2 migration scripts:
    dx_apply and dx_execute.

    dx_apply: creates the dataset hierarchy expected by linux upgrade
    and unpacks the migration image within it. It also adds a new
    entry in the FreeBSD bootloader so a VM can optionally boot from
    Linux. Note that the script has a few WIP comments that are there
    for 2 WIP issues. The first one is the fact that we can't do any
    checks about running on the 5.3 release because of the feature
    incompatibility of ZoL and our illumos pools (this is tracked by
    LX-1633). The other one is a placeholder for the point of execution
    where the migration scripts will run.

    dx_execute: checks that dx_apply has left around all the expected
    files and info and configures the bootloader to boot into Linux
    next time the system boots.

    Side-note - This patch also fixes issue #170.

    Testing:

    In a DRYRUN-5.3.2.0 VM:
    ```
    $ sudo /opt/delphix/server/bin/upgrade/dx_unpack internal-qa.migration.tar.gz
    Progress increment: 17:52:10:272024270+0000, 10, Extracting upgrade image
    17:52:10:274636821:+0000: Unpacking internal-qa.migration.tar.gz ...

    18:02:32:771501160:+0000: done.
    Progress increment: 18:02:32:774283991+0000, 40,  verifying format
    Progress increment: 18:02:32:802230044+0000, 50,  verifying signature
    18:02:32:804885537:+0000: Verify signature ... Verified OK
    Progress increment: 18:02:32:940595417+0000, 70, Verifying integrity of upgrade image
    18:02:32:943585708:+0000: Verifying contents of ./dx_apply ... ./dx_apply: OK
    18:02:32:959290157:+0000: Verifying contents of ./dx_execute ... ./dx_execute: OK
    18:02:32:973580872:+0000: Verifying contents of ./dx_prepare ... ./dx_prepare: OK
    18:02:32:988391332:+0000: Verifying contents of ./dx_verify ... ./dx_verify: OK
    18:02:33:005307726:+0000: Verifying contents of ./os-root.cpio ... ./os-root.cpio: OK
    18:04:28:832789838:+0000: Verifying contents of ./os-root.hashes ... ./os-root.hashes: OK
    18:04:28:970944902:+0000: Verifying contents of ./version.info ... ./version.info: OK
[...]

PR #190, opened 2019-01-11
Make migration image useable by UpgradeManager

pzakha:
This follows up on https://github.com/delphix/appliance-build/pull/167 and makes those scripts consumable by UpgradeManager.
The following changes had to be made:
- Add progress report to `dx_apply` and `dx_verify`
- Add a valid date field to `version.info`
- Reboot the system at the end of `dx_execute`

## TESTING
1. Run a 5.3 system through initial setup (Using `DRYRUN-migration-5.2.3.0` image)
1. Download migration image from S3 into /var/dlpx-update/upload
2. Remove `rpool/update` quota (`zfs set quota=none rpool/update`) as the migration image is very large and fails to unpack otherwise.
3. Run `dx_unpack` on the migration image
4. Go into the UI and click "Verify Upgrade" -- PASS.
5. Go into the UI and click "Apply Upgrade" -- This will apply the image and reboot into Linux. Note that domain0 is currently not importable in RW mode due to missing ZFS features.

### DIFF
--- a/live-build/misc/migration-scripts/dx_execute
+++ b/live-build/misc/migration-scripts/dx_execute
@@ -27,6 +27,27 @@ function die() {
 	exit 1
 }
 
+function usage() {
+	[[ -n "$1" ]] && echo "$1"
+	cat <<EOM
+Usage:
+  $(basename "$0") [-s]"
+    Reset system to the Linux version that has been previously installed using
+    dx_apply.
+      -s Shutdown (instead of reboot) after upgrade.
+EOM
+	exit 2
+}
+
+opt_s=false
+while getopts :hs c; do
+	case "$c" in
+	\?) usage "Invalid option: -$OPTARG." ;;
+	h) usage ;;
+	s) eval "opt_$c=true" ;;
+	esac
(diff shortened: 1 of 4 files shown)


live-build/misc/migration-scripts/dx_execute
```diff
@@ -85,4 +106,16 @@ MAIN_MENU_LINUX_CMD=$(grep -F 'mainmenu_command[8]' /boot/menu.rc.local |
...
 echo "set menu_timeout_command=$MAIN_MENU_LINUX_CMD" >>/boot/menu.rc.local
 
+# Constants used by the uadmin syscall.
+A_SHUTDOWN=2
+AD_POWEROFF=6
+AD_BOOT=1
+
+# reboot or shutdown
+if $opt_s; then
+	uadmin $A_SHUTDOWN $AD_POWEROFF
```
sdimitro: Is there any case of those `uadmin` cases failing? and if yes, is there a way we'll know?

prakashsurya:
bors delegate+

PR #191, opened 2019-01-11
LX-1664 Verify jar results json file is not present in /var/delphix/dropbox/ in Linux

pandar00:
The verify.jar results were purged when container logs were collected because support_info script by default cleans up the dropbox directory. I made a change to support_info script to skip the cleanup if `-s` flag is provided. The change here simply consumes the flag.


pandar00:
@dtzeng @prakashsurya

pandar00:
Closing the review since verification result is no longer stored in dropbox directory.

PR #194, opened 2019-01-15
Embed "delphix-entire" version in rootfs dataset name

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/373/)

## TODO
- [ ] Update migration scripts

prakashsurya:
After a not-in-place upgrade, with this change, the rootfs looks like this:
```
$ zfs list -r $(dirname $(zfs list -Ho name /))
NAME                                  USED  AVAIL     REFER  MOUNTPOINT
rpool/ROOT/2019.1.14.0.1QgfcYN       6.78G  27.6G       64K  none
rpool/ROOT/2019.1.14.0.1QgfcYN/data  1.71M  27.6G     45.6M  /var/delphix
rpool/ROOT/2019.1.14.0.1QgfcYN/home   350K  27.6G     10.0G  /export/home
rpool/ROOT/2019.1.14.0.1QgfcYN/root  6.78G  27.6G     6.78G  /
```

prakashsurya:
cc: @grwilson @pzakha @sdimitro

live-build/misc/migration-scripts/dx_apply
```diff
@@ -107,7 +107,7 @@ report_progress_inc 20
 #
 # Create dataset layout similar to the linux upgrade.
 #
-TMPDIR=$(mktemp -d -p "/tmp" -t delphix.XXXXXXX)
+TMPDIR=$(mktemp -d -p "/tmp" -t "$DLPX_VERSION.XXXXXXX")
```
sdimitro: Thanks for changing the migration scripts too. Unfortunately Github doesn't allow me to highlight this but in this same script we always do some cleanup of intermediate states from previous runs before generating anything new and something worth noting is this line:
```
rm -rf /tmp/delphix.* ||
	die "failed to destroy old delphix temporary directories"
```

The problem is that the new naming scheme is not very easy to detect. We could try to match it with a regex in cleanup but that seems complicated and hacky. Clearing out `/tmp` is dangerous for other applications.

IMHO an acceptable way to go about it would be to first create a directory like `/tmp/delphix-migration` and then do `TMPDIR=$(mktemp -d -p "/tmp/delphix-migration" -t "$DLPX_VERSION.XXXXXXX")`.
This is just the temporary mountpoint after all.

Then during cleanup we can just do:
```
rm -rf /tmp/delphix-migration ||
	die "failed to destroy old delphix temporary directories"
```

What do others think?

In addition, I think we should also make sure that `$DLPX_VERSION` was sourced correctly and it is not empty.

prakashsurya: Would it better to do this cleanup on exiting the script, rather than doing the cleanup on the next invocation? We could use `trap` to help ensure it gets called, even on error; then we could remove the specific directory we created, and nothing else, right?

prakashsurya: rather than do the refactoring I was thinking of, I went ahead and did what you suggest.. it now uses the `/tmp/migration-scripts` directory.

pzakha:
Not a huge fan personally of that naming scheme. It's even longer than what we had before.

prakashsurya:
We can drop the trailing `.XXXXXXX` if we don't want to support installation of a given version more than once. I didn't do that here, but was thinking about doing that later.

I was also going to look into your prior suggestion of using `.N`, but hadn't gotten to that yet; assuming we want to support installing a given version multiple times.

Also note, the version on a customer system will be shorter than the version we use for internal builds.

PR #195, opened 2019-01-15
Use "/etc/fstab" to mount non-rootfs datasets

prakashsurya:
We now have non-rootfs datasets that are distinct from the root
filesystem dataset, but are tied to a specific root filesystem. Thus,
when we boot a specific root filesystem, we need to ensure we always
mount the correct non-rootfs datasets that are tied to that given root
filesystem.

This change attempts to accomplish this, by relying on the "/etc/fstab"
file contained in the root filesystem, to mount these non-rootfs
datasets. This way, the file can be unique to each root filesystem,
pointing to the specific non-rootfs datasets that're tied to that root
filesystem, and result in the correct non-rootfs datasets automatically
being mounted at boot time.


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/374/)

## TODO
- [ ] Verify build
- [ ] Verify upgrade verification
- [ ] Verify in-place upgrades
- [ ] Verify not-in-place upgrades
- [ ] Verify migration

prakashsurya:
cc: @grwilson @pzakha @sdimitro

@pzakha @sdimitro can either of you verify this works with migration? otherwise, I think it's good to go; it's working in my upgrade testing.

pzakha:
> @pzakha @sdimitro can either of you verify this works with migration? otherwise, I think it's good to go; it's working in my upgrade testing.

I'll verify it with migration.

pzakha:
Applied migration upgrade successfully. The expected datasets were mounted on Linux boot:
```
delphix@pzakha-migr-ab195:~$ mount | grep ROOT
rpool/ROOT/delphix.LqquSlN/root on / type zfs (rw,relatime,xattr,noacl)
rpool/ROOT/delphix.LqquSlN/home on /export/home type zfs (rw,xattr,noacl)
rpool/ROOT/delphix.LqquSlN/data on /var/delphix type zfs (rw,xattr,noacl)
```

prakashsurya:
Thanks @pzakha. In that case, when folks have time, I think this is ready for review.

PR #197, opened 2019-01-16
Add README.md to "upgrade-scripts" directory

prakashsurya:


prakashsurya:
cc: @grwilson @pzakha @dtzeng @pandar00

The contents of this change are fairly minimal, but I think we should try and more documentation (e.g. add an FAQ, add docs/diagrams w.r.t. the upgrade architecture, describe how to effectively test changes to these scripts, etc.) so this is a first step in that direction.

PR #198, opened 2019-01-16
Use zfs-mount.service to mount datasets in container

prakashsurya:
The "zfs-mount" service runs and works within the upgrade container, so
there's no need to explicitly bind mount "/var/dlpx-update" using the
container configuration file. Since the dataset backing that directory
is set up to automatically mount on boot, it'll automatically get
mounted within the container when the container boots up (as a result of
the "zfs-mount" service).


prakashsurya:
cc: @grwilson

PR #201, opened 2019-01-18
Add 84-clear-apt-cache.binary live-build hook

prakashsurya:


prakashsurya:
`git-ab-pre-push -v internal-dev` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/387/)

prakashsurya:
cc: @pzakha

prakashsurya:
From that link:

> The clean command clears out the local repository of downloaded package files. It removes everything except the partials folder and lock file from /var/cache/apt/archives/.

It might not practically matter, but we probably want to delete the partial directory too.

Looking at [this comment](https://github.com/rocker-org/rocker/issues/35#issuecomment-58944297), we could probably run this command instead:
```
rm -f /var/cache/apt/archives/*.deb /var/cache/apt/archives/partial/*.deb /var/cache/apt/*.bin
```

> Are you sure it's safe to delete the whole of /var/lib/apt/ vs just /var/lib/apt/lists?

Without explicit documentation from APT, I'm not sure exactly what's safe and what isn't. I chose to try this way first, and was going to test with the image that `git-ab-pre-push` generates (still waiting for that job to finish). I also wanted to just get _something_ open in a PR, so we could open this discussion and iterate on the correct solution.

If you'd rather be a little more conservative, I'd be OK doing this instead:
```
rm -f /var/cache/apt/archives/*.deb
rm -f /var/cache/apt/archives/partial/*.deb
rm -f /var/cache/apt/*.bin
rm -rf /var/lib/apt/lists/*
```

What do you think?

pzakha:
I agree with being more conservative.
Do we know what is the partial folder for?
Running `apt-get clean` on a dlpx-trunk system deletes everything in `/var/cache/apt` except those directories and the lock file:
```
./archives
./archives/partial
./archives/lock
```

I'm fine with deleting `rm -rf /var/lib/apt/lists/*`

prakashsurya:
> Do we know what is the partial folder for?

I'd have to go dig up the docs to explain it fully, but I think it has something to do with partially downloaded packages. I'm OK leave the contents of that directory intact for now, we can look into it further if we find out it's a problem.

prakashsurya:
FWIW, the prior version appears to work; I was able to run `apt-get update` and `apt-get install` without issues. I'm fine going with this second iteration, though.

PR #209, opened 2019-01-25
Move upgrade README.md into top-level upgrade directory

prakashsurya:
Now that we have a top-level "upgrade" directory, it makes sense for the
upgrade related README.md file to live in that directory, rather than
one level down in the "upgrade-scripts" directory.


prakashsurya:
@jgallag88 with the new directory hierarchy, does this move make sense to you?

PR #210, opened 2019-01-25
Add script to choose which rootfs to boot next

prakashsurya:
This changes create a new upgrade script, rootfs-container, which can be
used to configure which "rootfs container" to be used as the appliance's
boot filesystem; i.e. the filesystem that will be used as the rootfs the
next time the system boots.

This logic had previously been contained in the "upgrade-container"
script's "convert_to_bootfs" function, so this change simply moves this
logic into a new script so it can be invoked seperately.

As an example, take the following system:

    $ zfs list /
    rpool/ROOT/delphix.a0Jj3pP/root  6.77G  32.2G     6.77G  /

    $ sudo zfs list -r rpool/ROOT -d 1
    NAME                         USED  AVAIL     REFER  MOUNTPOINT
    rpool/ROOT                  31.6G  32.2G       65K  none
    rpool/ROOT/delphix.a0Jj3pP  6.78G  32.2G       64K  none
    rpool/ROOT/delphix.dREvZi4  18.0G  32.2G       64K  none
    rpool/ROOT/delphix.g1RRkAS  6.84G  32.2G       65K  none

We can use this new script to switch the boot filesystem:

    $ sudo /var/dlpx-update/latest/rootfs-container set-bootfs delphix.dREvZi4
    Installing for i386-pc platform.
    Installation finished. No error reported.
    Generating grub configuration file ...
    Found linux image: /boot/vmlinuz-4.15.0-43-generic
    Found initrd image: /boot/initrd.img-4.15.0-43-generic
    done

    $ sudo reboot

    $ zfs list /
    rpool/ROOT/delphix.dREvZi4/root  8.21G  31.9G     7.61G  /

### DIFF
--- a/upgrade/upgrade-scripts/rootfs-container
+++ b/upgrade/upgrade-scripts/rootfs-container
@@ -19,6 +19,70 @@
 
 CONTAINER=
 
+function get_dataset_snapshots() {
+	zfs list -rt snapshot -Hpo name "$1"
+}
+
+function get_snapshot_clones() {
+	zfs get clones -Hpo value "$1"
+}
+
+function delete() {
+	zfs list "rpool/ROOT/$CONTAINER/root" &>/dev/null ||
+		die "rootfs container '$CONTAINER' does not exist"
+
+	MOUNTED=$(zfs get mounted -Hpo value "rpool/ROOT/$CONTAINER/root")
+	[[ "$MOUNTED" == "no" ]] ||
+		die "cannot delete mounted rootfs container: '$CONTAINER'"
+
+	#
+	# In the common case, the "root" dataset should never have
+	# clones of it. With that said, it is possible for clones to
+	# exist in a couple edge cases:
+	#
+	# 1. When "in-place" upgrade occurs, it'll clone the root
+	#    dataset during the upgrade verification phase; i.e. when
+	#    creating the in-place upgrade container. Thus, if there
+	#    happens to be a concurrent in-place upgrade running, a
+	#    clone of the "root" dataset could exist.
+	#
+	# 2. If an "in-place" upgrade container has been manually
+	#    created, i.e. explicitly via the "upgrade-container"
+	#    script, then it's possible for a clone of the "root"
+	#    dataset to exist.
+	#
+	# Thus, to handle these edge cases, we abort the delete
+	# operation if a clone of "root" exists.
+	#
+	for snap in $(get_dataset_snapshots "rpool/ROOT/$CONTAINER/root"); do
+		for clone in $(get_snapshot_clones "$snap"); do
+			die \
+				"cannot delete rootfs container: '$CONTAINER'," \
+				"'root' dataset clone exists: '$clone'"
+		done
+	done
+
+	#
+	# The "data" and "home" datasets of a rootfs container may have
+	# been cloned as part of an upgrade. Thus, in order to delete
+	# this specific rootfs container, we need to promote any clones
+	# that exist. Otherwise we won't be able to destroy the snapshots
+	# for the "data" and "home" datasets, and thus can't destroy the
+	# datasets themselves.
+	#
+	for snap in \
+		$(get_dataset_snapshots "rpool/ROOT/$CONTAINER/data") \
+		$(get_dataset_snapshots "rpool/ROOT/$CONTAINER/home"); do
+		for clone in $(get_snapshot_clones "$snap"); do
+			zfs promote "$clone" ||
+				die "'zfs promote $clone' failed"
+		done
+	done
+
+	zfs destroy -r "rpool/ROOT/$CONTAINER" ||
+		die "'zfs destroy -r rpool/ROOT/$CONTAINER' failed"
+}
+
 function get_bootloader_devices() {
 	#
 	# When installing/updating the bootloader during upgrade, we
@@ -184,6 +248,7 @@ function usage() {
 	PREFIX_NCHARS=$(echo -n "$PREFIX_STRING" | wc -c)
 	PREFIX_SPACES=$(printf "%.s " $(seq "$PREFIX_NCHARS"))
 
+	echo "$PREFIX_SPACES delete <container>"
 	echo "$PREFIX_SPACES set-bootfs <container>"
 
 	exit 2
@@ -192,6 +257,13 @@ function usage() {
 [[ "$EUID" -ne 0 ]] && die "must be run as root"
 
 case "$1" in
+delete)
+	[[ $# -lt 2 ]] && usage "too few arguments specified"
+	[[ $# -gt 2 ]] && usage "too many arguments specified"
+
+	CONTAINER="$2"
+	delete
+	;;
 set-bootfs)
 	[[ $# -lt 2 ]] && usage "too few arguments specified"
 	[[ $# -gt 2 ]] && usage "too many arguments specified"
--- a/upgrade/upgrade-scripts/rootfs-container
+++ b/upgrade/upgrade-scripts/rootfs-container
@@ -1,0 +1,218 @@
+#!/bin/bash
+#
+# Copyright 2019 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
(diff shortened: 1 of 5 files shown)


upgrade/upgrade-scripts/upgrade-container
```diff
@@ -364,81 +364,7 @@ function run() {
...
-	done
-
-	chroot "/var/lib/machines/$CONTAINER" \
-		grub-mkconfig -o /mnt/boot/grub/grub.cfg ||
-		die "'grub-mkconfig' failed in '$CONTAINER'"
-
-	convert_to_bootfs_cleanup
-	trap - EXIT
-
+function convert_to_rootfs() {
```
pzakha: It's not very obvious anymore what this command does.
It would be helpful to document what is a `rootfs` container vs an `upgrade` container.

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -364,81 +364,7 @@ function run() {
...
 	zfs umount "rpool/ROOT/$CONTAINER/root" ||
 		die "'zfs umount rpool/ROOT/$CONTAINER/root' failed"
 	zfs set mountpoint=/ "rpool/ROOT/$CONTAINER/root" ||
```
pzakha: Should we also explicitly set canmount=noauto (even if it's probably already the case)?

upgrade/upgrade-scripts/rootfs-container
```diff
@@ -0,0 +1,263 @@
...
+
+	#
+	# The "data" and "home" datasets of a rootfs container may have
+	# been cloned as part of an upgrade. Thus, in order to delete
+	# this specific rootfs container, we need to promote any clones
+	# that exist. Otherwise we won't be able to destroy the snapshots
+	# for the "data" and "home" datasets, and thus can't destroy the
+	# datasets themselves.
+	#
+	for snap in \
```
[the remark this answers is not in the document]
prakashsurya: How would those clones get created?

While I tend to be defensive, and try to check for things like that.. I opted not to do that, because there isn't a way for those clones to be created. So, it's hard for me to justify why we'd check for clones, vs. anything else that we don't expect to exist and don't explicitly create.

But if you think it'd be more clear to check for clones, I can do that.

upgrade/upgrade-scripts/rootfs-container
```diff
@@ -0,0 +1,263 @@
...
+	umount "/mnt" || warn "'umount' of '/mnt' failed"
+}
+
+#
+# This function assumes the rootfs container specified is currently
+# mounted; see the "set_bootfs_not_mounted" function for doing this same
+# operation, but for a rootfs container that's not mounted.
+#
+function set_bootfs_mounted() {
+	trap set_bootfs_not_mounted_cleanup EXIT
```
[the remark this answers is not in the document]
prakashsurya: :+1:

upgrade/upgrade-scripts/rootfs-container
```diff
@@ -0,0 +1,263 @@
...
+	# When installing/updating the bootloader during upgrade, we
+	# need to determine which devices are being used as bootloader
+	# devices. We determine this by listing the devices used by the
+	# rpool. Additionally, we have to filter out devices that could
+	# be attached to the rpool, but would never be used for the
+	# bootloader. Finally, we need to strip off any parition
+	# information, since we want to install the bootloader directly
+	# to the device, rather than to a partition of the device.
+	#
+	zpool list -vH rpool |
```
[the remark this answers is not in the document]
prakashsurya: I'm not changing this code, just changing the location of it.

I'd rather just be consistent with the code we're already using, than make an unrelated change in the context of this PR. I'm not against changing this function, I think that's just best left for a separate commit.

Can you open a bug to track this?

upgrade/upgrade-scripts/upgrade-container
prakashsurya: I agree; that's on my list. Do you have any suggestions on where that documentation should go?

upgrade/upgrade-scripts/upgrade-container
prakashsurya: Since we don't explicitly set the `canmount` property for the other "rootfs datasets" (i.e. `data` or `home`), I didn't think it was appropriate to do that for `root`.

Do you think we should duplicate the setting that we already make in `create_upgrade_container`, and set `mountpoint` and `canmount` for `root`, `data`, and `home`? If we do it for one, I think we should do it for all.

Alternatively, I can add a comment stating that we assume these properties are correct, and they get set when the datasets are first created.

upgrade/upgrade-scripts/rootfs-container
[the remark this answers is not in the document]
pzakha: If that container was an active `rootfs` at some point, then it could potentially have clones. While this will probably not happen in a normal use case, I feel like this check shouldn't cost anything but would safeguard against some potential issues, if we were to change the way some of the container creation works.

upgrade/upgrade-scripts/upgrade-container
pzakha: No particular suggestions, but somewhere in `upgrade/upgrade-scripts/` would seem like the logical place to me.

upgrade/upgrade-scripts/upgrade-container
pzakha: I think the comment would be fine.

upgrade/upgrade-scripts/rootfs-container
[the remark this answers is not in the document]
prakashsurya: > If that container was an active rootfs at some point, then it could potentially have clones.

i.e. if an "in-place" upgrade is concurrently running? the clone will only be created for in-place upgrade verification, right? I just want to make sure I'm understanding.

I guess it's also possible for an in-place upgrade container to have been manually created and currently running, which could allow for a clone exist, I think.

I'll add this check.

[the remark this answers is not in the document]
prakashsurya: I was able to get into this scenario by manually creating an upgrade container. Without a check, the failure message was bad (as might be expected):
```
$ sudo /var/dlpx-update/latest/rootfs-container delete delphix.JNHeZad
cannot destroy 'rpool/ROOT/delphix.JNHeZad': filesystem has dependent clones
use '-R' to destroy the following datasets:
rpool/ROOT/delphix.uE0noy5/root
rootfs-container: 'zfs destroy -r rpool/ROOT/delphix.JNHeZad' failed
```

With the check:
```
$ sudo /var/dlpx-update/latest/rootfs-container delete delphix.JNHeZad
rootfs-container: cannot delete rootfs container: 'delphix.JNHeZad', 'root' dataset clone exists: 'rpool/ROOT/delphix.uE0noy5/root'
```

Thanks for the feedback, this is much better.

upgrade/upgrade-scripts/upgrade
```diff
@@ -178,10 +190,41 @@ function upgrade_not_in_place() {
...
 
+	#
+	# In order for the "rollback" back command to work, we need to
+	# know which rootfs container to rollback to. Thus, we embed
+	# this information into the rootfs container using a ZFS
+	# property; rollback will read this information, to determine
+	# which rootfs container to rollback to.
+	#
+	zfs set \
+		"$ROLLBACK_PROPERTY=$(get_mounted_rootfs_container)" \
```
pzakha: We should probably call `$(get_mounted_rootfs_container)` on a separate line to prevent bash from swallowing errors.

upgrade/upgrade-scripts/upgrade
```diff
@@ -208,6 +251,10 @@ not-in-place)
 	verify_upgrade_is_allowed
 	upgrade_not_in_place "$@"
 	;;
+rollback)
+	shift 1
+	rollback "$@"
```
pzakha: Here and elsewhere, we aren't expecting any arguments, so we shouldn't be passing `"$@"` but rather should fail if any extra arguments are passed.

prakashsurya: Yea, we could fail. I didn't do that because the other commands didn't handle it that way. I'll open a bug to follow up on this.

prakashsurya: I've opened: https://github.com/delphix/appliance-build/issues/220

upgrade/upgrade-scripts/upgrade
prakashsurya: give my latest revision and look, and let me know what you think.

upgrade/upgrade-scripts/upgrade-container
prakashsurya: I've added a new commit that adds a new FAQ.md file; when you get a chance, let me know if you think it sufficiently documents what a rootfs and upgrade container is.

prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/448/)

PR #212, opened 2019-01-26
Only download .debs.tar.gz when building upgrade image

jgallag88:
When we build an upgrade image for a variant, we currently download all
of the artfacts produced by live build for that variant. We really only
need the .deb.tar.gz files though. By omitting the other files, we can
save time and disk space.


jgallag88:
**Testing:**

appliance-build: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build/job/master/job/pre-push/344/

Downloaded an upgrade image produced from the appliance-build job, checked the set of debian packages it contained, and confirmed that it was exactly the same as the set of packages in an upgrade image from a recent post-push job.

prakashsurya:
bors delegate+

PR #213, opened 2019-01-27
LX-1682 De-couple delphix-entire and delphix-virtualization version

dtzeng:
See app-gate review for full architecture explanation http://reviews.delphix.com/r/46631/
Jenkins changes here http://reviews.delphix.com/r/46521/

### DIFF
--- a/live-build/config/hooks/82-upgrade-repository.binary
+++ b/live-build/config/hooks/82-upgrade-repository.binary
@@ -91,25 +91,16 @@ cp config/hooks/template.ctl delphix-entire.ctl
 DEPENDS=$(chroot binary \
 	dpkg-query -Wf '${Package} (=${Version}), ' | sed 's/, $//')
 
-if chroot binary dpkg -l delphix-virtualization &>/dev/null; then
+if [[ -n "$DELPHIX_APPLIANCE_VERSION" ]]; then
 	#
-	# If the delphix-virtualization package is installed in this
-	# variant, we want the delphix-entire's package version to match
-	# the version of the delphix-virtualization package. The upgrade
-	# logic contained in the virtualization application requires
-	# these two package verions to match.
+	# If the build specified a version for the delphix-entire package,
+	# then use it.
 	#
-	# shellcheck disable=SC2016
-	VERSION=$(chroot binary dpkg-query \
-		-Wf '${Version}' delphix-virtualization)
+	VERSION=$DELPHIX_APPLIANCE_VERSION
 else
 	#
-	# If the delphix-virtualization package is not installed in this
-	# variant, then we just use a default value for the version,
-	# based on when he build occurs. In this case, we only need the
-	# version to be larger than any previous build, to enable
-	# upgrades from one build to another (e.g. when manually using
-	# the "execute" upgrade script).
+	# Otherwise, just use a default value for the version, based on when
+	# the build occurs.
 	#
 	VERSION=$(date '+%Y.%m.%d.%H')
 fi
--- a/scripts/docker-run.sh
+++ b/scripts/docker-run.sh
@@ -58,6 +58,7 @@ $DOCKER_RUN --rm \
 	--env DELPHIX_SIGNATURE_URL \
 	--env DELPHIX_SIGNATURE_TOKEN \
 	--env DELPHIX_PLATFORMS \
+	--env DELPHIX_APPLIANCE_VERSION \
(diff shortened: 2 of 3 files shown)


scripts/build-upgrade-image.sh
```diff
@@ -67,9 +67,21 @@ aptly repo create -distribution=bionic -component=delphix upgrade-repository
...
+#
+ENTIRE_VERSION=$(dpkg -f "$(find debs/ -name 'delphix-entire-*' | head -n 1)" version)
+virt_pkg=$(find debs/ -name 'delphix-virtualization-*' | head -n 1)
+if [[ "$virt_pkg" ]]; then
+	VERSION=$(dpkg -f "$virt_pkg" version)
+else
+	VERSION=$ENTIRE_VERSION
+fi
+cp version.info.template version.info
+sed -i "s/@@ENTIRE_VERSION@@/$ENTIRE_VERSION/" version.info
```
prakashsurya: I think we should discuss things.. it strikes me as awkward how we now have `VERSION` which maps to the virtualization package's version.. and then `ENTIRE_VERSION` which maps to the entire package.

I don't really like how we're leaking more knowledge of the virtualization package, into the upgrade process/image. At this point, I feel like we should be more explicit and rename `VERSION` to `VIRTUALIZATION_VERSION`, and perhaps even include other package versions like masking, zfs, etc.. but, I'd rather not have any of these package versions at all, if we can avoid it.

scripts/docker-run.sh
```diff
@@ -58,6 +58,7 @@ $DOCKER_RUN --rm \
 	--env DELPHIX_SIGNATURE_URL \
 	--env DELPHIX_SIGNATURE_TOKEN \
 	--env DELPHIX_PLATFORMS \
+	--env DELPHIX_ENTIRE_VERSION \
```
prakashsurya: I think it might be better to rename this to something more vague, like `DELPHIX_APPLIANCE_VERSION`.. this way, if we ever change the implementation of `delphix-entire`, we likely won't have to change the meaning/name of the variable (i.e. the interface of the build won't have to change).

upgrade/version.info.template
```diff
@@ -47,7 +55,7 @@ VERSION=@@VERSION@@
 MINIMUM_VERSION=5.3.0.0
 
 #
-# The minimum version from which a reboot after upgrade is optional.
+# The minimum product version from which a reboot after upgrade is optional.
```
prakashsurya: I'm a little skeptical using "product" and "virtualization" interchangeably.. since, to me, the "product" feels more like the culmination of all of our delphix packages (masking, virtualization, zfs, etc).. IMO, nomenclature is important to get right, and the use of "product" here (and elsewhere) is a bit sketchy to me.

If we could come up with a definition for "product" and document that somewhere, then I'd be more OK with it.

upgrade/upgrade-scripts/common.sh
```diff
@@ -41,10 +41,25 @@ function get_platform() {
...
+function get_installed_entire_version() {
 	dpkg-query -Wf '${Version}' "delphix-entire-$(get_platform)"
 }
 
+#
+# The virtualization package may not always exist (i.e. internal minimal
+# builds). In that case, we fall back to the delphix-entire version as the
+# product version.
+#
+function get_installed_product_version() {
```
prakashsurya: If "the product" isn't installed, would it be better to return an empty string from this function? This way, it's more clear that "the product" isn't installed, rather than trying to hide that fact from the user..?

To me, if "the product" is the virtualization package, and that package isn't installed, it seems wrong to lie and return the version of the delphix-entire package.. even if this behavior makes sense based on how this function is used, from an API/interface standpoint, this behavior strikes me as awkward.

live-build/config/hooks/82-upgrade-repository.binary
```diff
@@ -91,25 +91,16 @@ cp config/hooks/template.ctl delphix-entire.ctl
 DEPENDS=$(chroot binary \
 	dpkg-query -Wf '${Package} (=${Version}), ' | sed 's/, $//')
 
-if chroot binary dpkg -l delphix-virtualization &>/dev/null; then
+if [[ "$DELPHIX_APPLIANCE_VERSION" ]]; then
```
prakashsurya: can you add `-n` here? i.e. `if [[ -n "..." ]]; then`

PR #217, opened 2019-01-29
Only copy OVA when building for ESX

jgallag88:
When we build an OVA for ESX, we first build a VMDK, and then convert
that into an OVA. We end up moving both artifacts to the artifacts/
directory, and thus we end up uploading both, but we really only need
the OVA.

### DIFF
--- a/scripts/run-live-build.sh
+++ b/scripts/run-live-build.sh
@@ -120,6 +120,18 @@ if [[ ! -f binary/SHA256SUMS ]]; then
 	exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


jgallag88:
**Testing:**

Did an appliance build that included every platform, and checked that job uploaded exactly one vm artifact for each platform.

ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/411/console

PR #223, opened 2019-01-30
Add new commit status to bors configuration

prakashsurya:


prakashsurya:
cc: @pzakha

bors try

prakashsurya:
bors try

pzakha:
This should only be integrated after our automation posts statuses properly.
It would be worth it to add some context in the description.

prakashsurya:
We’re posting the statuses. You can see them in the UI. The “bors try” would have failed otherwise.

PR #224, opened 2019-01-31
Stop creating VHD

jgallag88:
When we create VM artifacts, we create both a VHD and a VHDX, but only
upload the VHDX. The VHD is completely unused and the script that
creates it can be removed.


jgallag88:
ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/458/console

jgallag88:
cc @grodr

prakashsurya:
bors delegate+

PR #230, opened 2019-02-19
LX-1767 Remove upstream td-agent from appliance-build

prashks:
Now that we'll be building custom _td-agent_  via linux-pkg framework (LX-1645), this change removes the upstream _td_agent_ that was added earlier via LX-1238.


prashks:
`git-ab-pre-push` : http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/515/console

PR #231, opened 2019-02-20
LX-1747 /var/delphix dataset should be migrated to Linux

pzakha:
`/var/delphix` contains some Delphix configuration and useful debug logs, so we want to carry that directory over into Linux. Note that we do that in `dx_execute` since that's when the stack is quiesced.

There's also a side fix to the code that edits `/etc/fstab` as it didn't seem to work consistently.

## TESTING

 - ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/503/
 - verify that  `/var/delphix` was carried over after rebooting into Linux:
```
delphix@localhost:/var/delphix$ df .
Filesystem                      1K-blocks  Used Available Use% Mounted on
rpool/ROOT/delphix.WhJtaCV/data 116131072  4864 116126208   1% /var/delphix
delphix@localhost:/var/delphix$ zfs get origin rpool/ROOT/delphix.WhJtaCV/data
NAME                             PROPERTY  VALUE                                                                   SOURCE
rpool/ROOT/delphix.WhJtaCV/data  origin    rpool/versions/5.3.2019.02.12/dlpx/5.3.3.0/running/var@delphix.WhJtaCV  -
```

### DIFF
--- a/live-build/misc/migration-scripts/dx_apply
+++ b/live-build/misc/migration-scripts/dx_apply
@@ -154,12 +154,18 @@ mount -F zfs "$RPOOL/ROOT/$FSNAME/data" "$TMP_ROOT/var/delphix"
 (
 	cd "$TMP_ROOT"
 	cpio -imu
-	cat <<-EOF >"/etc/fstab"
-		$RPOOL/ROOT/$FSNAME/home /export/home zfs defaults 0 0
-		$RPOOL/ROOT/$FSNAME/data /var/delphix zfs defaults 0 0
-	EOF
 ) <"$ARCHIVE_DIR/os-root.cpio" 2>&1 ||
 	die "failed to unpack os-root.cpio"
+
+#
+# /export/home and /var/delphix datasets are not mounted automatically.
+# On Illumos this kind of mounting logic was performed by dxinit and svc-boot.
+# On Linux /etc/fstab is used instead.
+#
+cat <<-EOF >"$TMP_ROOT/etc/fstab" || die "Failed to setup /etc/fstab"
+	$RPOOL/ROOT/$FSNAME/home /export/home zfs defaults 0 0
+	$RPOOL/ROOT/$FSNAME/data /var/delphix zfs defaults 0 0
+EOF
 
 #
 # WIP: Add here any operations that generate files from illumos,
@@ -269,8 +275,25 @@ MAIN_MENU_FICL=(
 } >>/boot/menu.rc.local ||
 	die "failed to update /boot/menu.rc.local"
 
-umount "$TMP_ROOT/var/delphix" ||
-	die "couldn't unmount linux directory $TMP_ROOT/var/delphix"
+#
+# Note that the new /var/delphix should not contain any useful data as it will
+# be replaced by a clone of the current /var/delphix dataset in dx_execute.
+# We do not clone the current /var/delphix right away as the clone will be
+# outdated by the time dx_execute is called. Before deleting the dataset we
+# first verify that it doesn't contain anything useful.
+#
+(
+	cd "$TMP_ROOT/var/delphix" || die "failed to cd into $TMP_ROOT/var/delphix"
+	# We only expect to find:
+	#  ./dropbox
+	[[ $(find . -mindepth 1 | wc -l) -eq 1 ]] ||
+		die "linux dataset for /var/delphix contains unexpected files"
+	[[ -d dropbox ]] ||
+		die "linux dataset for /var/delphix should contain directory dropbox"
+) || die "verification of /var/delphix failed"
+
+zfs destroy "$RPOOL/ROOT/$FSNAME/data" ||
+	die "failed to destroy linux dataset $RPOOL/ROOT/$FSNAME/data"
 umount "$TMP_ROOT/export/home" ||
 	die "couldn't unmount linux dataset $TMP_ROOT/export/home"
 
--- a/live-build/misc/migration-scripts/dx_execute
+++ b/live-build/misc/migration-scripts/dx_execute
@@ -23,7 +23,7 @@
 set -o pipefail
 
 function die() {
-	report "$(basename "$0"): $*" >&2
+	echo "$(basename "$0"): $*" >&2
 	exit 1
 }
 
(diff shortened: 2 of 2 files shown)


pzakha:
cc: @sdimitro @prakashsurya

live-build/misc/migration-scripts/dx_execute
```diff
@@ -97,6 +97,21 @@ LX_RDS="$LX_RDS_PARENT/root"
 	die "the expected Linux RDS ($LX_RDS) was either not found or has" \
 		"been specified more than once in the bootloader's menu file"
 
+#
+# Replace the empty /var/delphix dataset from the migration image by a clone
+# of the current /var/delphix dataset.
+#
+LX_VAR_DLPX="$LX_RDS_PARENT/data"
+LX_CONTAINER="${LX_RDS_PARENT##*/}"
+CUR_VAR_DLPX=$(mount | awk '/^\/var\/delphix /{ print $3 }')
```
prakashsurya: would it be better to use `zfs list` here? e.g.
```
$ zfs list -Ho name /var/delphix
rpool/versions/2017.6.2017.06.28/dlpx/2017.6.28.18/running/var
```

live-build/misc/migration-scripts/dx_execute
```diff
@@ -97,6 +97,21 @@ LX_RDS="$LX_RDS_PARENT/root"
...
+[[ -n $CUR_VAR_DLPX ]] || die "could not determine current /var/delphix dataset"
+zfs list "$CUR_VAR_DLPX" >/dev/null 2>&1 ||
+	die "'$CUR_VAR_DLPX' is not a valid zfs dataset"
+zfs destroy "$LX_VAR_DLPX" || die "failed to destroy '$LX_VAR_DLPX'"
+zfs snapshot "${CUR_VAR_DLPX}@${LX_CONTAINER}"
+zfs clone -o compression=on -o mountpoint=legacy \
```
prakashsurya: Will `rootfs-container delete` still work correctly on the initial dataset used for the LX migration? Does it currently work without this change?

Right now we can do a not-in-place upgrade, then use `rootfs-container delete` to remove the old rootfs container/datasets. So, I'm wondering if we'll be able to do a migration, then a not-in-place upgrade, and still use the `rootfs-container delete` script to delete that dataset that was used for the initial migration?

As a tangential concern, "leaking" things from illumos to linux worries me a little bit since it can make it more difficult to mimic customer systems in-house. E.g. in this particular example, I could envision us testing the linux equivalent of `dx_delete` in-house, without ever taking migration into account. This could give us false positives when it works with migration out of the picture, but it breaks when the customer previously did a migration.

I don't mean to imply this is wrong, I just think we need to be mindful about all the possible implications that our migration implementation might have on other parts of our software.

live-build/misc/migration-scripts/dx_apply
```diff
@@ -139,6 +139,10 @@ zfs create \
 	"$RPOOL/ROOT/$FSNAME/home" ||
 	die "failed to create linux dataset $RPOOL/ROOT/$FSNAME/home"
 
+#
+# Note that this dataset should not contain any useful data as it will be
```
prakashsurya: Are you sure the order of operations here is correct? i.e. to extract the contents of the archive first, then effectively delete anything the archive might have put in /var/delphix, and replace it with a clone..

For linux upgrade, we create the clone first, then update/install the new packages. So, if the package placed things in `/var/delphix`, they'd remain there. Whereas, I think in this implementation, if the package placed things in `/var/delphix`, they'd be removed.

Would it be better to first create the clone, then extract the archive? I know you state that you need to create the clone in `dx_execute`, but I'm wondering if that's actually "safe" to do.. Can we at least add some verification to ensure the linux `/var/delphix` dataset is empty before it's destroyed and replaced?

pzakha: The dataset is not actually fully empty, it contains the empty "dropbox" directory. I can add some verifications to make sure that there isn't anything else in there though. The dataset definitely has to be cloned in `dx_execute` as configuration in `/var/delphix/server/etc/` and just logs in general would be out of date if we do this in `dx_apply`.

live-build/misc/migration-scripts/dx_execute
pzakha: We'll definitely need to do some dataset cleanup after rebooting into Linux like promoting the clone and deleting the old Illumos datasets.

I do understand your concerns about "leaking" config from the migration process, however this is bound to happen in all the aspects of the product (i.e. datasets on domain0 that were created earlier or some OS configurations). We should strive to reduce those differences where we can, and just be mindful of doing the proper testing where we cannot. We do have plans of testing Linux upgrade after migration.

live-build/misc/migration-scripts/dx_execute
pzakha: Cool, I didn't realize zfs list could take mountpoints as arguments.

live-build/misc/migration-scripts/dx_apply
```diff
@@ -139,6 +139,10 @@ zfs create \
...
+# replaced by a clone of the current /var/delphix dataset in dx_execute.
```
sdimitro: Should we also explain "why?" here?
(e.g. have not quiesced VDB yet, there can be configuration changes, etc..)

live-build/misc/migration-scripts/dx_apply
pzakha: @prakashsurya I've done some refactoring so that the dataset is checked and destroyed in `dx_apply`. Could you take another look?

live-build/misc/migration-scripts/dx_apply
```diff
@@ -269,8 +275,25 @@ MAIN_MENU_FICL=(
...
+# We do not clone the current /var/delphix right away as the clone will be
+# outdated by the time dx_execute is called. Before deleting the dataset we
+# first verify that it doesn't contain anything useful.
+#
+(
+	cd "$TMP_ROOT/var/delphix" || die "failed to cd into $TMP_ROOT/var/delphix"
+	# We only expect to find:
+	#  ./dropbox
+	[[ $(find . -mindepth 1 | wc -l) -eq 1 ]] ||
+		die "linux dataset for /var/delphix contains unexpected files"
```
prakashsurya: these `die` statements will only exit the subshell, and not the parent shell, right? so, I think we should do something like:
```
(
    ...
) || die "verification of /var/delphix failed"
```

pzakha: Yeah, you are perfectly right... Nice catch, almost got tricked by a subshell issue again.

PR #232, opened 2019-02-22
Fix leading zeroes in version generation

dtzeng:


dtzeng:
@prakashsurya @pandar00 @jgallag88

PR #234, opened 2019-02-25
Add additional cloud-init datasources for internal variants

jgallag88:
On the internal variants, we have cloud-init configured to read EC2
userdata, which we use to set up SSH keys on newly-provisioned VMs in
EC2. We should also allow cloud-init to read userdata from GCP, so that
we can set up SSH keys on internal Delphix engines there too. The
devops-gate has a workaround in place such that reading userdata from
Azure isn't strictly necessary, but we might as well add it while we are
at it so that that workaround can be removed.


jgallag88:
ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/594/console

PR #235, opened 2019-02-26
Add additional cloud-init datasources for internal variants (revert)

jgallag88:
The original change broke cloud-init on the ESX. Reverting until we figure out the correct way to make this change.


jgallag88:
bors retry

PR #237, opened 2019-02-26
Run "apt-get autoremove" after an upgrade

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/598/)

prakashsurya:
I've run both a "not-in-place" and "in-place" upgrade and verified this worked. Without this change, the "krb5-kdc" package would remain installed after a "in-place" upgrade, after this change that package is removed (both the package, and it's configuration files).

See [here](http://reviews.delphix.com/r/46999/) for details about why this package originally was installed, and why it should no longer be installed after the upgrade.

upgrade/upgrade-scripts/execute
```diff
@@ -132,6 +132,16 @@ source_version_information
...
 
+#
+# After we've successfully installed the new delphix-entire package, we
+# need to remove all packages that are no longer required. For example,
+# if the old version depended on a package, and the new version no
+# longer depends on that package, the upgrade performed above will not
+# remove this package. Thus, we need to run the "autoremove" to remove
+# these packages that are no longer required.
+#
+apt_get autoremove --purge -y || die "autoremove after upgrade failed"
```
pzakha: Could that result in removing the current kernel package?
From the man pages:
> autoremove is used to remove packages that were automatically installed to satisfy dependencies for other packages and are now no longer needed.

Currently the kernel package is pulled as a dependency of `delphix-platform-<PLATFORM>`, so we do not explicitly install it.

prakashsurya: No. It won't remove the currently installed kernel package, and it likely won't remove the "old" kernel package either. The important part of that line, is determining what "no longer needed" means.

The only time a package is removed, is when nothing depends on it. If `delphix-entire` depends on `delphix-platform`, and `delphix-platform` depends on `delphix-kernel`, and `delphix-kernel` depends on the kernel package, the kernel package will remain because the dependency on it still remains.

There's some docs about old kernels [here](https://help.ubuntu.com/community/RemoveOldKernels), if you're worried about removing kernels that the currently installed `delphix-entire` does not depend on...
> Note: This way will not remove all automatically installed old kernel providing packages as fallback versions are kept; the list of kept kernels is maintained and automatically updated in the file /etc/apt/apt.conf.d/01autoremove-kernels as a list of matching regular expressions.

pzakha: I've tried running `apt-mark showauto` on a freshly cloned `dlpx-trunk` VM and it listed a bunch of our kernel packages (`delphix-kernel-4.15.0-1032-aws`, `connstat-module-4.15.0-1032-aws`, `zfs-modules-4.15.0-1032-aws`), so that's a bit different from the stock Ubuntu case shown in https://help.ubuntu.com/community/RemoveOldKernels.

That said, it's possible that since `delphix-platform-aws` depends on virtual package `delphix-kernel-aws`, and after upgrade we will have 2 `delphix-kernel-aws` packages (e.g. `delphix-kernel-old-vers-aws` and `delphix-kernel-new-vers-aws`) that satisfy this virtual package dependency, apt would not remove any of those packages.

I think it would be worth it to test this scenario, although I'm not sure we have useable images systems to test that upgrade from. I'm not going to hold this review on this, but this is definitely something that we should test when we get a chance.

prakashsurya: But it should show those as automatically pulled in.. they _are_ automatically pulled in. We don't explicitly install the kernel packages, we pull them in as dependencies of other packages. I'm not sure I understand the concern..?

Just because they're automatically pulled in, doesn't mean they "no longer needed". That just means they weren't explicitly installed with "apt-get install" .. they were pulled in as a dependency of another package (i.e. delphix-platform in this case).

prakashsurya: > and after upgrade we will have 2 `delphix-kernel-aws` packages

Why would we have two kernel packages installed? I don't think it'll work that way. the new package will replace the older one, IIRC.

pzakha: So the way it works with kernel packages is that each new kernel package has a different name, e.g. `zfs-modules-4.15.0-1032-aws` vs `zfs-modules-4.15.0-1033-aws`. In the example I gave, `delphix-kernel-aws` is not a real package, but a virtual package, and we can have as many of those installed at the same time as we want.

prakashsurya: I see.. so we might want/need to customize `APT::NeverAutoRemove` via a file in `/etc/apt/apt.conf.d` similar to how Ubuntu's kernel packages do it.

I can imagine `apt-get autoremove` might auto-remove the delphix kernel packages, but likely leave the Ubuntu kernel packages (since those are configured in `/etc/apt/apt.conf.d`), which will result in a non-functional system (assuming we need to fallback to the old kernel), since the ZFS kernel packages will be removed.

It's probably best to fix this in the delphix-kernel package, rather than allowing any `apt-get autoremove` to create a partially broken system.

pzakha: > I see.. so we might want/need to customize `APT::NeverAutoRemove` via a file in `/etc/apt/apt.conf.d` similar to how Ubuntu's kernel packages do it.

Right, if we were to do that, then that would guarantee that we never auto-remove the kernels, and might be a good thing to do in the long run. That said, for the time being, we _might_ be okay because of the virtual package dependencies.

prakashsurya: I cloned an old image on azure as a way to test this.. here's what it looks like without this change applied, after running an in-place upgrade:
```
$ dpkg -l | grep delphix-kernel
ii  delphix-kernel-4.15.0-1036-azure              1.0.0-delphix-2019.02.01.07        amd64        Kernel packages consolidation for the Delphix Appliance.
ii  delphix-kernel-4.18.0-1011-azure              1.0.0-delphix-2019.02.26.15        amd64        Kernel packages consolidation for the Delphix Appliance.

$ sudo apt-get autoremove -y --purge
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following packages will be REMOVED:
  krb5-kdc* libegl-mesa0* libegl1* libgbm1* libverto-libevent1* libverto1* libwayland-egl1-mesa* libwayland-server0* libxcb-xfixes0*
0 upgraded, 0 newly installed, 9 to remove and 0 not upgraded.
After this operation, 1,406 kB disk space will be freed.
(Reading database ... 165598 files and directories currently installed.)
Removing krb5-kdc (1.16-2ubuntu0.1) ...
Removing libwayland-egl1-mesa:amd64 (18.2.2-0ubuntu1~18.04.1) ...
Removing libegl1:amd64 (1.0.0-2ubuntu2.2) ...
Removing libegl-mesa0:amd64 (18.2.2-0ubuntu1~18.04.1) ...
Removing libgbm1:amd64 (18.2.2-0ubuntu1~18.04.1) ...
Removing libwayland-server0:amd64 (1.16.0-1ubuntu1.1~18.04.1) ...
Removing libxcb-xfixes0:amd64 (1.13-1) ...
Removing libverto1:amd64 (0.2.4-2.1ubuntu3) ...
Removing libverto-libevent1:amd64 (0.2.4-2.1ubuntu3) ...
Processing triggers for libc-bin (2.27-3ubuntu1) ...
Processing triggers for man-db (2.8.3-2ubuntu0.1) ...
(Reading database ... 165521 files and directories currently installed.)
Purging configuration files for krb5-kdc (1.16-2ubuntu0.1) ...
Processing triggers for systemd (237-3ubuntu10.13) ...

$ dpkg -l | grep delphix-kernel
ii  delphix-kernel-4.15.0-1036-azure              1.0.0-delphix-2019.02.01.07        amd64        Kernel packages consolidation for the Delphix Appliance.
ii  delphix-kernel-4.18.0-1011-azure              1.0.0-delphix-2019.02.26.15        amd64        Kernel packages consolidation for the Delphix Appliance.
```
I'm not quite sure _why_ the old `delphix-kernel` package wasn't removed, but it wasn't; which (I think) is what we want in this scenario.

pzakha: My theory is that apt has no way of knowing that `delphix-kernel-4.15.0-1036-azure` is better than `delphix-kernel-4.18.0-1011-azure` as a provider of the `delphix-kernel-azure` virtual package dependency, so it leaves them both.

PR #239, opened 2019-02-27
Fix for LX-1784 and LX-1785 MDS

pandar00:
These two were regressions caused by LX-1337. My apologies.

LX-1784 Upgrade verify fails due to lack of permission
* New scripts, `verify-jar` and `verify-stack` didn't have an execute permission. 

LX-1785 MDS clone isn't cleaned up after verification
* `zfs list <dataset_name>` takes full dataset path, not just part of it, so the snapshot names should appended by `domain0`.

Tested on failed VM.


pandar00:
@prakashsurya @dtzeng

PR #241, opened 2019-03-03
LX-1654 Use not in place for reboot upgrades through the API

pandar00:
This is an `appliance-build` side of change to allow app-stack to consume not-in-place upgrade.

Note: Commits are cluttered with my person commit notes. I'll clean this up before final push.

High-level changes
* Introduces another level of abstraction called "upgrade" in `upgrade-container` script which upgrades the bits in the container to the latest. Caller needs to supply `in-place` or `not-in-place` argument.
* Changed `verify` script to allow user to supply their container. In such case, container `create`, `start`, and `upgrade` is skipped. Container cleanup is skipped as well.
    * In order for Java stack to consume not-in-place upgrade, it needs to
        * Create container and keep the container name. This implies Java stack needs to call `upgrade-container create` separately. The `verify` script output is expected to only contain progress report.
        * Clean up the container. The cleanup code is dependent on two conditions:
            1. Is verification called by verify API. In this case, we cleanup container afterward regardless of container type.
            2. Is verification part of apply upgrade API. In this case, we need to cleanup container only for `in-place` upgrade and since container is used for applying upgrade for `not-in-place` upgrade.
* `verify` script can still be called without supplying container name. This will do all necessary container setup, run verification and then cleans up the container. 

With these two changes, `upgrade` script is bit simplified as shown in the code.


pandar00:
@dtzeng @prakashsurya 
I'll post app-gate side review shortly.

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -564,17 +603,29 @@ run)
 	shift 2
 	run "$@"
 	;;
-convert-to-rootfs)
-	[[ $# -lt 2 ]] && usage "too few arguments specified"
-	[[ $# -gt 2 ]] && usage "too many arguments specified"
+upgrade)
+	# TODO: Note to reviewer:
+	# We can omit container type argument if we are able to determine
+	# container type from the container, but how?
```
prakashsurya: We can do this if we look at the "origin" property of the root dataset.. e.g.

for an in-place container:
```
$ sudo systemd-run --machine delphix.SuwEKjh --quiet --pty /bin/bash
root@localhost:/# zfs get -Hpo value origin $(zfs list -Hpo name /)
rpool/ROOT/delphix.tdmMJy5/root@delphix.SuwEKjh
```

for a not-in-place container:
```
$ sudo systemd-run --machine delphix.cGR6lq3 --quiet --pty /bin/bash
root@localhost:/# zfs get -Hpo value origin $(zfs list -Hpo name /)
-
```

If we run this outside of the container, we probably need to change the `zfs list` to something like `zfs list -Hpo name /var/lib/machines/$CONTAINER`..

So perhaps something like this:
```
delphix@ps-lx-1524:~$ ROOT=$(sudo zfs list -Hpo name /var/lib/machines/delphix.cGR6lq3)
delphix@ps-lx-1524:~$ ORIGIN=$(sudo zfs get -Hpo value origin $ROOT)
delphix@ps-lx-1524:~$ case "$ORIGIN" in
> -) echo "not-in-place" ;;
> *) echo "in-place" ;;
> esac
not-in-place
```
and then replace the `echo` commands with w/e logic we want to do based on the container type..

upgrade/upgrade-scripts/common.sh
```diff
@@ -81,16 +81,21 @@ function verify_upgrade_is_allowed() {
 			"($MINIMUM_VERSION)"
 }
 
-function verify_upgrade_in_place_is_allowed() {
+function is_reboot_required_upgrade() {
```
prakashsurya: can we use consistent terminology between this, and the "verify" function below..

i.e. either settle on "reboot required" or "in-place allowed"

I'm fine with either one.. it just seems like we have two different terms to mean the same thing, so if we could make both use the same terminology that might make things more clear to readers of this code.

upgrade/upgrade-scripts/upgrade
```diff
@@ -99,25 +99,12 @@ function upgrade_in_place() {
...
+		die "Failed to do in-place upgrade on '$CONTAINER'"
 
 	if [[ "$DLPX_UPGRADE_SKIP_VERIFY" != "true" ]]; then
-		"$IMAGE_PATH/upgrade-container" run "$CONTAINER" \
-			"$IMAGE_PATH/verify-jar" "$@" ||
-			die "'$IMAGE_PATH/verify-jar' '$*' failed in '$CONTAINER'"
-		"$IMAGE_PATH/upgrade-container" run "$CONTAINER" \
-			"$IMAGE_PATH/verify-stack" "$@" ||
-			die "'$IMAGE_PATH/verify-stack' '$*' failed in '$CONTAINER'"
+		"$IMAGE_PATH/verify" -c "$CONTAINER" "$@" ||
```
prakashsurya: why use `verify` vs. `verify-jar` and `verify-stack`?

I ask, because I had originally thought that the `verify` script would eventually be removed, since the container creation code already exists here in the `upgrade` script.. and I was thinking the container creation logic would also get wrapped into the app-gate upgrade logic (and it'd use `verify-jar` and `verify-stack` directly).

I think I might have left a similar review on your app-gate change, so it might be good to discuss this at some point, so we're all on the same page.

upgrade/upgrade-scripts/upgrade-container
pandar00: thanks I'll make changes

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -527,8 +527,23 @@ function do_upgrade_container_not_in_place() {
 		die "failed to migrate configuration for '$CONTAINER'"
 }
 
+function get_type() {
+	local root origin
+
+	root=$(sudo zfs list -Hpo name /var/lib/machines/${CONTAINER})
```
prakashsurya: remove `sudo` here; we already force the caller to be root, so `sudo` is unnecessary.

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -527,8 +527,23 @@ function do_upgrade_container_not_in_place() {
...
+	[[ -n "${root}" ]] || die "Failed to obtain root filesystem for container '${CONTAINER}'"
+	origin=$(sudo zfs get -Hpo value origin ${root})
```
prakashsurya: remove `sudo` here

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -527,8 +527,23 @@ function do_upgrade_container_not_in_place() {
...
+	-)
+		echo "not-in-place" ;;
+	*)
+		echo "in-place" ;;
+	esac
+}
+
 function do_upgrade_container() {
-	local type="$1"
+	local type=${get_type}
```
prakashsurya: IMO, it'd be more simple to just do this:
```
origin=$(sudo zfs get -Hpo value origin ${root})
case "$origin" in
-)
    do_upgrade_container_in_place || ...
    ;;
*)
    do_upgrade_container_not_in_place || ...
    ;;
esac
```

The `get_type` function is only used once, so it'd be a little more clear (to me, at least) if we just combined it with this `do_upgrade_container` function.

prakashsurya: Although, now I see you're exposing the `get_type` function via `upgrade-container get-type`...

if we think it'd be useful to allow callers to get the type of the container like you're doing, then I think what you have makes sense.

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -517,8 +571,8 @@ function usage() {
 	echo "$PREFIX_SPACES stop <container>"
 	echo "$PREFIX_SPACES destroy <container>"
 	echo "$PREFIX_SPACES run <container> <command>"
+	echo "$PREFIX_SPACES upgrade <container> [in-place|not-in-place]>"
```
prakashsurya: need to update this too..

upgrade/upgrade-scripts/upgrade
pandar00: I initially thought of removing `verify` script as well since the verification logic is already defined in `upgrade` script, but I thought it would be important to maintain a way to run all verification logic (including container creation to cleanup) from a CLI since I believe there are other engineers who prefers doing that (although I'm not an avid user of this feature).

One benefit I noticed of having a separate `verify` script is that it simplifies client's code since it figures out which verification it needs to do, although this isn't a significant benefit now.

I'm fine with removing `verify` script

upgrade/upgrade-scripts/verify
```diff
@@ -75,36 +80,70 @@ function cleanup() {
...
+
+while getopts ':c:' c; do
+	case "$c" in
+	c)
+		CONTAINER=$OPTARG
+		CALLER_PROVIDED_CONTAINER=true
+		;;
+	esac
+done
+shift $((OPTIND - 1))
```
pandar00: Unfortunately, `shellcheck` throws an error if `*)` case is missing.
However, it isn't possible for this script since rest of the arguments are implicitly treated as argument to `verify-jar`.
There is an [open conversation](https://github.com/delphix/appliance-build/pull/241/files#r262644047) on whether we want to remove `verify` script and call `verify-jar` and `verify-stack` directly.

upgrade/upgrade-scripts/upgrade
prakashsurya: Rather than having both an `upgrade` and `verify` script, how about we add a "dry-run" option to the existing `upgrade` script? I opened a PR to do that here: https://github.com/delphix/appliance-build/pull/243

My main concerns with the current `verify` script, is it has a lot of duplication between the `upgrade` script.. but more importantly, it can only be used to verify an in-place upgrade.. if we want a way to "verify" the upgrade, we should provide a way to do that for both "in-place" and "not-in-place" upgrades, which my proposed PR enables.

let me know what you think.

pandar00: Your PR looks great. I'll remove the `verify` script.

upgrade/upgrade-scripts/verify
pandar00: We decided `verify` script will be removed in favor of ["dry-run" functionality](https://github.com/delphix/appliance-build/pull/243/files) in `upgrade` script

upgrade/upgrade-scripts/upgrade
```diff
@@ -176,12 +180,13 @@ function upgrade_not_in_place() {
...
-	"$IMAGE_PATH/upgrade-container" upgrade "$CONTAINER" not-in-place ||
+	"$IMAGE_PATH/upgrade-container" upgrade "$CONTAINER" ||
 		die "Failed to upgrade do not-in-place upgrade on '$CONTAINER'"
 
 	if [[ "$DLPX_UPGRADE_SKIP_VERIFY" != "true" ]]; then
-		"$IMAGE_PATH/verify" -c "$CONTAINER" "$@" ||
-			die "'$IMAGE_PATH/verify' '$*' failed in '$CONTAINER'"
+		"$IMAGE_PATH/upgrade-container" run "$CONTAINER" \
+			"$IMAGE_PATH/verify-jar" "$@" ||
+			die "'$IMAGE_PATH/verify-jar' '$*' failed in '$CONTAINER'"
```
prakashsurya: I was thinking about this a little, and I wonder... rather than always *not* running `verify-stack` here, should we instead check to see if in-place upgrade is potentially allowed, and only skip `verify-stack` when it wouldn't be allowed?

Specifically, I'm wondering about the case where the user chose to do a "not-in-place" upgrade, when "in-place" would have been allowed.. in that case, I think it should be possible to run `verify-stack`, right?

Rather than always assuming that when "not-in-place" is used, it's because "in-place" wasn't allowed.. does this make sense?

upgrade/upgrade-scripts/verify-stack
```diff
@@ -49,7 +60,9 @@ IS_ACTIVE_POSTGRES=$(systemctl is-active delphix-postgres@default)
 
 #
 # Create a copy of MDS and start the postgres service.
-#
+
+
```
prakashsurya: extra line?

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -505,6 +505,63 @@ function migrate_configuration() {
...
+	migrate_configuration ||
+		die "failed to migrate configuration for '$CONTAINER'"
+}
+
+function get_type() {
+	local root origin
+
+	root=$(zfs list -Hpo name /var/lib/machines/"${CONTAINER}")
+	[[ -n "${root}" ]] || die "Failed to obtain root filesystem for container '${CONTAINER}'"
+	origin=$(zfs get -Hpo value origin "${root}")
```
prakashsurya: let's also add a `[[ -n "$origin" ]] || die ...` line after this, to account for the possibility that `zfs get` fails

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -505,6 +505,63 @@ function migrate_configuration() {
...
+	*)
+		echo "in-place"
+		;;
+	esac
+}
+
+function do_upgrade_container() {
+	local type
+	type=$(get_type)
+	case $type in
```
prakashsurya: can we do this:
```
case "$(get_type)" in
```
I'm not sure if that's valid bash syntax, but if so, it'll allow us to remove the `type` variable.

prakashsurya:
This is looking great. Please squash this to a single commit before landing, and also verify the scripts work properly when run from the CLI. There's still a few relatively minor comments that I left (and some style issues to fix), but I'll proactively approve this so you can land this when you're ready.

bors delegate+

pandar00:
Previous builds failed during dc image cloning
```
00:39:46 + dc guest wait -t 600 pre-push-766-143e2cf6
00:53:27 ERROR:root:Timed out waiting for instances:
00:53:27   pre-push-766-143e2cf6
[Pipeline] }
00:53:28 ERROR: script returned exit code 1
00:53:28 Retrying
[Pipeline] {
[Pipeline] sh
00:53:29 + dc guest wait -t 600 pre-push-766-143e2cf6
01:07:10 ERROR:root:Timed out waiting for instances:
01:07:10   pre-push-766-143e2cf6
[Pipeline] }
01:07:10 ERROR: script returned exit code 1
01:07:10 Retrying
[Pipeline] {
[Pipeline] sh
01:07:11 + dc guest wait -t 600 pre-push-766-143e2cf6
01:20:52 ERROR:root:Timed out waiting for instances:
01:20:52   pre-push-766-143e2cf6
```

prakashsurya:
More infrastructure issues...

bors r+

pandar00:
Push was failing due to https://jira.delphix.com/browse/INC-481

PR #242, opened 2019-03-05
LX-1783 migration: reconcile /proc/cmdline between migration and initial install

pzakha:
Adding the same command line arguments as we have in https://github.com/delphix/delphix-platform/blob/master/etc/default/grub.d/override.cfg. Note that we should make a follow-up change to set the default serial port baud rate in the console to 115200 for regular Linux installations instead of the current 38400.

Ultimately we just needed to add 2 parameters to the command line passed to the kernel:
```
	'ipv6.disable=1'
	'crashkernel=1024M-:512M'
```
However, after adding those parameters to the original script, the bootloader would fail to boot with our new string:
```
fail at line 179
Parse error!
```
After doing some testing it appears that the bootloader interpreter has a 256 characters limit for any command entered in the `ok` prompt, preventing us from typing the full command, and thus resulting in the same parse error message. The same restriction probably applies for commands read from the `menu.rc.local` file.

I had to refactor the code so that the command line arguments are stored in a variable that is referenced from the boot string, thus reducing the length of any command to < 256 characters.

## DEPENDENCIES

This change depends on https://github.com/delphix/delphix-platform/pull/60. Since we are now passing a `crashkernel` option to the cmdline, the `kdump-tools` service creates the file `/var/crash/kexec_cmd`, which causes that bug.

## TESTING

* Built a migration image and tested migration. Made sure that the new command line arguments appeared in `/proc/cmdline` and verified that ifconfig doesn't report ipv6 addresses anymore.
* Relevant build: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/630

### DIFF
--- a/live-build/misc/migration-scripts/dx_apply
+++ b/live-build/misc/migration-scripts/dx_apply
@@ -77,6 +77,9 @@ rm -f /boot/vmlinuz-* /boot/initrd.img-* ||
 zfs destroy -r "$RPOOL/ROOT" 2>/dev/null
 zfs list "$RPOOL/ROOT" 2>/dev/null &&
 	die "could not destroy linux root dataset from previous run"
+sed -i '/set lxcmdline/d' /boot/menu.rc.local
+[[ "$(grep -cF 'set lxcmdline' /boot/menu.rc.local)" -eq 0 ]] ||
+	die "failed to cleanup lxcmdline from previous run"
 sed -i '/set mainmenu_caption\[8\]/d' /boot/menu.rc.local
 [[ "$(grep -cF 'mainmenu_caption[8]' /boot/menu.rc.local)" -eq 0 ]] ||
 	die "failed to cleanup mainmenu_caption from previous run"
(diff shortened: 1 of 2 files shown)


pzakha:
cc: @sdimitro

pzakha:
> More specifically, I'm wondering if we'd switch to GRUB immediately after migration is deemed successful (when would this be?)

Yes, that would be after the stack has started successfully and marked the upgrade as successful. We have this work planned out for LX-1554.

> at which point we could force a reboot and use the non-migration settings, to be consistent with non-migrated systems?

We should avoid extra reboots unless they are really required. In this case, I don't think it matters which bootloader we booted from as long as the arguments to the kernel process are the same.

prakashsurya:
> We should avoid extra reboots unless they are really required. In this case, I don't think it matters which bootloader we booted from as long as the arguments to the kernel process are the same.

In general, I agree.. but if we change the GRUB settings, this means we now have to update the migration code too, which creates a maintenance burden.. whereas if we adopt the approach that we reboot to use GRUB early after migration is complete, we no longer have this burden.

IMO, a switch to GRUB and reboot would be useful after migration for a number of reasons: it's a one time event, VDBs will already be down, it'd help us verify the switch to GRUB actually worked properly (and potentially avoid failures later), and it'd get us using the same code paths as non-migration sooner.

Regardless, this change is fine.

pzakha:
Yeah, we can revisit that idea later when we work on LX-1554.

prakashsurya:
bors delegate+

PR #247, opened 2019-03-12
Stop using Docker when executing the build

prakashsurya:
This change modifies the Ansible bootstrap logic to install all of the
necessary build dependencies onto the host that's running Ansible. This
way, we can execute appliance-build directly on the host, rather than
relying on the Docker contianer. Additionally, the documentation has
been updated to no longer reference the use of Docker.

With that said, we still build the Docker image and support building in
the Docker contianer just as before, so we can temporarily maintain
backwards compatibility with our automation, and avoid flag days.


prakashsurya:
bors try

prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/673/)

I have an associated devops-gate change [here](http://reviews.delphix.com/r/47638). This PR will need to land first, then the devops-gate change.

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
```diff
@@ -19,31 +19,94 @@
...
+    name:
+      - ansible
+      - aptly
+      - awscli
+      - bc
+      - coreutils
+      - equivs
+      - gdisk
+      - git
+      - java-package
```
pzakha: We don't need java-package anymore as we don't build Java in appliance-build anymore.

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
```diff
@@ -19,31 +19,94 @@
   register: toplevel
 
 - apt:
-    name: aptitude
+    name: software-properties-common
```
pzakha: Why do we need this?

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
```diff
@@ -19,31 +19,94 @@
...
+    update_cache: yes
 
-- apt:
+- apt_repository:
+    repo: ppa:ansible/ansible
     update_cache: yes
 
 - apt:
-    upgrade: yes
+    upgrade: dist
```
pzakha: Could you elaborate on this?

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
```diff
@@ -19,31 +19,94 @@
...
+    wget -nv -O /usr/local/bin/shfmt \
+      https://github.com/mvdan/sh/releases/download/v2.4.0/shfmt_v2.4.0_linux_amd64 && \
+    chmod +x /usr/local/bin/shfmt
+  args:
+    creates: /usr/local/bin/shfmt
+    executable: /bin/bash
+    warn: no
+
+- shell: |
+    git clone https://github.com/willthames/ansible-lint /opt/ansible-lint && \
```
pzakha: We don't have to address this here, but Ubuntu currently provides an `ansible-lint` package, perhaps we should use that instead?

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
prakashsurya: Probably. I'll remove, retest, and make sure things still work.

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
prakashsurya: To install the ansible PPA. We add it currently in the docker container, so this isn't changing anything.

pzakha: We should probably add a comment.

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
prakashsurya: I opted to change this to "dist" so we can drop the need for the aptitude package to be installed. I also think performing "apt-get dist-upgrade" (after) vs. "aptitude safe-upgrade" (before) is "more correct".

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
prakashsurya: I'll actually remove this, and the adding of the PPA, since we only support building on the bootstrap image, and that already has ansible installed.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.1-bin.zip
```
jgallag88: By relying on the upstream site instead of our artifactory mirror, this adds another external dependency to our build. I worked around this when installing Gradle in the Docker image by first trying to download from artifactory, and then falling back to upstream if that failed, which admittedly was pretty hacky and wouldn't work here anyway.

I'm not opposed to this if we are still running some parts of the build outside the Delphix network, but I just thought that I would point it out.

prakashsurya: Can we configure this based on if we're running in the container or not? I was planning to still leverage the container for running the style checks (so we get faster feedback via the PR status), so we'd still need gradle in TravisCI. So, if we're in the container, use upstream.. if not, use artifactory.. is this possible?

Alternatively, we can remove the use of the container completely, and run the style checks via the Jenkins job (i.e. only support the artifactory URL).

@jgallag88 do you have a preference?

jgallag88: > if we're in the container, use upstream.. if not, use artifactory.. is this possible?

That sounds reasonable to me. Since the container already uses upstream, all we would have to do is update this url to point to artifactory, right?

prakashsurya: I guess you're right. The container can pull gradle just like it does now, and we can still use `gradle` (instead of `gradlew`) in Travis. I'll go with that.

prakashsurya:
I'm not sure why, but when I use this change in the Jenkins job (i.e. with the jenkins changes to stop building in the container), the build fails when creating the raw disk image when it attempts to export the pool:
```
18:31:12 + zpool export rpool
18:31:12 cannot export 'rpool': pool is busy
```
I'll have to investigate this failure before integrating.

jgallag88:
I ran into this while attempting a similar change. I believe the problem is that some of the filesystems from `rpool` are still mounted in other mount namespaces. In namespace that my shell is running in, nothing from the pool is mounted:
```
$ cat /proc/self/mounts | grep rpool
$
```
However, checking some of the other mount spaces shows some filesystems still mounted:
```
$ sudo lsns -t mnt
        NS TYPE NPROCS   PID USER             COMMAND
4026531840 mnt     173     1 root             /sbin/init
4026531861 mnt       1    19 root             kdevtmpfs
4026532262 mnt       1   456 root             /lib/systemd/systemd-udevd
4026532263 mnt       1   594 systemd-timesync /lib/systemd/systemd-timesyncd
4026532264 mnt       1   779 systemd-network  /lib/systemd/systemd-networkd
4026532265 mnt       1   805 systemd-resolve  /lib/systemd/systemd-resolved
```
```
$ cat /proc/594/mounts | grep rpool
rpool/ROOT/delphix.v8IGiyn/root /mnt/delphix.v8IGiyn zfs rw,xattr,noacl 0 0
rpool/ROOT/delphix.v8IGiyn/home /mnt/delphix.v8IGiyn/export/home zfs rw,relatime,xattr,noacl 0 0
rpool/ROOT/delphix.v8IGiyn/data /mnt/delphix.v8IGiyn/var/delphix zfs rw,relatime,xattr,noacl 0 0
rpool/ROOT/delphix.v8IGiyn/log /mnt/delphix.v8IGiyn/var/log zfs rw,relatime,xattr,noacl 0 0
```
I don't think we intentionally mount these filesystems in other mount namespace, but perhaps something unexpected is happening with mount propagation.

PR #248, opened 2019-03-12
Build zfs gate during instance creation to remove developer overhead

pcd1193182:
This is still a rough draft of the proposed functionality, but by running through a zfs build before the instance finishes being created, we can save developer time by preventing 20+ minute waits while the zfs repo does an initial build. We could extend this to other repos (virtualization, particularly) as well. I mostly wanted to get this draft up to solicit comments on the concept and the approach.

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.zfsonlinux-development/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.zfsonlinux-development/tasks/main.yml
@@ -65,6 +65,39 @@
     accept_hostkey: yes
     update: no
 
+- shell: ls /lib/modules
+  register: kernel_versions
+
+- shell: |
+    CONFIG_OPTS=" \
+    --build=x86_64-linux-gnu \
+    --prefix=/usr \
+    --includedir=/usr/include \
+    --mandir=/usr/share/man \
+    --infodir=/usr/share/info \
+    --sysconfdir=/etc \
+    --localstatedir=/var \
+    --libdir=/usr/lib/x86_64-linux-gnu \
+    --libexecdir=/usr/lib/x86_64-linux-gnu \
+    --disable-maintainer-mode \
+    --bindir=/usr/bin \
+    --sbindir=/sbin \
+    --libdir=/lib \
+    --with-python=3 \
+    --with-udevdir=/lib/udev \
+    --with-systemdunitdir=/lib/systemd/system \
+    --with-systemdpresetdir=/lib/systemd/system-preset \
+    --with-linux=/lib/modules/{{ kernel_versions.stdout_lines[0] }}/build \
+    --with-linux-obj=/lib/modules/{{ kernel_versions.stdout_lines[0] }}/build"
+    cd /export/home/delphix/zfs
+    sh autogen.sh
+
+    ./configure $CONFIG_OPTS --enable-debug --enable-debuginfo
+
+    make -j$(nproc)
+  args:
+    executable: /bin/bash
+
 - file:
     path: "/export/home/delphix/zfs"
     owner: "delphix"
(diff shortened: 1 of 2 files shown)


live-build/misc/ansible-roles/appliance-build.zfsonlinux-development/tasks/main.yml
```diff
@@ -65,6 +65,39 @@
     accept_hostkey: yes
     update: no
 
+- shell: ls /lib/modules
+  register: kernel_versions
+
+- shell: |
+    CONFIG_OPTS=" \
```
prakashsurya: @pzakha could we move this into the zfs repository some how? so both git-utils and this code wouldn't have to do this? do we also have to do this for the zfs package build?

pcd1193182: These specific options are unique to delphix's use case, other users will probably have their own requirements for any such script, so it's not likely that such a script would land upstream. We could have it live in delphix/zfs and maintain the fork, which would probably not be prohibitively difficult. That would eliminate the need to store multiple copies of these options.  I'm not sure exactly how we do it in the zfs package build.

pzakha: We could definitely move this into the Delphix ZFS repository if we wanted. I'm assuming you got most of those options from git-zol-make. Originally, those were actually taken from the `debian/rules` in order to make the manual build as close as possible to the package build.

So I think if we add this to the internal-dev variant, the main issue is that it will add 20 minutes to all git-ab-pre-push runs. This may be fine, but we should probably have a discussion whether we want to go that route or not. Same goes for the app build.

prakashsurya: Right. I'm thinking of something that'd we'd land to delphix/zfs; e.g. changing the default values, or something similar.

I'm sure we could create some sort of "delphix-configure" script, with these values baked in, but I'm not sure if that's the best solution we can come up with.

Spreading the same values across different repositories (here, git-utils, etc) would become an issue when/if we need to change these, so if we could consolidate that'd be good.

live-build/misc/ansible-roles/appliance-build.zfsonlinux-development/tasks/main.yml
```diff
@@ -65,6 +65,39 @@
...
+    --bindir=/usr/bin \
+    --sbindir=/sbin \
+    --libdir=/lib \
+    --with-python=3 \
+    --with-udevdir=/lib/udev \
+    --with-systemdunitdir=/lib/systemd/system \
+    --with-systemdpresetdir=/lib/systemd/system-preset \
+    --with-linux=/lib/modules/{{ kernel_versions.stdout_lines[0] }}/build \
+    --with-linux-obj=/lib/modules/{{ kernel_versions.stdout_lines[0] }}/build"
+    cd /export/home/delphix/zfs
```
prakashsurya: I think you can add:
```
args:
  chdir: /export/home/delphix/zfs
```
and remove this command.. not that it really matters either way, though.

live-build/misc/ansible-roles/appliance-build.zfsonlinux-development/tasks/main.yml
```diff
@@ -65,6 +65,39 @@
     accept_hostkey: yes
     update: no
 
+- shell: ls /lib/modules
+  register: kernel_versions
+
+- shell: |
```
prakashsurya: I think we also want to add `ignore_errors: yes` (or w/e the syntax is) to prevent zfs build failures from causing appliance-build to fail.

We should probably test and verify that a zfs build failure will be ignored too.

PR #253, opened 2019-03-26
Create upgrade.properties during a migation upgrade

pzakha:
Together with the app-gate work in http://reviews.delphix.com/r/48049/, this makes sure that on the first stack startup after a migration is performed we mark the upgrade as successful.

This PR should be pushed after the app-gate is pushed.

## TESTING

migration testing: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/blackbox-chained/36/

### DIFF
--- a/live-build/misc/migration-scripts/dx_execute
+++ b/live-build/misc/migration-scripts/dx_execute
@@ -132,8 +132,17 @@ echo "Running MDS Command: '$command'"
 	die "Failed mds_client command."
 
 #
+# Let the app-stack know that an "OS migration" upgrade is being performed.
+#
+cat <<-EOF >"/var/dlpx-update/upgrade.properties" ||
+	UPGRADE_TYPE=OS_MIGRATION
+EOF
+	die "failed to create upgrade.properties"
+
+#
 # Create linux /var/delphix dataset from a clone of the current
-# /var/delphix dataset.
+# /var/delphix dataset. We want to do this last as this will carry-over all
+# the logs and the /var/delphix/migration directory into Linux.
 #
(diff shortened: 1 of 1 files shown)


pzakha:
cc: @prakashsurya

live-build/misc/migration-scripts/dx_execute
```diff
@@ -131,9 +131,17 @@ echo "Running MDS Command: '$command'"
 /opt/delphix/server/bin/mds_client -- -c "$command" ||
 	die "Failed mds_client command."
 
+#
+# Let the app-stack know that an "OS migration" upgrade is being performed.
+#
+cat <<-EOF >"/var/dlpx-update/upgrade.properties" || die "failed to create upgrade.properties"
```
grwilson: I'm not sure what is the correct structure here as I've seen other places where we do the following:

```
cat <-EOF >"/var/dlpx-update/upgrade.properties" ||
    UPGRADE_TYPE=OS_MIGRATION
EOF 
    die "failed to create upgrade.properties"
```

We should standardize on a particular convention

pzakha: I agree. I have seen many different ways of that being done. Unfortunately this is not being enforced in `shfmt` so we'd need to somehow add our own rules.

prakashsurya: Often, the "die" command is on a new line so it fits within the 80 column limit. If we don't need another line to fit in that limit, then I don't see any reason to use another line.

pzakha: I've put the die on a different line to make it fit within the 80 characters limit.

PR #257, opened 2019-03-28
LX-1810 upgrade: It should be possible to downgrade some packages on upgrade

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/773/)

I also tested this manually by building an upgrade image, cloning a few week old trunk VM, and successfully upgrading that VM with these changes. Without these changes, the upgrade will fail due to zfs packages going from a "1.0.0" based version to "0.8.0".

### TODO
- [ ] Verify the changes to remove the explicit installation of `delphix-platform` are correct
- [ ] Verify the changes w.r.t. `autoremove` and `apt-mark auto` are correct

prakashsurya:
cc: @jgallag88 @pzakha

upgrade/upgrade-scripts/execute
```diff
@@ -69,92 +70,93 @@ EOF
...
+# after explicitly installing all packages for the new "delphix-entire"
+# version, the only "auto" installed packages still remaining on the
+# system will be ones that aren't required by the new "delphix-entire".
 #
-#  - We run 'apt-get install delphix-entire-aws'
-#
-#  - Apt sees delphix-entire-aws relies on delphix-virtualization, so it
-#    marks that package and its dependencies for installation. One of
-#    its dependencies is the virtual package delphix-platform.
+dpkg-query -Wf '${Package}\n' | xargs apt-mark auto
```
pzakha: I think we need to make an exception for `delphix-kernel-$(uname -r)`. This will insure that any kernel version besides the one running currently _should_ be removed by auto-remove.

upgrade/upgrade-scripts/execute
```diff
@@ -69,92 +70,93 @@ EOF
...
-apt_get install -y "delphix-platform-$platform" ||
-	die "failed to install delphix-platform"
-
+#        a) also removing the "delphix-entire" package, or
+#        b) building and installing new "delphix-entire" package that
+#           depends on the new package being installed.
 #
-# Now that we know we have a delphix-platform-<platform> packages
-# installed, go ahead and install delphix-entire-<platform>.
+#    Both options, (a) and (b), are undesireable; by not using package
```
pzakha: typo: undesirable

upgrade/upgrade-scripts/execute
```diff
@@ -69,92 +70,93 @@ EOF
...
+#    dependencies we avoid having to deal this problem.
```
pzakha: typo: deal with

upgrade/upgrade-scripts/execute
```diff
@@ -69,92 +70,93 @@ EOF
...
 #
-dpkg-query -Wf '${Depends}' "delphix-entire-$platform" |
-	sed 's/, /\n/g' |
-	sed 's/^\(.*\) (\(.*\) \(.*\))$/\1 \2 \3/g' |
-	while read -r package op dependency; do
-		installed=$(dpkg-query -Wf '${Version}' "$package")
-		compare_versions "$installed" "$op" "$dependency" ||
-			die "'$package' package version incompatible;" \
-				"'$installed' '$op' '$dependency'"
+zcat "/usr/share/doc/delphix-entire-$platform/packages.list.gz" | sed 's/=/ /' |
```
pzakha: I think it would be valuable to also print all the packages that are installed that are not part of the current delphix-entire list. This should show packages pinned by the previous delphix-kernel version (given the changes I proposed above). This would be quite useful for debugging.

upgrade/upgrade-scripts/execute
```diff
@@ -69,92 +70,93 @@ EOF
...
-# fundemental piece of the architecture of the appliance. Among other
-# things, it is the mechanism that we use to orchestrate the installation
-# all of the platform-dependant bits of the appliance.
+# 2. Often times, developers want to build and install upgraded packages
+#    that "delphix-entire" would depend on. Since the dependencies of
+#    the "delphix-entire" package would have to depend on explicit
+#    versions of each package, this would prevent packages from being
+#    upgraded without either:
 #
-apt_get install -y "delphix-platform-$platform" ||
```
pzakha: I think we might have problems if we remove the explicit installation of delphix-platform. We have some dependencies where delphix-platform and its packages must be configured before installing other packages (e.g. delphix-platform creates the postgres user on postinst which should be done before installing postgres).

Also I don't believe that installing each package explicitly is any different from installing it through delphix-entire as a dependency, thus if we needed to install delphix-platform first before, we'd still need it now.

upgrade/upgrade-scripts/execute
prakashsurya: Intersting, I'll try that out.

In my testing last night, even the previous/current kernel was removed, so hopefully this will fix that issue.

upgrade/upgrade-scripts/execute
prakashsurya: > Also I don't believe that installing each package explicitly is any different from installing it through delphix-entire as a dependency, thus if we needed to install delphix-platform first before, we'd still need it now.

@jgallag88 do you have any thoughts on if we still need to explicitly install `delphix-platform` with these changes? is there any way to test and verify if it's still needed?

I'll plan on keeping it since I'm not really sure either way, but it'd be nice to know.

upgrade/upgrade-scripts/execute
prakashsurya: Do you think this is something we should do as part of the upgrade scripts? IMO, that's just adding more output that'll likely get ignored, so I'm wondering if that'd be better left as something we do when gathering a support bundle.. what do you think?

prakashsurya:
@pzakha I think I've addressed all of your feedback; if you could give this another look when you have time, that'd be great.

upgrade/upgrade-scripts/execute
pzakha: Yeah, I think that's a good idea to put that into the bundle instead

prakashsurya: OK, I'll open a bug to get that done.

upgrade/upgrade-scripts/execute
```diff
@@ -125,36 +132,112 @@ apt_get install -y "delphix-platform-$platform" ||
...
+# mark all packages currently installed as "auto" installed. This way,
+# after explicitly installing all packages for the new "delphix-entire"
+# version, the only "auto" installed packages still remaining on the
+# system will be ones that aren't required by the new "delphix-entire".
+#
+dpkg-query -Wf '${Package}\n' | xargs apt-mark auto ||
+	die "failed to mark packages as 'auto' installed"
+
+#
+# While we marked all packages to "auto" above, so they'd be removed
```
pzakha: Do we know the amount of extra space that those kernel packages would consume? I presume that they would get cleaned-up from time to time by not-in-place upgrades, but it would be a good idea to know if we need to implement some cleanup beyond a reasonable number of kernels.

prakashsurya: I don't know off-hand, but I can find out for the currently install package(s).

prakashsurya: I don't really know if this is accurate, or what the units are:
```
$ apt-cache depends delphix-kernel-4.15.0-1032-aws | grep Depends | awk '{print $2}' | xargs dpkg-query -Wf '${Installed-Size}\n' | awk '{s+=$1} END {printf "%.0f\n", s}'
96171
```

So here's another (more complicated) attempt:
```
$ apt-cache depends delphix-kernel-4.15.0-1032-aws | grep Depends | awk '{print $2}' | xargs -n 1 dpkg -L | while read path; do [[ -f $path ]] && echo $path; done | xargs -n 1 du -b | sort -n
31      /usr/share/doc/linux-aws/changelog.gz
178     /usr/share/doc/connstat-module-4.15.0-1032-aws/changelog.Debian.gz
383     /usr/share/doc/connstat-module-4.15.0-1032-aws/copyright
436     /etc/depmod.d/zfs-modules-4.15.0-1032-aws.conf
530     /usr/share/doc/linux-tools-aws/changelog.gz
684     /usr/share/doc/zfs-modules-4.15.0-1032-aws/README.Debian
1292    /usr/share/doc/linux-image-4.15.0-1032-aws/copyright
1598    /usr/share/doc/linux-aws/copyright
1598    /usr/share/doc/linux-tools-aws/copyright
4507    /usr/share/doc/zfs-modules-4.15.0-1032-aws/changelog.Debian.gz
7848    /lib/modules/4.15.0-1032-aws/kernel/net/connstat/connstat.ko
40358   /usr/share/doc/linux-image-4.15.0-1032-aws/changelog.Debian.gz
53390   /usr/share/doc/zfs-modules-4.15.0-1032-aws/copyright
240112  /lib/modules/4.15.0-1032-aws/extra/avl/zavl.ko
1055592 /lib/modules/4.15.0-1032-aws/extra/unicode/zunicode.ko
1183384 /lib/modules/4.15.0-1032-aws/extra/nvpair/znvpair.ko
3629312 /lib/modules/4.15.0-1032-aws/extra/zcommon/zcommon.ko
4466232 /lib/modules/4.15.0-1032-aws/extra/spl/spl.ko
7919472 /boot/vmlinuz-4.15.0-1032-aws
8076512 /lib/modules/4.15.0-1032-aws/extra/lua/zlua.ko
12367512        /lib/modules/4.15.0-1032-aws/extra/icp/icp.ko
59238592        /lib/modules/4.15.0-1032-aws/extra/zfs/zfs.ko
```
and summing these up, gives:
```
$ apt-cache depends delphix-kernel-4.15.0-1032-aws | grep Depends | awk '{print $2}' | xargs -n 1 dpkg -L | while read path; do [[ -f $path ]] && echo $path; done | xargs -n 1 du -b | sort -n | awk '{s+=$1} END {printf "%.0f\n", s}'
98289553
```
So that's just about 93MB, and doesn't include the kernel's debug symbols package.

The information for the kernel's "dbgsym" package is here:
```
 dpkg -L linux-image-4.15.0-1032-aws-dbgsym | while read path; do [[ -f $path ]] && echo $path; done | xargs -n 1 du -b | sort -n | awk '{s+=$1} END {printf "%.0f\n", s}'
3636521147
```
So, about 3.3GB just for the kernels debug symbols package.

I think it'd be reasonable to take a conservative approach and say we'll need at least 4GB to hold the kernel packages for each old release that happens to get installed on the system (assuming we keep the old kernel "dbgsym" package).

prakashsurya: IMO, I'd like to leave this as-is, and then if we determine keeping all kernel versions installed is a problem (e.g. due to the space consumption), we can address it later.

pzakha: I'm fine with addressing that later. Let's open a bug for that so that we can track this. 

Regarding debug symbols, I don't believe we currently install them on the external-standard image, do we? I think they would make the upgrade image double in size.

prakashsurya: We install it on the external images. We add the "dbgsym" packages in the "minimal-common" role, so it should be in all variants. We want them all variants so we can use kernel debug tools like `crash` (at least, that's been my understanding).

upgrade/upgrade-scripts/execute
jgallag88: It's not clear to me. Apt might or might not treat `apt-get install foo` the exactly same `apt-get install bar baz` with respect to how it resolves dependencies if `foo` depends on `bar` and `baz`. However, it doesn't make any guarantees either way, so it seems safer to keep the explicit installation of `delphix-platform`, like you have it now.

As for testing, I was just grepping through the `apt-get` for packages for other kernels (`*kvm*`, `*azure*`, etc) to see if they were selected at any point.

PR #259, opened 2019-03-30
DLPX-63341 Optimize build upgrade image job for Linux

dtzeng:
The current `buildUpgradeImage` task depends on the full `:live-build:build<variant><platform>` task.
However, this live-build task runs all the hooks which are not necessary if we just want to build an upgrade image. To fix this, notice that the live build hooks are split into 3 "phases":
1. run configuration (ansible)
2. build upgrade artifact (debs.tar.gz)
3. build all other VM artifacts
These hooks are now moved to their own directories.

I've refactored the `run-live-build.sh` script to take 3 run types: "upgrade", "vm", or "all", each corresponding to step 2 hooks only, step 3 hooks only, or both step 2 and 3 hooks, respectively.
All run types will run step 1 (configuration) regardless.

Now, each run type will be created as its own gradle task:
- `build<variant><platform>UpgradeArtifacts`: invokes `run-live-build.sh upgrade`
- `build<variant><platform>VmArtifacts`: invokes `run-live-build.sh vm`
- `build<variant><platform>`: invokes `run-live-build.sh all`

The inputs and outputs of each task are defined accordingly.

With these new tasks, we can now just have the `buildUpgradeImage` task depend on `buildUpgradeArtifacts`, and developers can also run the `buildVmArtifacts` task if they do not wish to build the upgrade image.

Testing on jenkins review http://reviews.delphix.com/r/48120


dtzeng:
@pandar00 @jgallag88 @prakashsurya

live-build/build.gradle
```diff
@@ -75,13 +122,72 @@ for (variant in allVariants) {
...
+    }
+}
+
+/*
+ * Create tasks to build VM artifacts. For instance, the task 'buildInternalDevEsxVmArtifacts' will produce
+ * an artifact 'internal-dev-esx.ova'. These tasks depend on their configuration task counterpart.
+ */
+for (variant in allVariants) {
+    for (platform in allPlatforms) {
+        def taskSuffix = "${toCamelCase(variant).capitalize()}${platform.capitalize()}"
```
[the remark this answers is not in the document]
dtzeng: As discussed offline, I've moved most of the work into `run-live-build.sh` instead of gradle tasks so that incremental builds won't break (updated PR description)
Also parameterized the run type into a for loop to remove duplicated code.

prakashsurya:
Would it make things more simple if we could refactor how the "hooks" are executed, perhaps not relying on live-build to execute them, and leverage gradle instead? I feel like the "build logic" (both gradle, and run-live-build.sh) are getting more complicated, and I'm wondering if this additional complexity can be simplified.

Before we had gradle, we used live-build to orchestrate the build.. but now that we have gradle, I would be in favor of relying less and less on live-build, and rely more and more on gradle instead. Honestly, at this point, I don't think live-build is providing that much value; we could probably replace it with gradle + debootstrap more directly, and perhaps achieve a solution that aligned better with our requirements.

I'm not necessarily suggesting we rip out live-build in this PR, but I at least wanted to bring up the topic in case relying less on live build could help make this PR more simple. E.g. how this change splits out the "hooks" directory, and then has to copy the correct subdirectory in run-live-build.sh based on the "run type" just makes me wonder if our use of live-build for that is only making things more complicated than if we used gradle to execute the hooks directly..

scripts/run-live-build.sh
```diff
@@ -51,6 +67,7 @@ set -o xtrace
 
 export APPLIANCE_VARIANT="$1"
 export APPLIANCE_PLATFORM="$2"
+export RUN_TYPE="$3"
 export ARTIFACT_NAME="$APPLIANCE_VARIANT-$APPLIANCE_PLATFORM"
```
prakashsurya: Does "ARTIFACT_NAME" need to also contain the "RUN_TYPE" so that we have a unique "build_dir" for each run type? e.g. if we ran the "VmArtifacts" task, then the "UpgradeArtifacts" task, will this cause any problems? The second task will remove the build directory of the first task, right? I don't know if this is a problem, but wanted to ask in case this wasn't considered.

dtzeng:
@prakashsurya I used to have it structured in a more "gradle"-ized way:
- buildConfig task would run live-build with step 1 hooks
- buildUpgradeImage task would depend on buildConfig and run live-build with step 2 hooks
- buildOtherArtifacts task would depend on buildConfig and run live-build with step 3 hooks
- buildAll (what we have now) would just simply depend on buildUpgradeImage and buildOtherArtifacts

However, as @jgallag88 mentioned, this would break incremental builds as buildConfig would not declare any outputs.  So, I went with this way instead.

Either way though, we would still need to refactor the hooks directory to be a little more complicated, but I'm not sure of a better way at the moment.

scripts/run-live-build.sh
dtzeng: This would actually be an issue with the older way (see below comment), but this way only runs each task once (there are no dependencies), so this is fine

live-build/build.gradle
```diff
@@ -48,46 +48,103 @@ def artifactTypes = ["aws": "vmdk",
                      "gcp": "gcp.tar.gz",
                      "kvm": "qcow2"]
 
+def configDir = "configuration"
+
+def upgradeArtifactsTaskSuffix = "UpgradeArtifacts"
+def upgradeArtifactsRunType = "upgrade-artifacts"
+def vmArtifactsTaskSuffix = "VmArtifacts"
+def vmArtifactsRunType = "vm-artifacts"
+def allTaskSuffix = ""
```
jgallag88: The `*Suffix` variables seem to only be used in one spot each. Can we inline them?

scripts/run-live-build.sh
```diff
@@ -27,12 +27,28 @@ if [[ $EUID -ne 0 ]]; then
 	exit 1
 fi
 
-if [[ $# -ne 2 ]]; then
-	echo "Must specify a single variant and a single platfom " \
-		"(e.g. 'internal-minimal esx')." 1>&2
+if [[ $# -ne 3 ]]; then
+	echo "Must specify a single variant, a single platform, and a run " \
+		"type (e.g. 'internal-minimal esx upgrade')." 1>&2
```
jgallag88: [nit] should this be `e.g. 'internal-minimal esx upgrade-artifacts'`?

PR #262, opened 2019-04-04
Use the "upgrade.properties" file for upgrade/rollback

prakashsurya:
This change modies upgrade and rollback to use the "upgrade.properties"
file, so that upgrading and rolling back via the command line (i.e. via
the upgrade scripts) integrates better with the upgrade and rollback
logic contained in the virtualization application.

The goal(s) being:

1. If an upgrade occurs via the command line, the virtualization logic
   should be able to detect this and roll the upgrade back, if the
   virtualization application determines the upgrade to have failed.

2. If the upgrade occurs via the virtualization application, we should
   be able to rollback the upgrade via the command line.

3. If the upgrade occurs via the command line, rollback via the command
   line should continue to work.

4. If the upgrade occurs via the virtualization application, rollback
   via the virtualization application should continue to work.

### DIFF
--- a/upgrade/upgrade-scripts/common.sh
+++ b/upgrade/upgrade-scripts/common.sh
@@ -136,3 +136,27 @@ function verify_upgrade_in_place_is_allowed() {
 			"($MINIMUM_REBOOT_OPTIONAL_VERSION)"
 	fi
 }
+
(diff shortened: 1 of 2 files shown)


prakashsurya:
cc: @dtzeng @pandar00

PR #263, opened 2019-04-04
LX-1771 Run masking upgrade checks from dx_verify

shartse:
**Overview**
These changes are also based very closely on the original `dx_verify`. The main differences are related to the `dmsuites`/`masking` name change. 

Prepare the `/var/delphix` clone required for the Flyway migration job in `MDSVerify` and run the masking properties check.  

**Testing**
_Automated_ 
Chained migration job: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/blackbox-chained/80/console
Job completed successfully, all services started as expected:
```
> systemctl list-units --state=failed 
0 loaded units listed.
```
_Manual_
Set up an engine using a masking blackbox suite and checked that the masking properties check worked as expected with real values in the `override.properties` file.

Made a bogus masking property and saw that the appropriate warning was raised:
``` 
/opt/jdk/bin/java -cp '/tmp/tmp.ZDai2F/upgrade-verification/*:/tmp/tmp.SGayEF/opt/delphix/masking/conf/:/var/delphix/dmsuite/overrides/' com.dmsuite.VerifyProperties
[main] INFO com.dmsuite.common.utils.properties.DMProperties - Loading properties
[main] ERROR com.dmsuite.common.utils.properties.DMProperties - Unrecognized property 'FOO' found in override.properties
Exception in thread "main" java.lang.RuntimeException: Unrecognized properties found in override.properties
        at com.dmsuite.common.utils.properties.DMProperties.checkOverrides(DMProperties.java:213)
        at com.dmsuite.common.utils.properties.DMProperties.init(DMProperties.java:173)
        at com.dmsuite.VerifyProperties.main(VerifyProperties.java:21)
+ die 'verification of overrides failed'
```

### DIFF
--- a/live-build/misc/migration-scripts/dx_verify
+++ b/live-build/misc/migration-scripts/dx_verify
@@ -18,6 +18,7 @@ export MDS_SNAPNAME="MDS-CLONE-upgradeverify"
 
 DEBUG=false
 
+MDS_CLONE=domain0/$MDS_SNAPNAME
 UPGRADE_VERIFY_PATH=/opt/delphix/server/lib/exec/upgrade-verify
 UPGRADE_VERIFY_JAR=$UPGRADE_VERIFY_PATH/upgrade-verify.jar
 LOG_DIR=/var/delphix/server/upgrade-verify
@@ -37,16 +38,29 @@ function usage() {
 
 function cleanup() {
 	if ! $DEBUG; then
-		if [ -n "$root" ]; then
+		if [[ -n "$root" ]]; then
 			cleanup_postgres
 			cleanup_datasets
-			#TODO LX-1771 cleanup masking
+			cleanup_masking
 		fi
 	fi
 }
 
 function die() {
-	#TODO save MDS logs, if they exist
+	report "$(basename "$0"): $*" >&2
+
+	#
+	# If we have MDS postgres logs, save them before cleaning up so
+	# they can be analyzed later.
+	#
+	if [[ -f "/$MDS_CLONE/db/stdout.log" ]]; then
+		(
+			cd "/$MDS_CLONE/db" || exit 1
+			tar -cvf /var/delphix/server/db/pg_log.upgrade.$$.tar \
+				stdout.log pg_log
+		)
+	fi
+
 	cleanup
 	exit 1
 }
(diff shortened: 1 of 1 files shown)


live-build/misc/migration-scripts/dx_verify
```diff
@@ -74,10 +88,26 @@ function mount_datasets() {
...
+	zfs destroy -d "$runningVar@$MDS_SNAPNAME" ||
+		die "unable to defer destroy $runningVar@$MDS_SNAPNAME"
+	mount -F zfs "$runningVar/$MDS_SNAPNAME" "$root/var/delphix" ||
+		die "unable to mount $root/var/delphix"
+
+	java=/opt/jdk/bin/java
 }
 
 function cleanup_datasets() {
+	if [ -n "$runningVar" ]; then
```
jgallag88: Here and below, should we be using a Bash-style conditional (`[[ ]]` ), instead of the posix shell style (`[ ]`)?

live-build/misc/migration-scripts/dx_verify
```diff
@@ -128,13 +184,19 @@ done
 
 # Where root is or will be mounted. Set in mount_datasets.
 root=
+# The path of the running dataset mounted at /var/delphix. Set in mount_datasets.
+runningVar=
+# Temporary diresctory used by the masking checks. Set in test_masking.
+masking_temp_dir=
+# The installation of java to use when executing our jars.
+java=
```
jgallag88: Since the copy of java that we are using isn't coming from the new version, could we just set the value here? If the goal is to keep this script as similar as possible to the Illumos upgrade dx_verify, then I could also understand keeping it as is.

shartse: I think that's a good idea.

live-build/misc/migration-scripts/dx_verify
```diff
@@ -74,10 +88,24 @@ function mount_datasets() {
...
+	zfs clone "$runningVar@$MDS_SNAPNAME" "$runningVar/$MDS_SNAPNAME" ||
+		die "unable to create $runningVar/$MDS_SNAPNAME"
+	zfs destroy -d "$runningVar@$MDS_SNAPNAME" ||
+		die "unable to defer destroy $runningVar@$MDS_SNAPNAME"
+	mount -F zfs "$runningVar/$MDS_SNAPNAME" "$root/var/delphix" ||
+		die "unable to mount $root/var/delphix"
 }
 
 function cleanup_datasets() {
+	if [[ -n "$runningVar" ]]; then
```
sdimitro: Question: Is there any case where the control flow of the script gets here and `runningVar` is empty?

I don't think we should change this code either way as you never know if this function will be reused in the future. I'm just curious.

live-build/misc/migration-scripts/dx_verify
```diff
@@ -37,16 +38,29 @@ function usage() {
 
 function cleanup() {
 	if ! $DEBUG; then
-		if [ -n "$root" ]; then
+		if [[ -n "$root" ]]; then
```
sdimitro: Same question here.

live-build/misc/migration-scripts/dx_verify
shartse: Yeah - if we call `die` in the first half of `mount_datasets`, then we'll call `cleanup` but this will be unset.

live-build/misc/migration-scripts/dx_verify
shartse: Yeah, if we can't parse the root dataset name properly: `[[ -n $rds ]] || die "could not find migration dataset for version $version"`

live-build/misc/migration-scripts/dx_verify
sdimitro: ugh I completely missed the call to `cleanup` on `die`. Thanks for pointing it out!

prakashsurya:
bors delegate+

PR #264, opened 2019-04-04
LX-1826 /etc/issue file created/populated after not-in-place upgrade

pandar00:
As noted in the bug, we are seeing a default `/etc/issue` file present on VM upgraded with "not-in-place" upgrade.

**Cause**
* The current implementation of `migrate_file` function only migrates the file to the container if it exists on the host. This means if a host doesn't have the file but the container somehow does, then it'll leave the file in the container.
This case happens with `/etc/issue` file where host doesn't have the file, but the `not-in-place` container created it while executing`debootstrap` which has the content "Ubuntu 18.04.2 LTS \n \l".

**Fix**
* We force delete the migration target file on the container file system if the file doesn't exist on the host.

### DIFF
--- a/upgrade/upgrade-scripts/upgrade-container
+++ b/upgrade/upgrade-scripts/upgrade-container
@@ -428,6 +428,15 @@ function migrate_file() {
 			die "'mkdir -p $parentdir' failed for '$CONTAINER'"
 		cp "$path" "/var/lib/machines/${CONTAINER}${path}" ||
 			die "'cp $path' failed for '$CONTAINER'"
+	else
+		if [[ -f "/var/lib/machines/${CONTAINER}${path}" ]]; then
+			# If the target file doesn't exist in the host then
+			# we delete the file. This case could happen if the host
+			# configuration doesn't exist but a fresh install placed
+			# package default configuration.
+			rm "/var/lib/machines/${CONTAINER}${path}" ||
+				die "failed to remove '$path' in '$CONTAINER'"
+		fi
 	fi
 }
 


pandar00:
@dtzeng @prakashsurya 
I'll continue testing if you guys agree with the fix

prakashsurya:
I'm not sure I agree with this approach in general, but let me think about it a bit more.

For the `/etc/issue` file specifically, we remove it during the initial build, via an ansible role; see [here](https://github.com/delphix/appliance-build/blob/master/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml#L53).

So, I wouldn't be opposed to something more targeted to this file in particular.. but maybe what you propose is actually best, in general.. i.e. if the file doesn't exist pre-upgrade, perhaps it shouldn't exist post-upgrade either. Hm..

pandar00:
> So, I wouldn't be opposed to something more targeted to this file in particular.. 

I'm fine with this approach too

I'm also not in any hurry to fix the bug so take time as needed. I'll also try to see if there are better options.

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -428,6 +428,12 @@ function migrate_file() {
 			die "'mkdir -p $parentdir' failed for '$CONTAINER'"
 		cp "$path" "/var/lib/machines/${CONTAINER}${path}" ||
 			die "'cp $path' failed for '$CONTAINER'"
+	else
+		# If a migration target file doesn't exist pre-upgrade then
```
prakashsurya: Can we reword this to something like: "If the target file doesn't exist" .. it's minor, but since we use "migration" as the name for illumos to linux upgrades, it can be a little confusing to call the file a "migration target file" (e.g. if one assumes "migration" used here means "illumos to linux migration").

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -428,6 +428,12 @@ function migrate_file() {
...
+		# we delete the file. This case could happen if pre-upgrade
+		# configuration doesn't exist but a refresh install placed
```
prakashsurya: Should "refresh" be "fresh" ?

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -428,6 +428,12 @@ function migrate_file() {
...
+		# package default configuration.
+		rm -f "/var/lib/machines/${CONTAINER}${path}"
```
prakashsurya: add `|| die "failed to remove '$path' in '$CONTAINER'"` in case `rm` fails? since we're using `-f`, even if the file doesn't exist in the container, it shouldn't report an error.

pandar00:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/874/)

PR #265, opened 2019-04-08
Print curl errors

jgallag88:
We recently had a couple jobs fail because a request made using `curl` failed. However, we don't know what the error was because we use the `-s`/`--silent` flag. This change adds the `-S` flag, which makes `curl` print errors when it fails.


jgallag88:
Yes, I reported it here: https://jira.delphix.com/browse/TOOL-7515. I wanted to open this though so that if it happens again we could report a more specific error.

jgallag88:
ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/826/console

PR #266, opened 2019-04-10
LX-1812 Increase freebsd bootloader timeout

pzakha:
The current 1 second timeout is often too short to be able to choose the right bootloader option through the ESX console or the serial console on dcenter.

As we are getting ready for the Lumen migration being able to have more time in the bootloader could become more critical to insure better supportability.

## TESTING
- Created image: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/833/
- Tested that migration works
- Verified that bootloader timeout was actually increased to 20 seconds

### DIFF
--- a/live-build/misc/migration-scripts/dx_apply
+++ b/live-build/misc/migration-scripts/dx_apply
@@ -292,13 +292,6 @@ MAIN_MENU_FICL=(
 	die "failed to update /boot/menu.rc.local"
 
 #
-# Increase the automatic boot delay to 20 seconds. This would help debug
-# any potential boot issues.
-#
-sed -i '/set autoboot_delay/c\set autoboot_delay=20' /boot/menu.rc.local ||
-	die "failed to change the autoboot_delay"
-
-#
 # Note that the new /var/delphix should not contain any useful data as it will
 # be replaced by a clone of the current /var/delphix dataset in dx_execute.
 # We do not clone the current /var/delphix right away as the clone will be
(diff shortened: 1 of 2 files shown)


pzakha:
cc: @sdimitro @sebroy

sdimitro:
Maybe I'm missing something, but how is the above different than the following in `dx_apply`:
```
#
# Increase the automatic boot delay to 20 seconds. This would help debug
# any potential boot issues.
#
sed -i '/set autoboot_delay/c\set autoboot_delay=20' /boot/menu.rc.local ||
	die "failed to change the autoboot_delay"
```

pzakha:
> Maybe I'm missing something, but how is the above different than the following in `dx_apply`:
> 
> ```
> #
> # Increase the automatic boot delay to 20 seconds. This would help debug
> # any potential boot issues.
> #
> sed -i '/set autoboot_delay/c\set autoboot_delay=20' /boot/menu.rc.local ||
> 	die "failed to change the autoboot_delay"
> ```

Hmm, I don't recall actually pushing those changes...

Well I'm wondering if we shouldn't remove the one in dx_apply, since dx_apply would have those changes applied on verification.

sdimitro:
I'm ok with it either way, I just wanted to make sure that we keep one of them.

pzakha:
Ran another round of testing with the latest changes.

Issue #157, opened 2018-11-19
Use pigz and enable verbosity for tar when generating the upgrade artifact


sdimitro:

basil:
Most of this task was covered in #267. There are just two more things that need to be done:

- Consistently enabling verbosity when calling `tar(1)`. Right now there are still a few `tar(1)` invocations that don't use the `-v` flag.
- Making `90-linux-migration-artifact.binary` invoke `pigz(1)` with the `-I pigz` argument to `tar(1)` rather than a pipe. As mentioned in #267, this would make all usage of `pigz(1)` consistent.

PR #267, opened 2019-04-15
Use pigz(1) when compressing tarballs.

basil:
`pigz(1)` is a parallel implementation of `gzip(1)` which speeds up builds by taking advantage of multiple cores. `tar -I pigz` is a drop-in replacement for `tar -z`.

This change replaces all calls to `tar -z` with `tar -I pigz`. Note that according to [the `pigz(1)` man page](https://manpages.ubuntu.com/manpages/bionic/man1/pigz.1.html), "decompression can't be parallelized […]. As a result, `pigz` uses a single thread (the main thread) for decompression." Therefore, I have only modified the `tar` invocations that pertain to compression rather than the invocations that pertain to decompression.

On our two-core build VMs, the change results in a modest but noticeable performance improvement. For example, we save 2.5 minutes building the GCP image. Before:

```
37:25 + tar -z --transform s/internal-dev-gcp.img/disk.raw/r -cvf internal-dev-gcp.gcp.tar.gz internal-dev-gcp.img
37:25 internal-dev-gcp.img
51:52 + [[ gcp == kvm ]]
```

After:

```
13:33 + tar -I pigz --transform s/internal-dev-gcp.img/disk.raw/r -cvf internal-dev-gcp.gcp.tar.gz internal-dev-gcp.img
13:33 internal-dev-gcp.img
25:28 + [[ gcp == kvm ]]
```

### DIFF
--- a/live-build/config/hooks/upgrade-artifacts/82-upgrade-repository.binary
+++ b/live-build/config/hooks/upgrade-artifacts/82-upgrade-repository.binary
@@ -133,6 +133,6 @@ rm -rf "binary/$DEBS_DIRECTORY"
 # This intermediate build artifact is eventually used to build an
 # upgrade image. See scripts/build-upgrade-image.sh for details.
 #
-tar -czf "$ARTIFACT_NAME.debs.tar.gz" ./*.deb -C binary/packages .
+tar -I pigz -cf "$ARTIFACT_NAME.debs.tar.gz" ./*.deb -C binary/packages .
 
 rm -rf binary/packages
--- a/live-build/config/hooks/vm-artifacts/91-gce-machine-image.binary
+++ b/live-build/config/hooks/vm-artifacts/91-gce-machine-image.binary
@@ -28,5 +28,5 @@
 # Google cloud only accepts gzip-ed tarballs with a single "disk.raw"
 # physical image.
 #
-tar -z --transform "s/$ARTIFACT_NAME.img/disk.raw/r" \
+tar -I pigz --transform "s/$ARTIFACT_NAME.img/disk.raw/r" \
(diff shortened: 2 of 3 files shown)


basil:
**Testing done:** `git ab-pre-push -v 'internal-dev internal-qa' -p 'aws azure esx gcp' --tests` including upgrade test from trunk to trunk (by setting the job parameter `TEST_UPGRADE_FROM_VERSION=trunk`). Run is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/854/consoleFull).

basil:
> Can you update the “90-linux-migration-artifact.binary” script to be consistent with these calls too? 

The `90-linux-migration-artifact.binary` script already uses `pigz(1)` and doesn't need to be updated:

```
tar -cvf - $DEPOT_DIRECTORY | pigz >"$MIGRATION_ARTIFACT"
```

prakashsurya:
Right, but now we have two ways that invoke pigz.. I’d rather us be consistent, and always use “| pigz” or always use “-I pigz”.

I’m fine with this as-is, if you disagree.

basil:
> Right, but now we have two ways that invoke pigz.. I’d rather us be consistent, and always use “| pigz” or always use “-I pigz”.

Testing migration isn't something I have time for right now. How about if we repurpose #157 to cover adding the verbosity flags (which I am not doing in this change) and making the migration script use `-I`?

prakashsurya:
IMO, if there isn’t a easy, turn-key, way to test the migration related change, then we should accept occasional breakage when making semi-unrelated changes like this.

But I’m good either way; I’m ok with your suggestion.

PR #271, opened 2019-04-16
LX-1161 enable verification of upgrade repository using hash and signature

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/870/)

prakashsurya:
cc: @rasantel @dtzeng @pandar00 @grwilson @pzakha

scripts/build-upgrade-image.sh
```diff
@@ -71,6 +71,49 @@ aptly publish repo -skip-contents -skip-signing upgrade-repository
 VERSION=$(dpkg -f "$(find debs/ -name 'delphix-entire-*' | head -n 1)" version)
 sed "s/@@VERSION@@/$VERSION/" <version.info.template >version.info
 
-tar -I pigz -cf "$APPLIANCE_VARIANT.upgrade.tar.gz" version.info -C upgrade-scripts . -C ~/.aptly .
+sha256sum version.info >SHA256SUMS
+for dir in upgrade-scripts ~/.aptly; do
```
sanbhara: Hi Prakash, as discussed offline, we could evaluate an alternate approach where we can reduce number of hashes we need to generate as below :

generate a hash of the tarball that we currently use
then sign that hash
bundle the signature, hash, and tarball in another tarball

Advantages would be :
a) small number of hashes to be computed (1 or maybe 2 if we pull out version.info to outer tarball)
b) simpler code both here and in unpack-image

Potential con could be the amount of time it takes to unpack on the engine. Need to compare with the amount of time it takes to generate the large number of hashes (which needs to be repeated during unpack to verify the hash) vs hashing a large file.

Security wise, both approaches achieve what we need.

scripts/build-upgrade-image.sh
```diff
@@ -71,6 +71,49 @@ aptly publish repo -skip-contents -skip-signing upgrade-repository
...
+		jq -r .signature <request.response >signature.base64
+		base64 -d signature.base64 >"signature.${signature_version}"
+
+		rm request.response signature.base64
+	done
+fi
+
+tar -I pigz -cf "$APPLIANCE_VARIANT.upgrade.tar.gz" \
+	"$(ls signature.* 2>/dev/null)" \
+	SHA256SUMS \
```
pandar00: What's the use case for including `SHA256SUMS` file in an upgrade image?

sanbhara: apart from checking that the signature file was generated by delphix, we also need to verify that the signed hashes match the software in the tarball. Both of these together satisfies the security needs.

prakashsurya: correct; my plan is to update the `unpack-image` script (in the delphix-platform repo) to run: `sha256sum -c SHA256SUMS` immediately after extracting the upgrade image, so we can verify the files contained in the image haven't been modified.

prakashsurya:
When I try to verify the signature with this change, it fails:
```
$ openssl dgst -sha256 -verify /opt/delphix/server/etc/key-public.pem -signature SHA256SUMS.sig.5.3 SHA256SUMS
Verification Failure
```
am I doing this correctly?

sanbhara:
That statement to verify the signature looks good to me. I am guessing (hoping :)) we do not have direct access to the delphix private key so not sure how to verify this. One way to just POC the code is to generate a key pair and test the rest of the logic. Was this working with the previous implementation?

1.Create private/public key pair
openssl genrsa -out private.pem 1024
2. Extracting Public key.
openssl rsa -in private.pem -out public.pem -outform PEM -pubout
3. Sign the SHA256SUMS file
openssl dgst -sha256 -sign private.pem -out SHA256SUMS.sig.5.3 SHA256SUMS
4. Verify the signature
openssl dgst -sha256 -verify public.pem -signature SHA256SUMS.sig.5.3 SHA256SUMS

scripts/build-upgrade-image.sh
```diff
@@ -62,15 +62,65 @@ for deb_tarball in "$LIVE_BUILD_OUTPUT_DIR/$APPLIANCE_VARIANT"*.debs.tar.gz; do
...
+	echo "{\"data\": \"$(base64 -w 0 SHA256SUMS)\"}" >sign-request.payload
+
+	#
+	# Here, we need to generate signature files for all of the appliance
+	# versions that'll be allowed to upgrade from, using this upgrade
+	# image. For now, we hardcode this to simply generate a signature for
+	# the "5.3" release, but eventually we'll need to update this code to
+	# support more versions (as more linux-based versions are released).
+	#
+	for signature_version in "5.3"; do
```
rasantel: Is there a reason not to port immediately to Linux the logic from `mkimage.sh` that obtains the list of all major releases by querying the app gate repo via `git ls-remote`?

prakashsurya:
@rasantel Verification of the signature will be done in the `unpack-image` script of the delphix-platform repository. I have some initial changes for that here: https://github.com/delphix/delphix-platform/pull/70 .. but that doesn't yet do the verification (as we've discussed, I need to figure out how to get the correct keys installed on the DE first).

I can change the title if you prefer me to.

scripts/build-upgrade-image.sh
prakashsurya: That's just work I haven't implemented yet, so I thought I could land this as-is and do that work afterwards. I'm also not sure if that implementation should change based on the linux build architecture changes. I want some more time to think about this, and about what the correct solution should be on linux, before porting over the illumos logic.

PR #273, opened 2019-04-22
Use a seperate dataset for "/var/log" contents

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/915/)

prakashsurya:
```
$ zfs list
rpool                            12.5G  54.8G       64K  none
rpool/ROOT                       12.5G  54.8G       64K  none
rpool/ROOT/delphix.507eJt0       12.5G  54.8G       64K  none
rpool/ROOT/delphix.507eJt0/data  23.9M  54.8G     23.9M  legacy
rpool/ROOT/delphix.507eJt0/home  6.51G  54.8G     6.51G  legacy
rpool/ROOT/delphix.507eJt0/log   3.04M  54.8G     3.04M  legacy
rpool/ROOT/delphix.507eJt0/root  5.96G  54.8G     5.96G  /
rpool/crashdump                  65.5K  17.4G     65.5K  /var/crash
rpool/grub                       3.03M  54.8G     3.03M  legacy
$ mount | grep zfs
rpool/ROOT/delphix.507eJt0/root on / type zfs (rw,relatime,xattr,posixacl)
rpool/ROOT/delphix.507eJt0/data on /var/delphix type zfs (rw,relatime,xattr,noacl)
rpool/ROOT/delphix.507eJt0/log on /var/log type zfs (rw,relatime,xattr,noacl)
rpool/ROOT/delphix.507eJt0/home on /export/home type zfs (rw,relatime,xattr,noacl)
rpool/crashdump on /var/crash type zfs (rw,xattr,noacl)
rpool/update on /var/dlpx-update type zfs (rw,xattr,noacl)
```

prakashsurya:
After "not-in-place" upgrade:
```
$ sudo zfs list -o name,mountpoint,origin
NAME                             MOUNTPOINT        ORIGIN
rpool                            none              -
rpool/ROOT                       none              -
rpool/ROOT/delphix.507eJt0       none              -
rpool/ROOT/delphix.507eJt0/data  legacy            -
rpool/ROOT/delphix.507eJt0/home  legacy            -
rpool/ROOT/delphix.507eJt0/log   legacy            -
rpool/ROOT/delphix.507eJt0/root  /                 -
rpool/ROOT/delphix.TezootU       none              -
rpool/ROOT/delphix.TezootU/data  legacy            rpool/ROOT/delphix.507eJt0/data@container-delphix.TezootU
rpool/ROOT/delphix.TezootU/home  legacy            rpool/ROOT/delphix.507eJt0/home@container-delphix.TezootU
rpool/ROOT/delphix.TezootU/log   legacy            rpool/ROOT/delphix.507eJt0/log@container-delphix.TezootU
rpool/ROOT/delphix.TezootU/root  /                 -
rpool/crashdump                  /var/crash        -
rpool/grub                       legacy            -
rpool/update                     /var/dlpx-update  -

$ mount | grep zfs
rpool/ROOT/delphix.TezootU/root on / type zfs (rw,relatime,xattr,posixacl)
rpool/ROOT/delphix.TezootU/log on /var/log type zfs (rw,relatime,xattr,noacl)
rpool/ROOT/delphix.TezootU/home on /export/home type zfs (rw,relatime,xattr,noacl)
rpool/ROOT/delphix.TezootU/data on /var/delphix type zfs (rw,relatime,xattr,noacl)
rpool/update on /var/dlpx-update type zfs (rw,xattr,noacl)
rpool/crashdump on /var/crash type zfs (rw,xattr,noacl)
```

prakashsurya:
Here we can also see the logs from before and after the upgrade (the first line is booting before the upgrade, the second line is booting after the upgrade):
```
$ sudo journalctl | grep "Command line:"
Apr 23 23:18:17 localhost.localdomain kernel: Command line: BOOT_IMAGE=/ROOT/delphix.507eJt0/root@/boot/vmlinuz-4.15.0-1035-aws root=ZFS=rpool/ROOT/delphix.507eJt0/root ro console=tty0 console=ttyS0,38400n8 ipv6.disable=1 crashkernel=1024M-:512M
Apr 24 04:01:06 ip-10-110-220-50 kernel: Command line: BOOT_IMAGE=/ROOT/delphix.TezootU/root@/boot/vmlinuz-4.15.0-1035-aws root=ZFS=rpool/ROOT/delphix.TezootU/root ro console=tty0 console=ttyS0,38400n8 ipv6.disable=1 crashkernel=1024M-:512M
```

prakashsurya:
cc: @pzakha @grwilson

pzakha:
If we want to  create a separate dataset for `/var/log`, we should make sure that we aren't breaking any packages by mounting the old `/var/log` over the new one. In particular, we should probably copy over all of the contents of the `/var/log` directory created by debootstrap into the new clone.

There are a few packages that create directories in `/var/log`. While most of them are probably not installed by debootstrap, we should still be careful when creating separate datasets for directories where dpkg can install files/folders:
```
$ dpkg -S '/var/log/*'
auditd: /var/log/audit
nginx-common: /var/log/nginx
samba-common: /var/log/samba
ntp: /var/log/ntpstats
apt: /var/log/apt
delphix-virtualization: /var/log/fluent
sysstat: /var/log/sysstat
postgresql-common: /var/log/postgresql
unattended-upgrades: /var/log/unattended-upgrades
```

Also, we'll need to create this dataset for the migration case too.

prakashsurya:
Will any of those packages be installed when `debootstrap` runs? I'm not yet sure if blindly overwriting the old contents, with the contents laid out by `debootstrap`, is correct in general. In these cases, it looks to just be a matter of creating directories, but if there were files listed here that conflicted with the old files, how should we resolve that conflict?

pzakha:
What concerns do you have over overwriting the old content?

1. So far there only seems to be directories generated by the existing packages, so for those we should just make sure they exist.

2. If a package decides to install a file somewhere in /var/log, then the file can either be a configuration file or a regular file required for the normal function of the package. Other possibilities could be files created by `postinst` scripts. In any case, if we end up overwriting a log file this way it would not be that bad as we'll have a snapshot of the previous dataset and would be able to look at the log file. If we have a configuration file or a regular file created, then we'll want to take the new versions. So overall taking the new versions seems like the right thing to do here.

prakashsurya:
Just to be clear, not all of those packages are installed when we run `deboostrap`. Only a minimal set of packages is installed at that time, enough to start the container; all others are installed after the clone is created and the container started. Here's what your command shows with only the `deboostrap` packages installed:
```
# dpkg -S '/var/log/*'
apt: /var/log/apt
```

And here's the contents of that directory (not the clone, just what's installed via `deboostrap`):
```
# ls -l /var/log
total 50
-rw-r--r-- 1 root   root              8420 Apr 24 15:51 alternatives.log
drwxr-xr-x 2 root   root                 2 Mar 11 09:34 apt
-rw-r--r-- 1 root   root             56896 Apr 24 15:51 bootstrap.log
-rw-rw---- 1 root   utmp                 0 Apr 24 15:50 btmp
drwxr-xr-x 2 root   root                 2 Apr 23 19:48 delphix-upgrade
-rw-r--r-- 1 root   root            129594 Apr 24 15:51 dpkg.log
-rw-r--r-- 1 root   root              3360 Apr 24 15:51 faillog
drwxr-sr-x 3 root   systemd-journal      3 Apr 24 16:06 journal
-rw-rw-r-- 1 root   utmp             30660 Apr 24 15:51 lastlog
-rw-r----- 1 syslog adm               3285 Apr 24 16:06 syslog
-rw------- 1 root   root              6720 Apr 24 15:51 tallylog
-rw-rw-r-- 1 root   utmp              1536 Apr 24 16:06 wtmp
```
---
> Other possibilities could be files created by postinst scripts.

I don't think files created by `postinst` hooks will show up as "owned" by the package, so we can rule out this concern.

> if we end up overwriting a log file this way it would not be that bad as we'll have a snapshot of the previous dataset and would be able to look at the log file.

We can already do this; so, if we're not going to try and "get it right" for all cases, I'd rather just rely on this mechanism always (and not use a separate log dataset).

> If we have a configuration file or a regular file created, then we'll want to take the new versions.

Only when that file hasn't been modified from what was initially installed, right? Since, for normal conffiles when there's a conflict, we keep the old version of the file, based on how we've configured `apt-gate` via the `execute` script.

If we kept the old version for normal conffiles, but the new version for files in `/var/log`, that inconsistency would be awkward.
---
Ultimately, I think there's definitely some concern to be had about how to properly preserve the old files, while also ensuring files and directories from the new packages are properly handled.

I don't think that we should blindly overwrite the "old" files with the "new" ones, unless we can verify: 1) the old file is owned by the old package, and 2) the old file hasn't been modified.. If either of these two aren't met, overwriting the old file would be inconsistent with how we deal with this for packages installed via `execute`, so I don't think that's appropriate.

I would be OK with creating directories that're owned by packages, if they don't already exist; and ignore the files (due to the complexity required to handle conflicts).

I would also be OK keeping this change as-is, since honestly I don't think this will be a problem in practice, due to how `deboostrap` is not used to install all the `delphix-entire` packages. At least for the current set of packages, it'll "just work" since we're only concerned with the `/var/log/apt` directory which will already exist in the cloned dataset.

prakashsurya:
I would also be OK investigating the feasibility of running `deboostrap` _after_ we create all the clones. This way we can rely on it to handle the creation of files/directories, and handle conflicts if they arise.

prakashsurya:
Thinking about this a little more, don't we have this same issue for the other datasets that we clone? i.e. `/var/delphix`, and `/export/home`? Shouldn't we handle this issue for `/var/log` and those other directories the same (which, right now, is to do nothing)?

pzakha:
Okay, given your investigations it does appear indeed that there are some files created by debootstrap that would overwrite some useful logs. Running debootstrap after creating the clones would be a good thing to test.

Regarding `/var/delphix` and `/export/home`, I do not believe them to be an issue since debootstrap will never install any packages in there. Even on our side, I do not believe that we have packages with file paths in there (which we should probably keep this way).

prakashsurya:
>  Even on our side, I do not believe that we have packages with file paths in there (which we should probably keep this way).

IIRC, we do install files into `/export/home`.

> Running debootstrap after creating the clones would be a good thing to test.

This appears to be working, and I think it'll alleviate your concerns and do so in a clean, simple, and maintainable way. I'll post another commit soon (finishing up testing of it now).

prakashsurya:
@pzakha I just pushed another commit, let me know what you think.

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -42,14 +42,17 @@ function create_cleanup() {
...
-function get_mounted_rootfs_name() {
-	local rootfs
+function mount_upgrade_container_dataset() {
+	[[ -n "$1" ]] || die "failed to provide dataset to mount"
+	[[ -n "$2" ]] || die "failed to provide directory for mount"
 
-	rootfs=$(zfs list -Ho name /)
-	[[ -n "$rootfs" ]] ||
-		die "failed to determine the mounted root filesystem name"
+	mkdir -p "$2" || die "failed to create directory '$2'"
```
pzakha: Not related to this review, but since we are running the upgrade scripts from the app-stack, could they have the same issue as we see here: https://jira.delphix.com/browse/LX-1852

prakashsurya: I haven't seen any problems with the delphix CLI after an upgrade.

pzakha: When you run the not-in-place upgrade test, can you double-check that the permissions for `/export` are indeed right? If so, then either we don't use the same umask in the app-stack on Linux, the umask somehow doesn't get propagated to the script, or the `/export` directory's permissions are modified somewhere else.

prakashsurya:
Which tests are we pending on?

pzakha:
Migration (blackbox-chained) run with new image would be the main one for this change.
The permissions for `/export` are not necessary for this but it would be nice to do them when we get a chance.
I'll leave the not-in-place upgrade tests up to your discretion.

upgrade/upgrade-scripts/upgrade-container
prakashsurya: I have run "not-in-place" upgrade multiple times before using the scripts, and have successfully used the delphix CLI afterwards without issue.

I can do this again, and can also do it via the delphix GUI.. but either way, I think this is unrelated to this change.

prakashsurya:
How do I run the migration tests, again? I don't remember. Do we have a git-utils for that yet?

pzakha:
We do not have git-utils for this yet.
The steps are:
1. Generate an ESX internal-qa image
2. Pass this image to a blackbox-chained job (see Configuration A for https://docs.google.com/document/d/1zQ0A5di6eOwpDrGIQCppgtdiXkm9vR8NZPwt-G16doU)

upgrade/upgrade-scripts/upgrade-container
prakashsurya: After a not-in-place upgrade via the GUI:
```
$ ls -alhd /export/ /export/home/ /export/home/cli/
drwxr-xr-x 3 root root  3 Apr 24 20:24 /export/
drwxr-xr-x 4 root root  4 Apr 23 22:51 /export/home/
drwxr-xr-x 3 cli  staff 7 Apr 23 22:51 /export/home/cli/
```
I'm able to use the CLI too:
```
± ssh admin@ps-upgrade.dlpxdc.co
Password: 
ip-10-110-208-210> ls
Children
about
action
...
```

prakashsurya: You can do the same yourself with the following:
```
$ download-latest-image internal-dev
$ sudo unpack-image -x internal-dev.upgrade.tar.gz
```
Then go into the GUI and run the upgrade as normal.

pzakha: That's interesting, thanks for testing this. I'll need to see how come this is different from the migration case.

PR #274, opened 2019-04-22
Migrate two more directories on upgrade

prakashsurya:
This change adds the following directories to be migrated on upgrade:

 * /var/lib/nfs
 * /var/target/pr

### DIFF
--- a/upgrade/upgrade-scripts/upgrade-container
+++ b/upgrade/upgrade-scripts/upgrade-container
@@ -468,17 +468,39 @@ function migrate_file() {
 		parentdir="$(dirname "$path")"
 		mkdir -p "/var/lib/machines/${CONTAINER}${parentdir}" ||
 			die "'mkdir -p $parentdir' failed for '$CONTAINER'"
-		cp "$path" "/var/lib/machines/${CONTAINER}${path}" ||
-			die "'cp $path' failed for '$CONTAINER'"
+		cp -p "$path" "/var/lib/machines/${CONTAINER}${path}" ||
+			die "'cp -p $path' failed for '$CONTAINER'"
 	else
-		if [[ -f "/var/lib/machines/${CONTAINER}${path}" ]]; then
-			# If the target file doesn't exist in the host then
-			# we delete the file. This case could happen if the host
-			# configuration doesn't exist but a fresh install placed
-			# package default configuration.
-			rm "/var/lib/machines/${CONTAINER}${path}" ||
-				die "failed to remove '$path' in '$CONTAINER'"
-		fi
+		#
+		# If the target file doesn't exist in the host then we
+		# delete the file. This case could happen if the host
+		# configuration doesn't exist but a fresh install placed
+		# package default configuration.
+		#
+		rm -f "/var/lib/machines/${CONTAINER}${path}" ||
+			die "failed to remove file '$path' in '$CONTAINER'"
+	fi
+}
+
+function migrate_dir() {
+	local path="$1"
+	local parentdir
+
+	if [[ -d "$path" ]]; then
+		parentdir="$(dirname "$path")"
+		mkdir -p "/var/lib/machines/${CONTAINER}${parentdir}" ||
+			die "'mkdir -p $parentdir' failed for '$CONTAINER'"
+		cp -prT "$path" "/var/lib/machines/${CONTAINER}${path}" ||
+			die "'cp -prT $path' failed for '$CONTAINER'"
+	else
+		#
+		# If the target dir doesn't exist in the host then
+		# we delete the dir. This case could happen if the host
+		# configuration doesn't exist but a fresh install placed
+		# package default configuration.
+		#
+		rm -rf "/var/lib/machines/${CONTAINER}${path}" ||
+			die "failed to remove directory '$path' in '$CONTAINER'"
 	fi
 }
 
@@ -529,7 +551,7 @@ function migrate_configuration() {
 	EOF
 
 	#
-	# These files are specific to the delphix virtualization and/or
+	# These paths are specific to the delphix virtualization and/or
 	# masking application, and must be preserved across upgrades.
 	# Ideally this list would be provided by the applications
 	# themselves, rather than specified here, but the infrastructure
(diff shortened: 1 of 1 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/916/)

prakashsurya:
Created some files in `/var/target/pr` (manually with `dd`) and `/var/lib/nfs` (by creating VDBs), and verified the files remained after the upgrade.

Before upgrade:
```
$ find /var/lib/nfs /var/target/pr -type f | xargs sudo sha256sum > SHA256SUMS
$ cat SHA256SUMS 
9200f2e2a214a948ad30e6039f96116354b74fb6a5953621420e0ffec60cd3f2  /var/lib/nfs/etab
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855  /var/lib/nfs/export-lock
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855  /var/lib/nfs/.xtab.lock
9d9f290527a6be626a8f5985b26e19b237b44872b03631811df4416fc1713178  /var/lib/nfs/state
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855  /var/lib/nfs/rmtab
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855  /var/lib/nfs/xtab
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855  /var/lib/nfs/.etab.lock
c1ef599f98dc5bbcf517fd1d4ef52764a9dca4ab4257d619adb7fe7fb774aca1  /var/target/pr/f2
af68d9ee3de20bd99aa83b297e98ba55180ded1dbfbe3dc7c61c3fa5b36b9c29  /var/target/pr/README
11c20d9aad1309eef71cf8de43ed1b8370648c651d0fce88cb6bddca69bb1f17  /var/target/pr/f1
3c44855d68b27c10f6c8fab77b0b75197df6cd47e6a5f6303197c73714f6154c  /var/target/pr/d2/f1
039dd2b8eb3df7f7a196662707221b4800b5371c4285f97abd509cf4fc0f9856  /var/target/pr/d2/f2
0c944fd432f25d6507e483caea3532e8cc1c6644d8a174c1b8e6564431e1dd8f  /var/target/pr/d1/f1
35ae1093c731a99eb0546638aad439c80c1c9fab1cc62551d86fce9200f87f5c  /var/target/pr/d1/f2
```

Do upgrade:
```
$ sudo /var/dlpx-update/latest/upgrade -v not-in-place
```

After upgrade:
```
$ sudo sha256sum -c SHA256SUMS 
/var/lib/nfs/etab: OK
/var/lib/nfs/export-lock: OK
/var/lib/nfs/.xtab.lock: OK
/var/lib/nfs/state: OK
/var/lib/nfs/rmtab: OK
/var/lib/nfs/xtab: OK
/var/lib/nfs/.etab.lock: OK
/var/target/pr/f2: OK
/var/target/pr/README: OK
/var/target/pr/f1: OK
/var/target/pr/d2/f1: OK
/var/target/pr/d2/f2: OK
/var/target/pr/d1/f1: OK
/var/target/pr/d1/f2: OK
```

prakashsurya:
cc: @pzakha @grwilson

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -429,14 +429,36 @@ function migrate_file() {
...
-			rm "/var/lib/machines/${CONTAINER}${path}" ||
-				die "failed to remove '$path' in '$CONTAINER'"
-		fi
+		#
+		# If the target file doesn't exist in the host then we
+		# delete the file. This case could happen if the host
+		# configuration doesn't exist but a fresh install placed
+		# package default configuration.
+		#
+		rm -f "/var/lib/machines/${CONTAINER}${path}" ||
```
grwilson: NIT: `rm -f` never fails so you don't need the `die`

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -429,14 +429,36 @@ function migrate_file() {
...
+
+function migrate_dir() {
+	local path="$1"
+	local parentdir
+
+	if [[ -d "$path" ]]; then
+		parentdir="$(dirname "$path")"
+		mkdir -p "/var/lib/machines/${CONTAINER}${parentdir}" ||
+			die "'mkdir -p $parentdir' failed for '$CONTAINER'"
+		cp -rT "$path" "/var/lib/machines/${CONTAINER}${path}" ||
```
grwilson: Do we need to preserve the attributes for these files?

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -429,14 +429,36 @@ function migrate_file() {
...
+			die "'cp -rT $path' failed for '$CONTAINER'"
+	else
+		#
+		# If the target dir doesn't exist in the host then
+		# we delete the dir. This case could happen if the host
+		# configuration doesn't exist but a fresh install placed
+		# package default configuration.
+		#
+		rm -rf "/var/lib/machines/${CONTAINER}${path}" ||
+			die "failed to remove directory '$path' in '$CONTAINER'"
```
grwilson: Same nit about `rm -f`

upgrade/upgrade-scripts/upgrade-container
prakashsurya: It won't fail if the file doesn't exist, but presumably the command execution could potentially fail for other reasons (signal, etc.)?

upgrade/upgrade-scripts/upgrade-container
prakashsurya: I don't know. We're not currently doing it when migrating files, so I didn't bother doing it for these directories.

I'd be OK adding the `-p` option (or w/e the option is) to preserver the attributes if you think that'd be the safest bet; I'm not really sure either way, and not sure how to verify if it's needed or not.

upgrade/upgrade-scripts/upgrade-container
pzakha: > It won't fail if the file doesn't exist, but presumably the command execution could potentially fail for other reasons (signal, etc.)?

That's right, it would fail if the fail exists and the removal fails, which is something we'd want to catch.

PR #276, opened 2019-04-24
LX-1852 migration: Failure to access CLI after migration

pzakha:
This makes sure that any directory created by the migration script has the default umask.

See https://jira.delphix.com/browse/LX-1852 for more details.

## Testing

build image: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/929/
Run migration and perform some checks:
- http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/blackbox-chained/104/ (failure of the job is expected)
- Make sure cli works
- Verify that `/export` has the right permissions set.

### DIFF
--- a/live-build/misc/migration-scripts/dx_apply
+++ b/live-build/misc/migration-scripts/dx_apply
@@ -24,6 +24,15 @@
 
 set -o pipefail
 
+#
+# Reset the umask to the default value. When called from the app-stack the
+# umask is set to 0027. Since those scripts can be called manually, we want
+# to have a consistent result regardless of the caller. A umask of 0022
+# makes directories created by this script accessible by everyone by default,
+# which is important for directories such as /export/home.
+#
+umask 0022
+
 function die() {
 	echo "$(basename "$0"): $*" >&2
 	exit 1
--- a/live-build/misc/migration-scripts/dx_execute
+++ b/live-build/misc/migration-scripts/dx_execute
@@ -16,9 +16,19 @@
 #
 
 #
-# Sets up the FreeBSD bootloader to boot into that Linux image next time
-# we reboot.
+# Finalizes the Linux image in preparation for the reboot now that the app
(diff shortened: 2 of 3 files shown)


pzakha:
cc: @prakashsurya @sebroy @sdimitro @shartse

live-build/misc/migration-scripts/dx_apply
```diff
@@ -24,6 +24,13 @@
 
 set -o pipefail
 
+#
+# Reset the umask to the default value. When called from the app-stack the
+# umask is set to 0027. Since those scripts can be called manually, we want
+# to have a consistent result regardless of the caller.
+#
+umask 0022
```
sdimitro: Feel free to drop this if you think it is unnecessary but I was wondering if we should add
a comment explaining why this matter (e.g. we are creating directories that will need to
be accessed by multiple users in linux and we need the appropriate permissions, like `/export/home`).

prakashsurya: I agree, and think it's important. Otherwise, I'd be inclined to ask, why not change the umask for the default shell to be `0027`, to match the application?

An alternative solution to this would be, we could add some verification of the permissions for certain directories to the delphix-platform ansible logic. This would be more explicit and allow us to comment on why we need the specific permissions that are being set via ansible; was this considered?

I'm fine with this, just trying to weigh the different available options.

pzakha: I've put `0022` because that's the usual behaviour that people are expecting (i.e. the ability to traverse directories for `others`). I think that if some files need to have their access restricted, I'd rather see it done explicitly.

I did look at `delphix-platform` initially, however since I wanted to make this change for consistency anyway I thought that a delphix-platform change wasn't required anymore.

prakashsurya:
bors delegate+

PR #278, opened 2019-04-29
Use "/etc/environment" in "verify-stack" script

prakashsurya:


prakashsurya:
cc: @sravyamks @pandar00

prakashsurya:
This doesn't actually work; we'd still need to explicitly `export` all variables found in the `/etc/environment` file.

PR #284, opened 2019-05-03
verify-jar : add support for 4 new long parameters for initialization timestamps

andycwood:
This is a part of a larger change to upgrade verify UX. The user can now view the verification steps.
Some PRE-verify steps are included in the report so that it is relatively in-sync with the messaging in the action pane, and the time spent waiting for verify to complete.
This change is a dependency for this app-gate commit:
http://reviews.delphix.com/r/48499/

### DIFF
--- a/upgrade/upgrade-scripts/verify-jar
+++ b/upgrade/upgrade-scripts/verify-jar
@@ -15,16 +15,11 @@
 # limitations under the License.
 #
 
+# shellcheck disable=SC1090
 . "${BASH_SOURCE%/*}/common.sh"
 
 IMAGE_VERSION=$(get_image_version)
 [[ -n "$IMAGE_VERSION" ]] || die "failed to determine image version"
-
-function usage() {
-	echo "$(basename "$0"): $*" >&2
-	echo "Usage: $(basename "$0") [-d] [-f format] [-l locale] [-o path]"
-	exit 2
-}
 
 function verify_jar_verify_cleanup() {
 	local rc="$?"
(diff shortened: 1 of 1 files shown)


prakashsurya:
Can you comment on why we have to pass these time stamps into the verify JAR? Why does the verify JAR need these timestamps? Why can't we have the app-gate consume these timestamps outside of the verify JAR? If this is covered elsewhere, feel free to point me to some existing docs/explanation.

My concern is with increasing the interface/API of the upgrade scripts, which can be problematic if we ever need to change this interface, since the interface needs to be backwards compatible with older versions of the product.

upgrade/upgrade-scripts/verify-jar
```diff
@@ -54,20 +55,34 @@ function verify_jar_verify_cleanup() {
 		report_progress 100 "Upgrade verification was successful"
 	return "$rc"
 }
+#
+# use command line program because it supports long arguments.
+#
+TEMP=`getopt -a -o :df:l:o: --long fsCheckStartTs:,fsCheckEndTs:,containerInitStartTs:,containerInitEndTs: -n 'verify-jar' -- "$@"`
```
prakashsurya: I took a stab and cleaning up the option handling and usage printing code; see [here](https://github.com/prakashsurya/appliance-build/commit/70c4719ca60b67d98150d321cdee972cf3d5cad8)

IMO, with my changes, it's more readable and maintainable; let me know what you think. If you like it, you can easily pull it into your local tree with the following:
```
$ curl -s https://github.com/prakashsurya/appliance-build/commit/70c4719ca60b67d98150d321cdee972cf3d5cad8.patch | git am -
```
Then use the usual tools to squash it to a single commit, reset the author, etc.

andycwood:
> Can you comment on why we have to pass these time stamps into the verify JAR? Why does the verify JAR need these timestamps? Why can't we have the app-gate consume these timestamps outside of the verify JAR? If this is covered elsewhere, feel free to point me to some existing docs/explanation.
> 
> My concern is with increasing the interface/API of the upgrade scripts, which can be problematic if we ever need to change this interface, since the interface needs to be backwards compatible with older versions of the product.

The review has some attached files with the new json output of the verify jar
The upgrade verify jar performs localization of the raw report output. 
So, with the current role of verify jar, it makes sense to provide the raw data to the verify jar so that it can add a localized message to the data. Here is how the relevant sections of the report look:
{
"runStatus": "SUCCESS",
"className": "FileSystemCheckRunReport",
"description": "Verify integrity of internal file systems",
"startTimestamp": "2019-05-02T16:56:16.289+0000",
"endTimestamp": "2019-05-02T16:56:16.434+0000",
"duration": "00:00:00",
"runStatusMessage": null,
"notifications": []
},
{
"runStatus": "SUCCESS",
"className": "InitializationRunReport",
"description": "Initialize verification",
"startTimestamp": "2019-05-02T16:56:16.511+0000",
"endTimestamp": "2019-05-02T17:00:13.295+0000",
"duration": "00:03:56",
"runStatusMessage": null,
"notifications": []
}

upgrade/upgrade-scripts/verify-jar
```diff
@@ -55,48 +50,147 @@ function verify_jar_verify_cleanup() {
...
+
+GETOPT=$(getopt --alternative \
+	--name "$(basename "$0")" \
+	--options "$(printf "%s," "${OPTIONS_SHORT[@]}")" \
+	--longoptions "$(printf "%s," "${OPTIONS_LONG[@]}")" \
+	-- "$@") || usage "$@"
+eval set -- "$GETOPT"
+
+#
+# Add support for optional timestamp parameters from managment stack. These timestamps
```
prakashsurya: IMO, this comment isn't very useful.. but if you think it is, perhaps we can at least move it to be closer to the options it's commenting on? e.g. move it directly above the `--container-InitStartTs`?

upgrade/upgrade-scripts/verify-jar
```diff
@@ -55,48 +50,147 @@ function verify_jar_verify_cleanup() {
...
+		"-fsCheckStartTs" "$opt_fsStartTs"
+		"-fsCheckEndTs" "$opt_fsEndTs"
+	)
+fi
+
+if [[ -n "$opt_initStartTs" ]] && [[ -n "$opt_initEndTs" ]]; then
+	VERIFY_OPTIONS=(
+		"${VERIFY_OPTIONS[@]}"
+		"-containerInitStartTs" "$opt_initStartTs"
+		"-containerInitEndTs" "$opt_initEndTs")
```
prakashsurya: perhaps move the closing `)` to the next line, to be consistent with what you did in lines above?

upgrade/upgrade-scripts/verify-jar
```diff
@@ -55,48 +50,147 @@ function verify_jar_verify_cleanup() {
 	return "$rc"
 }
 
+#
+# Use formatted usage
```
prakashsurya: IMO this comment isn't helpful and can be removed.

prakashsurya:
Please also rebase/squash this to a single commit before landing, and make sure you get a successful run of `git-ab-pre-push`.

andycwood:
This migration job passed, so I am going to start the merge 
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/blackbox-chained/137/console
bors r+

Issue #100, opened 2018-09-06
81-upgrade-repository.binary can fail when trying to download packages


prakashsurya:
Currently, when running the `81-upgrade-repository.binary` script, it can fail when running `apt-get download` as seen in the snippet below:
```
14:34:39 + xargs apt-get download
14:34:39 + dpkg-query -Wf '${Package}=${Version}\n'
14:34:39 E: Can't find a source to download version '2.02-2ubuntu8.3' of 'grub-pc-bin:amd64'
14:34:39 E: Can't find a source to download version '2.02-2ubuntu8.3' of 'grub-pc:amd64'
14:34:39 E: Can't find a source to download version '2.02-2ubuntu8.3' of 'grub2-common:amd64'
14:34:39 E: Can't find a source to download version '2.02-2ubuntu8.3' of 'grub-common:amd64'
14:34:39 E: config/hooks/81-upgrade-repository.binary failed (exit non-zero). You should check for errors.
```

The problem lies in the fact that `apt-get download` can only download the latest version of the package.

In the failure seen above, we're attempting to download version `2.02-2ubuntu8.3` of the grub packages, but version `2.02-2ubuntu8.4` is available in Ubuntu's repositories. This fails because `apt-get download` is not capable of downloading the "old" version of the packages; it works fine if we had instead attempted to download the "new" version `2.02-2ubuntu8.4` of these same packages.

For example:
```
$ dpkg-query -Wf '${Package}=${Version}\n' | grep grub
grub-common=2.02-2ubuntu8.3
grub-gfxpayload-lists=0.7
grub-pc=2.02-2ubuntu8.3
grub-pc-bin=2.02-2ubuntu8.3
grub2-common=2.02-2ubuntu8.3

$ dpkg-query -Wf '${Package}=${Version}\n' | grep grub | xargs apt-get download
E: Can't find a source to download version '2.02-2ubuntu8.3' of 'grub-pc-bin:amd64'
E: Can't find a source to download version '2.02-2ubuntu8.3' of 'grub-pc:amd64'
E: Can't find a source to download version '2.02-2ubuntu8.3' of 'grub2-common:amd64'
E: Can't find a source to download version '2.02-2ubuntu8.3' of 'grub-common:amd64'

$ apt-cache show grub-common | grep Version
Version: 2.02-2ubuntu8.4
Version: 2.02-2ubuntu8.3
Version: 2.02-2ubuntu8

$ sudo apt-get dist-upgrade -y
[...]

PR #292, opened 2019-05-20
Add new mypy version to crash-python dev role

pcd1193182:
We install the pip apt package and then use it to install mypy.  Tested by building an internal dev kvm image and verifying that mypy is installed and runnable:
```
delphix@localhost:~$ mypy --version
mypy 0.701
```

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.crash-python-development/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.crash-python-development/tasks/main.yml
@@ -19,6 +19,15 @@
     name:
       - git
       - python3-pyelftools
+      - python3-pip
+    state: present
+  register: result
+  until: result is not failed
+  retries: 3
+  delay: 60
+
+- pip:
+    name: mypy
     state: present
   register: result
   until: result is not failed


live-build/misc/ansible-roles/appliance-build.crash-python-development/tasks/main.yml
```diff
@@ -19,6 +19,15 @@
...
       - git
       - python3-pyelftools
+      - python3-pip
+    state: present
+  register: result
+  until: result is not failed
+  retries: 3
+  delay: 60
+
+- pip:
```
prakashsurya: Do we need to ensure we use “pip3”? Can you verify this will install for python3 on the image you tested?

pcd1193182: It uses pip3 by default because we set `ansible_python_interpreter` to python3. I have verified that the mypy installed is based on python3.

prakashsurya:
```
00:33:41.474 + dpkg-query -Wf '${Package}=${Version}\n'
00:33:41.474 + xargs apt-get download
00:33:41.792 E: Can't find a source to download version '0.96.24.32.7' of 'python3-software-properties:amd64'
00:33:41.792 E: Can't find a source to download version '0.96.24.32.7' of 'software-properties-common:amd64'
00:33:41.792 E: config/hooks/82-upgrade-repository.binary failed (exit non-zero). You should check for errors.
```
failure appears to be unrelated and most likely #100

bors r+

pcd1193182:
another environmental failure

bors r+

PR #298, opened 2019-05-28
crashdump dataset not mounted when generating crash dumps

grwilson:


live-build/config/hooks/vm-artifacts/90-raw-disk-image.binary
```diff
@@ -197,12 +202,14 @@ rsync --info=stats3 -Wa binary/* "$DIRECTORY/"
 # We rely on the "/etc/fstab" file to mount the non-root ZFS
 # filesystems, so that when a specific rootfs dataset is booted, it'll
 # automatically mount the correct non-rootfs datasets associated with
-# that rootfs dataset.
+# that rootfs dataset. In addition, we need to mount /var/crash before the
```
prakashsurya: While we're here describing the "/var/crash" directory, it might also be a good place to comment on the fact that we need to use "/etc/fstab" for that mountpoint so it gets properly mounted after a kernel crash, and the dump is generated in that dataset, as opposed to the root dataset. If not here, where we configure this dataset to be a "legacy" mount might be another good spot to document this requirement.

grwilson: Fixed.

pzakha:
This should also be added here for migration: https://github.com/delphix/appliance-build/blob/master/live-build/misc/migration-scripts/dx_apply#L187

Please add this and test that migration works.

prakashsurya:
The upgrade scripts will also need to be updated.

grwilson:
I posted a new [PR](https://github.com/delphix/appliance-build/pull/300) for those changes

PR #301, opened 2019-05-31
DLPX-64338 Remove conditions from bash scripts

dtzeng:
See app-gate review here: http://reviews.delphix.com/r/49763/

dx_manage_pg will still be used on illumos when running migration. To prevent the need of trunk's dx_manage_pg still branching on is_linux calls, I copied an illumos-only version of dx_manage_pg here, which will only be used in the migration image.

As such, the trunk version of dx_manage_pg in app-gate can be cleaned up and not need to handle the illumos case anymore.

I decided to ignore fmt/shellcheck for dx_manage_pg because there were a lot of errors, and there's not much point to put a lot of effort into fixing the styles since this script is only going to be used for one release for migration.

### DIFF
--- a/live-build/config/hooks/vm-artifacts/90-linux-migration-artifact.binary
+++ b/live-build/config/hooks/vm-artifacts/90-linux-migration-artifact.binary
@@ -76,6 +76,10 @@ mkdir $DEPOT_DIRECTORY
 #
 (
 	cd binary
+
+	# Copy an illumos version of dx_manage_pg for migration purposes.
+	cp ../migration-scripts/dx_manage_pg_illumos opt/delphix/server/bin/
+
 	find . -print | cpio -oc
 ) >$DEPOT_DIRECTORY/os-root.cpio
 
@@ -84,6 +88,7 @@ mkdir $DEPOT_DIRECTORY
 # dx_unpack.
 #
 cp migration-scripts/* $DEPOT_DIRECTORY
+rm $DEPOT_DIRECTORY/dx_manage_pg_illumos
 
 #
 # There may be a version.info file in the current directory already.
--- a/build.gradle
+++ b/build.gradle
@@ -79,7 +79,9 @@ def shellScripts = fileTree("scripts") +
                    fileTree("live-build/config/hooks").include({ details ->
                         details.file.canExecute()
                    }) +
-                   fileTree("live-build/misc/migration-scripts") +
+                   fileTree("live-build/misc/migration-scripts", {
+                       exclude "dx_manage_pg_illumos"
+                   }) +
                    fileTree("upgrade/upgrade-scripts", {
                        exclude "README.md"
                    })
--- a/live-build/misc/migration-scripts/dx_manage_pg_illumos
+++ b/live-build/misc/migration-scripts/dx_manage_pg_illumos
@@ -1,0 +1,356 @@
+#!/bin/bash
+#
+# Copyright (c) 2015, 2019 by Delphix. All rights reserved.
+#
+
+#
+# Provides management of snapshots and clones of mds, ability to start up a
+# postgres instance on an empty port, and cleaning up after.
+# The postgres SVC name will always be:
+#     svc:/system/delphix/postgres:$snapshot_name
+#
+
+. ${BASH_SOURCE%/*}/../lib/common.sh
+
+DEFAULT_START_TIMEOUT=2100
+DEFAULT_RESTARTER_TIMEOUT=1200
+DEFAULT_QUERY_TIMEOUT=900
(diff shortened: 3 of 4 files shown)


dtzeng:
@prakashsurya @jgallag88 @pandar00

live-build/config/hooks/vm-artifacts/90-linux-migration-artifact.binary
```diff
@@ -76,6 +76,13 @@ mkdir $DEPOT_DIRECTORY
 #
 (
 	cd binary
+
+	#
+	# Replace dx_manage_pg with the illumos version, since dx_manage_pg
+	# will be used when still on illumos during migration.
+	#
+	cp ../migration-scripts/dx_manage_pg opt/delphix/server/bin/dx_manage_pg
```
jgallag88: Won't replacing `dx_manage_pg` be problematic after we reboot and we are running Linux? I think this script is used for more than just upgrade verification.

dtzeng: Oh yea... you are right. I was assuming it was only ever used for upgrade/migration purposes.
Looks like it is also used for support bundle generation.

I think I can change this to just leaving the illumos version of `dx_manage_pg` in the DEPOT, and then `dx_verify` will invoke that dx_manage_pg directly instead of the one in the dataset.

How does that sound?

dtzeng: Actually, that doesn't quite work either... The upgrade verify jar will use dx_manage_pg from the dataset

jgallag88: Doesn't `dx_manage_pg` get invoked from `UpgradeVerify.java`? But yeah, I think having both copies around should do it. We could even leave the Illumos copy in the app-gate, but we wouldn't have to update it at all, and could delete it after 6.0 branches.

dtzeng: I've updated the code to copy a new `dx_manage_pg_illumos` so we can have both the illumos version for migration and linux version to use after we successfully migrate onto Linux.

The UpgradeVerify jar has also been updated in the app-gate review

jgallag88:
This looks fine to me, but I wonder if it would be simpler to put `dx_manage_pg_illumos` in the app-gate. That way the script would be present on migrated and non-migrated engines alike. Even though we don't need it apart from migration, it would be nice to keep the two types of installations identical if possible.

Also, in the unlikely event that we do need to modify  `dx_manage_pg_illumos`, it would probably be as part of an app-gate change, so putting it in the app-gate would help avoid a flag day.

dtzeng:
@jgallag88 I was thinking it was easier to keep as much of the migration code together in appliance-build, now that we are trying to remove all illumos logic from app-gate. But if you feel strongly about this, I can move `dx_manage_pg_illumos` back to app gate.

live-build/config/hooks/vm-artifacts/90-linux-migration-artifact.binary
```diff
@@ -76,6 +76,10 @@ mkdir $DEPOT_DIRECTORY
 #
 (
 	cd binary
+
+	# Copy an illumos version of dx_manage_pg for migration purposes.
+	cp ../migration-scripts/dx_manage_pg_illumos opt/delphix/server/bin/
```
prakashsurya: Will this script only be used for migration? If so, why do we need it installed in the root filesystem, rather than the migration process using the script directly from the upgrade directory (like the "dx" scripts)?

If we _need_ the file to be installed in the root directory like this, I agree with John, that we should install the file via some package (i.e. the app-gate). If we don't need it to be installed, I'd be OK leaving it in the "migration-scripts" directory of this repository, and then having consumers execute it directly from the unpacked directory in "/var/dlpx-update".

prakashsurya: It might be best to just keep current `dx_manage_pg` that works for both linux and illumos, due to the requirements of migration. I feel like removing the illumos-specific code from the app-gate script, only to have it duplicated here, makes things more complicated for little benefit.

jgallag88:
My feeling is that since this is basically app-gate code and interacts primarily with other app-gate code, it should live in the app-gate. I'm also OK with Prakash's suggestion that we just leave the bits of Illumos code that we need for migration until after 6.0 branches, and clean it up then.

dtzeng:
Makes sense, I'll close this out, and update the app-gate review to keep the illumos parts needed for dx_manage_pg to work for migration

PR #305, opened 2019-06-03
Remove check for dropbox in migration dx_apply

shartse:
After https://github.com/delphix/delphix-platform/pull/93, we no longer expect to see the `dropbox` directory in the migration image. Remove the relevant check in `dx_apply`

Appliance build: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1260/

Chained migration: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/blackbox-chained/249/

### DIFF
--- a/live-build/misc/migration-scripts/dx_apply
+++ b/live-build/misc/migration-scripts/dx_apply
@@ -321,12 +321,9 @@ MAIN_MENU_FICL=(
 #
 (
 	cd "$TMP_ROOT/var/delphix" || die "failed to cd into $TMP_ROOT/var/delphix"
-	# We only expect to find:
-	#  ./dropbox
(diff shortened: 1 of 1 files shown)


shartse:
cc @pzakha @prakashsurya

shartse:
The chained migration job lost connection with the host timed out - I haven't been able to log in to see what happened. However, I was able to use the migration image from the build to do a successful manual upgrade, and got another successful chained migration run: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/blackbox-chained/250/, so I'll merge it. My reasoning is that the impacts are limited to migration, which is already broken.

PR #308, opened 2019-06-08
DLPX-64339 Remove/refactor Illumos specific scripts/code blocks

dtzeng:
dxsvcinit has been silently failing to take a snapshot on migration because we do not set the base version in the migration case
```
Jun 07 08:22:52 dlpx-5350-QAR-563f3ba9.dc4 systemd[1]: Starting Delphix default PostgreSQL database server...
Jun 07 08:22:52 dlpx-5350-QAR-563f3ba9.dc4 svc-postgres[5819]: dxsvcinit: missing UPGRADE_BASE_VERSION peroperty
```
This bug was exposed when working on this app-gate change: http://reviews.delphix.com/r/49923
Testing on app-gate review


prakashsurya:
bors delegate+

PR #310, opened 2019-06-11
Don't use Artifactory's APT repository

prakashsurya:
Previously, we used the Artifactory APT repository to fetch the
"crypt-blowfish" packages. That package is now provided by "linux-pkg",
so we can remove our dependency on Artifactory.


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1340/)

Note: This depends on the linux-pkg PR here: https://github.com/delphix/linux-pkg/pull/38

prakashsurya:
Actually, I'm not sure I can push this as-is.. @basil don't we use Artifactory for storing the java package?

It's hard to tell, but the ab-pre-push run with my linux-pkg changes failed, presumably because it relies on Artifactory for the java package (I wish it would tell me which package it couldn't find).

basil:
> Actually, I'm not sure I can push this as-is.. @basil don't we use Artifactory for storing the java package?

The Java package is stored in `http://artifactory.delphix.com/artifactory/adoptopenjdk/`, not `http://artifactory.delphix.com/artifactory/appliance-packages`. As far as I know, only the following files are currently stored in `http://artifactory.delphix.com/artifactory/appliance-packages`:

```
libcrypt-blowfish-dev_1.3_all.deb
libcrypt-blowfish_1.3_amd64.deb
pam-challenge-response_1.0_amd64.deb
```

prakashsurya:
Ah, ok.. I'll look into the failure some more than.. it might be due to that `pam-challenge-response` package.. not sure what that is yet.

basil:
> it might be due to that `pam-challenge-response` package.. not sure what that is yet.

Raul would be the one to ask. It was added [here](http://reviews.delphix.com/r/43525/) and [here](http://reviews.delphix.com/r/43543/) and seems to live [here](https://gitlab.delphix.com/os-platform/linux-dlpx-pkgs/tree/master/challenge_response).

prakashsurya:
Thanks.. based on that, I opened https://github.com/delphix/linux-pkg/pull/40 .. I'll retry this, with that change, and see if that'll resolve the issue.

prakashsurya:
The "challenge-response" and "crypt-blowfish" packages have now been integrated into linux-pkg, and I've verified with my pre-push testing that this PR works properly with those changes.

I'm going to wait until the "linux-pkg" post-push Jenkins job runs, to update the "latest" S3 bucket to contain those new packages, then I'll land this PR. I plan to land without using "bors" so we can skip the build latency, and reduce the time that appliance-build "master" is broken (since without this change, and once the "linux-pkg" post push job completes, appliance-build will not work).

PR #315, opened 2019-06-17
Don't use the PostgreSQL APT repository

prakashsurya:


prakashsurya:
- `git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1401/)
- `git-ab-pre-push --test-upgrade-from 5.3.5.0` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1402/)

basil:
bors delegate+

basil:
Change looks good to me, but I'm curious as to the motivation.

prakashsurya:
> Change looks good to me, but I'm curious as to the motivation.

George was thinking about how security updates will work, when using the package supplied by the postgres repository.. I opened this not because I think it's ready, but to see if testing would "just work".. but if it does work, then it might be a good change, for the following reasons:

1. It's one less repository we'd have to mirror/update/etc. internally
2. We'd get updates from Ubuntu, based on there policy for security patches, etc.

PR #326, opened 2019-07-01
Prevent SSH idle timeouts on development systems

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1535/)

prakashsurya:
cc: @grwilson

prakashsurya:
This approach won't work, since the platform service will revert this change on first boot. We'll have to figure out a way to inject the variant name into the platform service, similar to how we have information about the platform, and then conditionally apply the SSH configuration based on the variant name.

PR #328, opened 2019-07-10
CP-2777 CRA enabled for Linux

rasantel:
**Description**
Challenge-response authentication for the OS support user (`delphix`) must be enabled in 6.0, both for fresh new engines and migrations from 5.3, for external engines. For internal engines, we leave CRA disabled for convenience.

There is a new role for enabling CRA, which is invoked only for the external variants.

Also, we ensure that `/etc/engine-uuid`, needed by CRA on Linux, exists. This file is not present on illumos, so without this change in `minimal-common`, CRA would not work for engines migrated from 5.3.

**Testing**
Verified using `appliance-build-pre-push` that internal engines do not get CRA enabled, and external engines do have it enabled (both fresh and migrated engines).

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.virtualization-external/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.virtualization-external/tasks/main.yml
@@ -21,3 +21,4 @@
     line: "{{ item.key }}={{ item.value }}"
   with_items:
     - { key: 'cr_auth', value: 'true' }
+- command: pam-auth-update --enable challenge-response


live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -76,3 +76,8 @@
   delay: 30
   register: result
   until: result is succeeded
+
+#
+# CRA on Linux requires that the engine UUID is stored in a file
+#
+- shell: dmidecode -s system-uuid | awk '{print tolower($0)}' > /etc/engine-uuid
```
pzakha: This would not work as intended as it will be executed on the build system and so will hardcode that engine-uuid into the install and upgrade image. For migration we would need `/etc/engine-uuid` to be generated dynamically on the customer's system.

sebroy: I assume this requirement exists so that a given set of private keying material remains valid and verifiable even if the underlying `system-uuid` changes?

Are there any valid customer use cases that result in a change in `system-uuid`? For example, some customers have been known to create multiple engines by cloning a pre-setup/pre-registration image. If the `/etc/engine-uuid` file is identical in all of those clones, then I won't that be problematic?

jgallag88: Does `/etc/engine-uuid` need to be different for each engine? If we create the file here, I think it will be the same for every engine created from this build. Also, do we need the engine UUID to match system-uuid, or could we generate something really random using something like `cat /proc/sys/kernel/random/uuid` or `dd if=/dev/urandom bs=16 count=1 2>/dev/null | xxd -p | sed -E 's/(.{8})(.{4})(.{4})(.{4})(.{12})/\1-\2-\3-\4-\5/'`?

jgallag88:
In the interest of keeping our testing as realistic as possible, I would feel significantly more comfortable if we enabled CRA on the internal-qa variant as well. Any difference between the external variants and the internal-qa variant is an opportunity for bugs to go uncaught by our testing, and CRA seems like a pretty important feature to have minimal automated testing for.

That said, I realize that enabling CRA on internal-qa would require significant work from qa and devops, which would require a larger discussion that shouldn't necessarily block this PR.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
jgallag88: Ah I see Seb and Pavel beat me to it.

pzakha:
> In the interest of keeping our testing as realistic as possible, I would feel significantly more comfortable if we enabled CRA on the internal-qa variant as well. Any difference between the external variants and the internal-qa variant is an opportunity for bugs to go uncaught by our testing, and CRA seems like a pretty important feature to have minimal automated testing for.
> 
> That said, I realize that enabling CRA on internal-qa would require significant work from qa and devops, which would require a larger discussion that shouldn't necessarily block this PR.

I believe we currently allow root login in the QA builds. Perhaps if we enable CRA but still allow root login we could run some CRA specific tests using the CRA login.

live-build/misc/ansible-roles/appliance-build.challenge-response-authentication/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+---
+- command: /opt/delphix/server/bin/register/cra_control enable
```
prakashsurya: Why do we want this to be in a new `challenge-response-authentication` role, rather than included in the existing `virtualization-external` role?

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -76,3 +76,10 @@
   delay: 30
   register: result
   until: result is succeeded
+
+#
+# CRA on Linux requires that the engine UUID is stored in a file
+#
+- apt:
+    name: 'dmidecode'
+- shell: dmidecode -s system-uuid | awk '{print tolower($0)}' > /etc/engine-uuid
```
prakashsurya: we're currently running:
```
192 - command: /opt/delphix/server/bin/register/generate_secret -q
```
via the app-gate ansible role.. should we really be adding this logic, rather than relying on the app-gate ansible role?

live-build/misc/ansible-roles/appliance-build.challenge-response-authentication/tasks/main.yml
prakashsurya: Also, would it be better to just run:
```
pam-auth-update --enable challenge-response
```
rather than using this script?

I worry that change to that script, and the assumption it makes about its run time environment (e.g. it expecting to only run on a running DE, rather than this chroot/build environment), might result in build failures later (e.g. even if works now, maybe it'll break later).

prakashsurya:
Does this build-time configuration need to be preserved across an upgrade? if so, how does it get preserved (e.g. across a not-in-place upgrade)?

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
rasantel: Right. I'm undoing this part of the change and replacing it with http://reviews.delphix.com/r/50879/ which gets rid of the dependency of CRA on that file altogether.

rasantel: I'm undoing this part of the change and replacing it with http://reviews.delphix.com/r/50879/ which gets rid of the dependency of CRA on that file altogether.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
rasantel: I just got rid of the need for uuid file altogether in http://reviews.delphix.com/r/50879/, and I'll remove its generation in `generate_secret` next.

live-build/misc/ansible-roles/appliance-build.challenge-response-authentication/tasks/main.yml
rasantel: I didn't notice that existing role :) Will consolidate.

I can run `pam-auth-update` directly. Will do in the next diff.

rasantel:
> > In the interest of keeping our testing as realistic as possible, I would feel significantly more comfortable if we enabled CRA on the internal-qa variant as well. Any difference between the external variants and the internal-qa variant is an opportunity for bugs to go uncaught by our testing, and CRA seems like a pretty important feature to have minimal automated testing for.
> > That said, I realize that enabling CRA on internal-qa would require significant work from qa and devops, which would require a larger discussion that shouldn't necessarily block this PR.
> 
> I believe we currently allow root login in the QA builds. Perhaps if we enable CRA but still allow root login we could run some CRA specific tests using the CRA login.

I would be more comfortable too if testing happens on engines as close as possible to production ones.

Is this root login available only for internal variants? I hope it's not for external ones or else CRA is not of much help :)

pzakha:
> Is this root login available only for internal variants? I hope it's not for external ones or else CRA is not of much help :)

It's only for the `internal-qa` variant as blackbox needs to run some scripts as root, see https://github.com/delphix/appliance-build/blob/master/live-build/misc/ansible-roles/appliance-build.virtualization-development/tasks/main.yml#L81

rasantel:
> > Is this root login available only for internal variants? I hope it's not for external ones or else CRA is not of much help :)
> 
> It's only for the `internal-qa` variant as blackbox needs to run some scripts as root, see https://github.com/delphix/appliance-build/blob/master/live-build/misc/ansible-roles/appliance-build.virtualization-development/tasks/main.yml#L81

Looking at the Blackbox code, the Server class (and the DelphixLinuxServer subclass) use `delphix` as the "root" user, so in reality Blackbox would need some serious work and testing to make sure it can safely switch to using `root` everywhere.

rasantel:
@pzakha @jgallag88 : If you look at http://blackbox.delphix.com/master, there is a CRA box you can check to run the suite with CRA enabled. There is also a `cra` test suite to focus on that functionality and the interaction with the registration and authentication portals.

prakashsurya:
> > Is this root login available only for internal variants? I hope it's not for external ones or else CRA is not of much help :)
> 
> It's only for the `internal-qa` variant as blackbox needs to run some scripts as root, see https://github.com/delphix/appliance-build/blob/master/live-build/misc/ansible-roles/appliance-build.virtualization-development/tasks/main.yml#L81

AFAIK, we don't use the "virtualization-development" role with the "internal-qa" variant.

prakashsurya:
@rasantel do we still need to specify `cr_auth` in the `server.conf` file?
```
± g g cr_auth
live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml:# The "product", "platform", and "cr_auth" fields are intentionally left
live-build/misc/ansible-roles/appliance-build.virtualization-external/tasks/main.yml:    - { key: 'cr_auth', value: 'true' }
live-build/misc/ansible-roles/appliance-build.virtualization-internal/tasks/main.yml:    - { key: 'cr_auth', value: 'false' }
```
```
± tail -n 7 live-build/misc/ansible-roles/appliance-build.virtualization-external/tasks/main.yml
---
- lineinfile:
    path: /var/opt/delphix/server.conf
    regexp: "^{{ item.key }}="
    line: "{{ item.key }}={{ item.value }}"
  with_items:
    - { key: 'cr_auth', value: 'true' }
```
Does this have anything to do with CRA?

Sorry if this field is unrelated to this PR...

rasantel:
> @rasantel do we still need to specify `cr_auth` in the `server.conf` file?
> 
> ```
> ± g g cr_auth
> live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml:# The "product", "platform", and "cr_auth" fields are intentionally left
> live-build/misc/ansible-roles/appliance-build.virtualization-external/tasks/main.yml:    - { key: 'cr_auth', value: 'true' }
> live-build/misc/ansible-roles/appliance-build.virtualization-internal/tasks/main.yml:    - { key: 'cr_auth', value: 'false' }
> ```
> 
> ```
> ± tail -n 7 live-build/misc/ansible-roles/appliance-build.virtualization-external/tasks/main.yml
> ---
> - lineinfile:
>     path: /var/opt/delphix/server.conf
>     regexp: "^{{ item.key }}="
>     line: "{{ item.key }}={{ item.value }}"
>   with_items:
>     - { key: 'cr_auth', value: 'true' }
> ```
> 
> Does this have anything to do with CRA?
> 
> Sorry if this field is unrelated to this PR...

@prakashsurya I'm not aware of any use of this field, from looking at the usual gates. I asked Dave Kapell and he doesn't know of any use either. It was introduced during the first implementation of CRA: http://reviews.delphix.com/r/16118/diff

PR #331, opened 2019-07-11
Change the order of live-build configuration hooks

prakashsurya:
This change re-orders the configuration hooks. Previously the order was:

 1. Run "delphix-platform" ansible roles
 2. Run "appliance-build" ansible roles
 3. Build and install "delphix-entire" package

Since landing https://github.com/delphix/delphix-platform/pull/110, the
"delphix-platform" ansible roles now depend on the "delphix-entire"
package being installed before they run.

Thus, we need to reorder the steps above, such that (1) runs after (3).
Additionally, we still need to ensure (2) runs before (3), since we rely
on (2) to install various packages for each appliance-build variant.

Taking this into account, the new order of these hooks is the following:

 a. Run "appliance-build" ansible roles
 b. Build and install "delphix-entire" package
 c. Run "delphix-platform" ansible roles

As a bonus, this also enables Ansible roles installed by packages
installed via (a), to get run during the build via (c), which previously
would not have been the case.


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1627/)

PR #332, opened 2019-07-11
Change the order of live-build configuration hooks

prakashsurya:
This change re-orders the configuration hooks. Previously the order was:

 1. Run "delphix-platform" Ansible roles
 2. Run "appliance-build" Ansible roles
 3. Build and install "delphix-entire" package

Since landing https://github.com/delphix/delphix-platform/pull/110, the
"delphix-platform" Ansible roles now depend on the "delphix-entire"
package being installed before they run.

Thus, we need to reorder the steps above, such that (1) runs after (3).
Additionally, we still need to ensure (2) runs before (3), since we rely
on (2) to install various packages for each appliance-build variant.

Taking this into account, the new order of these hooks is the following:

 a. Run "appliance-build" Ansible roles
 b. Build and install "delphix-entire" package
 c. Run "delphix-platform" Ansible roles

As a bonus, this also enables Ansible roles installed by packages
installed via (a), to get run during the build via (c), which previously
would not have been the case.


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1637/)

prakashsurya:
This depends on the following app-gate change: http://reviews.delphix.com/r/50917/

prakashsurya:
cc: @pzakha @grwilson

prakashsurya:
Before:
```
$ grep Client /etc/ssh/sshd_config
ClientAliveInterval 1800
ClientAliveCountMax 0
```
After:
```
$ grep Client /etc/ssh/sshd_config
#ClientAliveInterval 0
#ClientAliveCountMax 3
```
This verifies that the variant is now being consumed properly by the delphix-platform Ansible roles.

live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -25,20 +25,23 @@
   when: lookup('env', 'APPLIANCE_PASSWORD') == ''
   no_log: true
 
-#
-# While the "delphix-platform" package will take care of creating the
-# delphix user, it won't configure the user's password. Thus, we have to
-# do that here, during the intial build of the appliance.
-#
+- file:
```
pzakha: We should follow this up by removing the duplicate code from delphix-platform

prakashsurya: Yea, I'll do this after landing.

PR #333, opened 2019-07-12
upgrade script should be able to do deferred upgrades

grwilson:
We want to be able to upgrade our dcenter instances on a nightly basis and the plan is to use "deferred" upgrades to ensure that we don't see any kernel/userland inconsistencies. This change allows the user to select either "deferred" or "full" upgrades similar to our application stack. For "full" upgrades, it will determine if a not-in-place upgrade is required.

### DIFF
--- a/upgrade/upgrade-scripts/upgrade
+++ b/upgrade/upgrade-scripts/upgrade
@@ -51,8 +51,8 @@ function usage() {
 	PREFIX_NCHARS=$(echo -n "$PREFIX_STRING" | wc -c)
 	PREFIX_SPACES=$(printf "%.s " $(seq "$PREFIX_NCHARS"))
 
-	echo "$PREFIX_STRING [-n] [-v] in-place"
-	echo "$PREFIX_SPACES [-n] [-v] not-in-place"
+	echo "$PREFIX_STRING [-n] [-v] deferred"
+	echo "$PREFIX_SPACES [-n] [-v] full"
 	echo "$PREFIX_SPACES rollback"
 
 	exit 2
@@ -98,6 +98,14 @@ function cleanup_in_place_upgrade() {
 function upgrade_in_place() {
 	trap cleanup_in_place_upgrade EXIT
 
+	case "$UPGRADE_TYPE" in
+	DEFERRED | FULL) ;;
+	*)
+		die "upgrade type ('$UPGRADE_TYPE') is not supported for " \
+			"in-place upgrade."
+		;;
+	esac
+
 	CONTAINER=$("$IMAGE_PATH/upgrade-container" create in-place)
 	[[ -n "$CONTAINER" ]] || die "failed to create container"
 
@@ -122,8 +130,8 @@ function upgrade_in_place() {
 	#
 	[[ "$DLPX_UPGRADE_DRY_RUN" == "true" ]] && return
 
-	set_upgrade_property "UPGRADE_TYPE" "FULL" ||
-		die "failed to set upgrade property 'UPGRADE_TYPE'"
+	set_upgrade_property "UPGRADE_TYPE" "$UPGRADE_TYPE" ||
+		die "failed to set upgrade property 'UPGRADE_TYPE' to '$UPGRADE_TYPE'"
 
 	set_upgrade_property "UPGRADE_BASE_CONTAINER" \
 		"$(get_mounted_rootfs_container_name)" ||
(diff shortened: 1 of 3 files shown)


grwilson:
[ab-pre-push](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1643/flowGraphTable/)

grwilson:
I have tried both deferred and full upgrades on an AWS instance.

upgrade/upgrade-scripts/upgrade
```diff
@@ -329,16 +331,30 @@ done
 shift $((OPTIND - 1))
 
 case "$1" in
-in-place)
+deferred)
+	UPGRADE_TYPE="${1^^}"
```
pzakha: nit: Today I learned a new obscure bash string substitution ;). I think it might be more readable to just put `UPGRADE_TYPE ="DEFERRED"` here though.

upgrade/upgrade-scripts/upgrade
```diff
@@ -329,16 +331,30 @@ done
...
+	# FULL upgrade always perform a reboot but can take on two
+	# flavors -- in-place or not-in-place. The flavor of a FULL upgrade
+	# is determined by the MINIMUM_REBOOT_OPTIONAL_VERSION. If our
+	# upgrade version is not greater than MINIMUM_REBOOT_OPTIONAL_VERSION
+	# than the upgrade will be a not-in-place upgrade.
+	#
+	if is_upgrade_in_place_allowed; then
+		upgrade_in_place "$@"
+	else
+		upgrade_not_in_place "$@"
```
pzakha: Do you think it could be useful to have an option to force `not-in-place` upgrade, at least to make it easier to test that feature?

grwilson: I had discussed this with @prakashsurya yesterday and there is a `-x` option you can use when you `unpack-image` that will force that. So I left a force flag out here since that interface was already being used by QA. I'm happy to add an override.

upgrade/upgrade-scripts/upgrade
grwilson: Hehehe. I will change this.

upgrade/upgrade-scripts/upgrade
pzakha: I see. I'm fine with that as long as we have a way. Perhaps add a comment here that mentions that `unpack-image` flag?

pzakha: Regarding the change in the name of the commands, I noticed that upgrade testing succeeded although I haven't seen any related app-gate changes in your ab-pre-push link. Either I'm missing something, or somehow you haven't been testing the right changes?

grwilson: Upgrade testing doesn't use this interface. This will only be used by developers and dcenter upgrade. QA has discussed adding testing for this to ensure that it doesn't break in the future but that doesn't exist right now.

pzakha: I see, thanks for the clarifications.

upgrade/upgrade-scripts/upgrade
```diff
@@ -329,16 +331,31 @@ done
 shift $((OPTIND - 1))
 
 case "$1" in
-in-place)
+deferred)
+	UPGRADE_TYPE="DEFERRED"
```
prakashsurya: I'm OK with this.. but I probably would prefer to put:
```
UPGRADE_TYPE="${1^^}"
```
just before the `case` statement, so that we only need to have a single line to set the `UPGRADE_TYPE` variable, and it'll "just work" if we ever add another type of upgrade..

but I'm OK with this approach too, if folks prefer this..

upgrade/upgrade-scripts/upgrade
```diff
@@ -122,8 +122,8 @@ function upgrade_in_place() {
 	#
 	[[ "$DLPX_UPGRADE_DRY_RUN" == "true" ]] && return
 
-	set_upgrade_property "UPGRADE_TYPE" "FULL" ||
-		die "failed to set upgrade property 'UPGRADE_TYPE'"
+	set_upgrade_property "UPGRADE_TYPE" "$UPGRADE_TYPE" ||
+		die "failed to set upgrade property 'UPGRADE_TYPE' to $UPGRADE_TYPE"
```
prakashsurya: Can we wrap `$UPGRADE_TYPE` with quotes.. it's not necessary, but I find this convention useful since it can denote when the variable's value begins and ends, when the message is printed, which can be useful if the variable contains something unexpected (e.g. when it's empty, or when it has trailing/leading whitespace, etc).

prakashsurya:
Also, can you update the `README.md` doc in the `upgrade` directory: https://github.com/delphix/appliance-build/blob/master/upgrade/README.md .. it currently refers to passing in `in-place`, so we should change this to `deferred`.

upgrade/upgrade-scripts/upgrade
grwilson: I also wanted to do this but opted to move it into the `case` block since `rollback` also uses the `UPGRADE_TYPE` variable and did not want to pass that value into that function.

PR #335, opened 2019-07-15
Enable use of primary and secondary APT mirrors

### LINKED ISSUE #319: Want parameters for setting upstream repositories
`appliance-build` fetches package metadata from upstream repositories in three different places:

1. Through [the command-line arguments to `lb config` in `live-build/auto/config`](https://github.com/delphix/appliance-build/blob/91bd8c9c99af02e81a3fce3e8354a88be8186ca1/live-build/auto/config#L33-L63)
2. Through [the [non-`localhost`] `.key` and `.list` files in `live-build/config/archives`](https://github.com/delphix/appliance-build/tree/91bd8c9c99af02e81a3fce3e8354a88be8186ca1/live-build/config/archives)
3. Through [the `apt_repository` Ansible module in `live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml`](https://github.com/delphix/appliance-build/blob/91bd8c9c99af02e81a3fce3e8354a88be8186ca1/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml#L57-L66)

As part of the automatic update process for third-party packages being implemented in [LB-117](https://jira.delphix.com/browse/LB-117), we need a way to test a possible update of third-party packages prior to committing it. As such, we need a way to invoke `appliance-build` with staging (i.e., pre-push) versions of all three types of third-party package repositories. This could be in the form of command-line arguments to the Gradle invocation or environment variables. We would invoke `appliance-build` with these command-line arguments or environment variables to test a possible update of third-party packages. If the build and tests pass, we could then commit that update.

prakashsurya:
This change better enables us to use internal APT package mirrors in the
build, and to dynamically specify the URLs of such mirrors.

Specifically, this allows us to specify a "primary" mirror (which
is enabled directly with "lb config"), and a "secondary" mirror (which
is enabled via the "config/archives" directory). Further, the URLs for
both the "primary" and "secondary" mirrors can (and must!) be specified
via the two new environment variables:

 * DELPHIX_PACKAGE_MIRROR_PRIMARY
 * DELPHIX_PACKAGE_MIRROR_SECONDARY

Closes #319

### DIFF
--- a/live-build/build.gradle
+++ b/live-build/build.gradle
@@ -105,9 +105,11 @@ for (variant in allVariants) {
                 inputs.file "${rootProject.projectDir}/scripts/run-live-build.sh"
 
                 for (envVar in ["APPLIANCE_PASSWORD",
+                                "DELPHIX_APPLIANCE_VERSION",
+                                "DELPHIX_PACKAGE_MIRROR_PRIMARY",
+                                "DELPHIX_PACKAGE_MIRROR_SECONDARY",
                                 "DELPHIX_SIGNATURE_URL",
-                                "DELPHIX_SIGNATURE_TOKEN",
-                                "DELPHIX_APPLIANCE_VERSION"]) {
+                                "DELPHIX_SIGNATURE_TOKEN"]) {
                     inputs.property(envVar, System.getenv(envVar)).optional(true)
                 }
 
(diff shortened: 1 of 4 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1677/)

I've also tested this via the following:
```
```

Note, this change cannot land until we have deployed mirrors that can be used by default, and the `run-live-build.sh` script updated to use these default mirrors, when no mirrors are specified by the user.

PR #337, opened 2019-07-18
deferred upgrades need to restart delphix-platform

grwilson:


upgrade/upgrade-scripts/upgrade
```diff
@@ -155,6 +155,9 @@ function upgrade_in_place() {
 
 	if [[ "$UPGRADE_TYPE" == "FULL" ]]; then
 		systemctl reboot || die "'systemctl reboot' failed"
+	else
+		systemctl restart delphix-platform ||
```
prakashsurya: I think we also need to `systemctl reload` the service, so that it'll remove the `ansible-done` file, and ensure that the new Ansible logic will actually get executed.

grwilson: Fixed.

jgallag88: Not directly related to this change, but I wonder if we should start having code remove the `ansible-done` file directly, rather than using `systemctl reload`. `systemctl reload` won't work if the service isn't currently active. If `delphix-platform` is in an unexpected state for some reason (maybe we are rerunning this script after if previously failed or something), then this will fail.

prakashsurya: @jgallag88 I'm open to discussing this further, perhaps in some other bug or slack.. I've thought about the same, but don't have anything to point to that being a practical problem.

PR #338, opened 2019-07-22
Ship with an NFSv4-only configuration by default.

sebroy:
This change causes our virtualization images to ship with an NFSv4-only NFS server configuration by default out of the box. The virtualization software is responsible for dynamically enabling NFSv3 services as needed. This is being implemented in http://reviews.delphix.com/r/50851. Without the dlpx-app-gate changes, the virtualization software will not be able to mount NFSv3-based VDBs, so those changes must integrate before this PR is merged into appliance-build.

There are three changes implemented here:

1. The ansible task that creates the initial NFS server configuration in `/etc/default/nfs-kernel-server`. Note that this code previously lived in dlpx-app-gate, and is being removed from there as part of http://reviews.delphix.com/r/50851. It is being moved here because this is one-time initial configuration, and the ansible tasks from dlpx-app-gate are applied at arbitrary times when the delphix-platform service is reloaded.
2. Upgrade code that copies the `/etc/default/nfs-kernel-server` file for not-in-place upgrades (because that file can be modified by the virtualization software as reviewed at http://reviews.delphix.com/r/50851).
3. Upgrade code that preserves the state of NFSv3-related systemd services for not-in-place upgrades. This code had to be enhanced to support carrying over masked state of services.

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
@@ -45,3 +45,34 @@
       domain=
       dns1=
       dns2=
+
+#
+# Initial NFS Server configuration for virtualization. We do this here instead
+# of within the virtualization package because the virtualization software
+# modifies this configuration at runtime, and we don't want ansible roles that
+# are re-run as part of upgrades to undo these runtime modifications.
+#
+- name: Link initial NFS configuration
+  file:
+    src: /opt/delphix/server/etc/nfs-kernel-server-v4
+    dest: /etc/default/nfs-kernel-server
+    force: yes
+    owner: root
+    group: root
+    state: link
+
+#
+# Because we want an NFSv4-only configuration out of the box, we need to mask
+# NFSv3 services so that they don't get automatically started at boot via
+# dependencies. The virtualization software is responsible for unmasking and
+# starting these services if NFSv3 needed at runtime.
+#
+- name: Mask NFSv3 services
+  file:
+    src: "/dev/null"
+    dest: "/etc/systemd/system/{{ item }}"
+    state: link
+  with_items:
+      - rpc-statd.service
+      - rpcbind.service
+      - rpcbind.socket
--- a/upgrade/upgrade-scripts/upgrade-container
+++ b/upgrade/upgrade-scripts/upgrade-container
@@ -471,8 +471,8 @@ function migrate_file() {
 		parentdir="$(dirname "$path")"
 		mkdir -p "/var/lib/machines/${CONTAINER}${parentdir}" ||
 			die "'mkdir -p $parentdir' failed for '$CONTAINER'"
-		cp -p "$path" "/var/lib/machines/${CONTAINER}${path}" ||
-			die "'cp -p $path' failed for '$CONTAINER'"
+		cp -Pp "$path" "/var/lib/machines/${CONTAINER}${path}" ||
+			die "'cp -Pp $path' failed for '$CONTAINER'"
 	else
 		#
 		# If the target file doesn't exist in the host then we
@@ -507,13 +507,30 @@ function migrate_dir() {
 	fi
 }
 
+#
+# Preserve the persistent service state of the service named as an argument.
+#
 function migrate_svc() {
 	local svc="$1"
+	local state
+	state=$(systemctl is-enabled "$svc")
 
-	if systemctl is-enabled -q "$svc"; then
-		chroot "/var/lib/machines/$CONTAINER" systemctl enable "$svc"
+	if [[ $state == "masked" ]]; then
+		chroot "/var/lib/machines/$CONTAINER" systemctl mask "$svc"
 	else
-		chroot "/var/lib/machines/$CONTAINER" systemctl disable "$svc"
+		#
+		# The service may be masked by default, so always unmask before
+		# doing anything else. Otherwise, systemctl will ignore the new
+		# setting.
+		#
+		chroot "/var/lib/machines/$CONTAINER" systemctl unmask "$svc"
+		if systemctl is-enabled -q "$svc"; then
+			chroot "/var/lib/machines/$CONTAINER" systemctl \
+				enable "$svc"
+		else
+			chroot "/var/lib/machines/$CONTAINER" systemctl \
+				disable "$svc"
+		fi
 	fi
 }
 
@@ -540,7 +557,11 @@ function migrate_configuration() {
 		migrate_svc "$svc"
 	done <<-EOF
 		delphix-masking.service
+		nfs-mountd.service
 		ntp.service
+		rpc-statd.service
+		rpcbind.service
+		rpcbind.socket
 	EOF
 
 	#
@@ -560,6 +581,7 @@ function migrate_configuration() {
 	while read -r file; do
 		migrate_file "$file"
 	done <<-EOF
+		/etc/default/nfs-kernel-server
 		/etc/hostid
 		/etc/hostname
 		/etc/hosts


upgrade/upgrade-scripts/upgrade-container
```diff
@@ -560,6 +580,7 @@ function migrate_configuration() {
 	while read -r file; do
 		migrate_file "$file"
 	done <<-EOF
+		/etc/default/nfs-kernel-server
```
prakashsurya: Why do we migration this file on upgrade?

Since we edit the file using delphix-platform, I think we can simply use the default contents for not-in-place upgrade, and then (again) rely on delphix-platform to properly modify it, right? This would help us get any changes made by the default version of the file, while also allowing us to make our "local" customizations on upgrade.

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -507,13 +507,30 @@ function migrate_dir() {
 	fi
 }
 
+#
+# Preserve the persistent service state of the service named as an argument.
```
prakashsurya: I don't know if this would be better, but we could perhaps change this function to only enable services if they're currently enabled. This would then rely on the default value of all services, which would start out as masked or disabled (at least for the services we manipulate), and only enable them when needed. This might make this function's implementation much more simple.

With that said, I'm OK with what you have here too.

live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
```diff
@@ -45,3 +45,49 @@
...
+# dependencies. The virtualization software is responsible for unmasking and
+# starting these services if NFSv3 needed at runtime.
+#
+- name: Mask NFSv3 services
+  file:
+    src: "/dev/null"
+    dest: "/etc/systemd/system/{{ item }}"
+    state: link
+  with_items:
+      - nfs-mountd.service
```
prakashsurya: Can we move this to the `debian/postinst` file, just to be consistent with where we manipulate other services?

Also, can we not call `systemctl mask` in the build/chroot environment?

upgrade/upgrade-scripts/upgrade-container
sebroy: We don't edit this file using delphix-platform. The virtualization code modifies the file as-needed at runtime (see http://reviews.delphix.com/r/50851/).

prakashsurya: > Upgrade code that copies the /etc/default/nfs-kernel-server file for not-in-place upgrades (because that file can be modified by the virtualization software as reviewed at http://reviews.delphix.com/r/50851).

Sorry, I see this was already explained. If we also modify this file via the virtualization software, then yes, we need to migrate this file like you're doing. I'll close this issue.

live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
sebroy: I think putting this in `debian/postinst` of any package would result in our potentially losing our customizations during in-place upgrades, right?

We might be able to call systemctl, I'll have to try it.

upgrade/upgrade-scripts/upgrade-container
prakashsurya: Now that I better understand how we won't (re-)mask the NFS services on upgrade, I think we have to keep the logic you have, to ensure we keep a masked service as masked after upgrade.

live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
prakashsurya: > I think putting this in debian/postinst of any package would result in our potentially losing our customizations during in-place upgrades, right?

Ah yes.. Sorry, I mistakenly read this as logic that'll run on upgrade too (i.e. in delphix-platform); ignore my comment.

> Also, can we not call systemctl mask in the build/chroot environment?

I doubt that we can use the `systemd` ansible module, so I'm good with your solution. I think it's better than using `command: systemctl ...` in ansible (`command` isn't idempotent, whereas your solution is).

sebroy:
I made a change to the original PR to add `rpcbind.socket` to the set of services to mask, and also to make sure that `rpc.mountd` gets started with `--no-nfs-version 2 --no-nfs-version 3` so that it doesn't attempt to register any external services with `rpcbind` on startup (`rpcbind` itself is masked). Having the `rpcbind.socket` active will cause `rpc.rpcbind` to take a very long time to start even when it's not registering external services because it unconditionally always tries to unregister services as a matter of course.

pzakha:
Looking at those changes and what we find in the app-gate review, I'm wondering if we shouldn't just supply 2 different configuration files: one with nfsv3 enabled and another one with nfsv3 disabled. Then we just replace the config file by a symlink to the right file depending if we want nfsv3 enabled or not.

sebroy:
@pzakha I love that idea. It makes it clear what each configuration looks like as checked into the source, and is less error prone. Thanks for that suggestion, I'll see what the best way to implement that would be, and whether the packaging subsystem is ok with replacing a plain config file with a symbolic link.

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -471,7 +471,7 @@ function migrate_file() {
 		parentdir="$(dirname "$path")"
 		mkdir -p "/var/lib/machines/${CONTAINER}${parentdir}" ||
 			die "'mkdir -p $parentdir' failed for '$CONTAINER'"
-		cp -p "$path" "/var/lib/machines/${CONTAINER}${path}" ||
+		cp -Pp "$path" "/var/lib/machines/${CONTAINER}${path}" ||
```
pzakha: We might want to also update the logic on line 572 for `resolve.conf` to check that it is not a symlink before calling `migrate_file` on it.

sebroy: I think the comment had good intentions, but the code was previously wrong. It looks like migration was always copying `/etc/resolv.conf` regardless of whether or not it was a symlink.
```
root@ip-10-110-235-164:~# ls -l /etc/default/nfs-kernel-server 
lrwxrwxrwx 1 root root 44 Jul 29 15:20 /etc/default/nfs-kernel-server -> /opt/delphix/server/etc/nfs-kernel-server-v4
root@ip-10-110-235-164:~# if [[ -f /etc/default/nfs-kernel-server ]]; then
> echo "yep, a regular file"
> fi
yep, a regular file
```

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -471,7 +471,7 @@ function migrate_file() {
...
 			die "'cp -p $path' failed for '$CONTAINER'"
```
prakashsurya: please also update this error message; it’s slightly off now.

sebroy:
Now that the dlpx-app-gate changes have integrated and have stabilized, here are ab-pre-push runs for this change:
`git ab-pre-push`: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1897/
`git ab-pre-push --test-upgrade-from 5.3.6.0`: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1898/

sebroy:
From a VM cloned from one of the above builds:
```
delphix@ip-10-110-241-243:~$ ls -l /etc/default/nfs-kernel-server 
lrwxrwxrwx 1 root root 44 Aug  7 21:27 /etc/default/nfs-kernel-server -> /opt/delphix/server/etc/nfs-kernel-server-v4
delphix@ip-10-110-241-243:~$ rpcinfo
rpcinfo: can't contact rpcbind: RPC: Remote system error - No such file or directory
delphix@ip-10-110-241-243:~$ sudo netstat -anp --tcp | grep LISTEN | grep -v '127\.0\.0'
tcp        0      0 10.110.241.243:80       0.0.0.0:*               LISTEN      4649/nginx: master  
tcp        0      0 0.0.0.0:8341            0.0.0.0:*               LISTEN      5326/java           
tcp        0      0 0.0.0.0:22              0.0.0.0:*               LISTEN      669/sshd            
tcp        0      0 0.0.0.0:35799           0.0.0.0:*               LISTEN      5326/java           
tcp        0      0 10.110.241.243:443      0.0.0.0:*               LISTEN      4649/nginx: master  
tcp        0      0 0.0.0.0:3260            0.0.0.0:*               LISTEN      -                   
tcp        0      0 0.0.0.0:8415            0.0.0.0:*               LISTEN      5326/java           
tcp        0      0 0.0.0.0:2049            0.0.0.0:*               LISTEN      -                   
tcp        0      0 0.0.0.0:873             0.0.0.0:*               LISTEN      4573/rsync          
```

PR #340, opened 2019-07-24
Add pause and stress options scripts to migration images

pzakha:
This adds the pause and stress options scripts to the migration image.

Note that for now the scripts are taken as is from the 5.3/stage branch. They have been slightly modified to comply with the shell format checks.

Note that we are going to add some new options and remove those that do not apply to migration in follow-up work, as we will be iterating on the migration-specific stress tests.

## TESTING
ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1772/

### DIFF
--- a/live-build/misc/migration-scripts/dx_execute
+++ b/live-build/misc/migration-scripts/dx_execute
@@ -32,6 +32,8 @@ umask 0022
 
 set -o pipefail
 
+DX_UPG_PAUSE="${BASH_SOURCE%/*}/dx_upg_pause_options"
+
 function die() {
 	echo "$(basename "$0"): $*" >&2
 	exit 1
@@ -161,6 +163,9 @@ sed -i '/autoboot_delay/c\set autoboot_delay=20' /boot/menu.rc.local ||
 MAIN_MENU_LINUX_CMD=$(grep -F 'mainmenu_command[8]' /boot/menu.rc.local |
 	cut -d = -f 2-)
 echo "set menu_timeout_command=$MAIN_MENU_LINUX_CMD" >>/boot/menu.rc.local
+
+$DX_UPG_PAUSE --pause "PAUSE_IN_DX_EXECUTE_BEFORE_RESTART" ||
+	die "failed to pause fully on stress option"
 
 # Constants used by the uadmin syscall.
 A_SHUTDOWN=2
--- a/live-build/misc/migration-scripts/dx_upg_pause_options
+++ b/live-build/misc/migration-scripts/dx_upg_pause_options
@@ -1,0 +1,216 @@
+#!/bin/bash
+#
+# Copyright (c) 2017, 2018 by Delphix. All rights reserved.
+#
+
+#
+# Utility for managing upgrade pause options.
+# Supported pause options listed below.
+#
+
+PAUSE_OPTIONS_JSON=$(cat <<-EOF
+{
+	"PAUSE_IN_DX_EXECUTE_BEFORE_RESTART": {
+	    "location": "right before reboot"
+	},
+	"PAUSE_IN_DXSVCINIT_BEFORE_UPGRADE": {
+	    "location": "post-reboot before any app upgrade logic"
+	},
+	"PAUSE_BEFORE_UPGRADE_COMMIT": {
+	    "location": "mgmt service (before point of no return)"
+	},
+	"PAUSE_BEFORE_CHECKPOINT_DISCARD": {
+	    "location": "mgmt service (before point of no return)"
+	},
+	"PAUSE_IN_SVC_ROLLBACK_BEFORE_ROLLBACK": {
+	    "location": "rollback service"
+	},
+	"PAUSE_BEFORE_REWIND_CHECKPOINT": {
+	    "location": "rollback service"
+	},
+	"PAUSE_AFTER_REWIND_CHECKPOINT": {
+	    "location": "rollback service"
+	},
+	"PAUSE_IN_DXSVCINIT_AFTER_ROLLBACK": {
+	    "location": "post-rollback before app stack starts"
+	},
+	"PAUSE_IN_SVC_MGMT_BEFORE_STACK_STARTUP": {
+	    "location": "before app stack starts"
+	}
+}
+EOF
+)
+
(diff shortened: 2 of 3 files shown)


pzakha:
cc: @prakashsurya , @palashgandhi

pzakha:
@prakashsurya There is a few reasons for this.
So we have 3 places where the script is to be found: 
1. `/opt/delphix` of the current version (say 5.3.6.0 as example)
2. `/opt/delphix` of the new version (say 6.0.0.0 as example)
3. The upgrade image: `/var/dlpx-update/` 

When the scripts are being referenced by code running on the current (pre-upgrade) version of the product, we use the scripts in (1). E.g. the app-stack in 5.3.6.0 will call the scripts in `/opt/delphix` to check if it needs to pause.

When the scripts are being referenced by upgrade-specific code (e.g. running `dx_execute`), it will call the scripts in (3), since those are the most up-to-date scripts, that know about the particular options that were added for this version of `dx_execute`.

After an upgrade, the new code will run the post-upgrade logic and may need to check if some options are set. It will use the scripts in `/opt/delphix` of the running (post-upgrade) version, i.e the scripts from (2).

Now, all that I have mentioned so far was how the scripts were referenced from the product itself. There is another user of the scripts, which is blackbox, and it is the one that actually sets those options. BlackBox currently uses the scripts in `/var/dlpx-update` to set those options.

---

The scripts that BlackBox uses need to know about both the options from the old version and the new version, hence why it currently references the scripts in `/var/dlpx-update` for Illumos upgrade stress tests.

prakashsurya:
It's still not clear to my _why_ we need option (3). I understand that you're trying to add (3), but I don't see why it's needed.

For example, if we need to add a new option to be used by `dx_execute` on the current version, why don't we just add the option to the current version of the product (e.g. perhaps via a backport commit)? Especially considering these scripts/options are only used internally.

With your change, if we need to add an option to be available in all situation (1, 2, and 3), we'd need to make a changes to the scripts here, and to the scripts in the app-gate, right? and/or if we only need to target one of the situations, we'd have to determine which script to change (and understand the subtle differences you mentioned)?

While I don't particularly like how we have differences between the scripts used (and stress points available) "pre-upgrade" and "post-upgrade", I think adding a third difference (i.e. adding this new "during-upgrade" case) only makes the situation more complicated, and I don't yet understand why this additional complexity is required.

pzakha:
So right now blackbox is already expecting to find the script in (3).

With this change, whatever options we add will need to be present in both scripts (2) and (3). We do not need to do backports to script (1).

With the change that you are proposing we'd need to change scripts (1) and (2) and we'd need to change blackbox.

Ideally at migration image build time, we would copy the script from (1) into (3), but I'm not sure how easy that would be.

prakashsurya:
> So right now blackbox is already expecting to find the script in (3).

Which branch of blackbox? I remember talking to Jeff, and asking him to start using the script from `/opt/delphix` via the qa-gate code.

> With this change, whatever options we add will need to be present in both scripts (2) and (3). We do not need to do backports to script (1).
> 
> With the change that you are proposing we'd need to change scripts (1) and (2) and we'd need to change blackbox.

Correct. We'd make the change on trunk to the app-gate, and then backport that to any release branch(es) that require it.

I think this is a more simple solution, since we have a single "source" script (the script in the app-gate repository), rather than two "source" scripts (the script in the app-gate repository, and the script in the appliance-build repository). Further, during runtime, we only have cases (1) and (2) to worry about, rather then (1), (2), and (3).

> Ideally at migration image build time, we would copy the script from (1) into (3), but I'm not sure how easy that would be.

I'm not sure I agree, but honestly I haven't thought about it enough. I feel like the approach of copying the script from the app-gate into the upgrade image is also difficult to reason about, since we still wind up with differences between cases (1) and (2).

I'm not sure if this would work for all cases, but I could imagine a scheme where blackbox can specify which stress options are available, and then we have the app-gate consume that. This way, the options that're available will be the same regardless of whether it's "pre-upgrade", "during-upgrade, or "post-upgrade" since blackbox will be controlling that. Although, this wouldn't work if we need to use these stress options outside of blackbox.

With that said, since we don't change the available options frequently, I don't think we need any complicated mechanism for this.

pzakha:
> With that said, since we don't change the available options frequently, I don't think we need any complicated mechanism for this.

For the next weeks / months we'll be adding more options for migration.

The issue with having to backport the scripts is that let's say we add most options to 5.3.6.0, but then afterwards we cut-off to 5.3.7.0 and want to add more options at that time. We will then have different options depending on whether we are at 5.3.7.0 or 5.3.6.0, which would make it more complex if we want to test upgrade from 5.3.6.0.

Neither alternative is ideal, but I find that the alternative I'm proposing is easier to work with as all of the new options come together with the migration image.

prakashsurya:
> The issue with having to backport the scripts is that let's say we add most options to 5.3.6.0, but then afterwards we cut-off to 5.3.7.0 and want to add more options at that time. We will then have different options depending on whether we are at 5.3.7.0 or 5.3.6.0, which would make it more complex if we want to test upgrade from 5.3.6.0.

Yes, I understand. I'm still not convinced. I admit the current design is lacking (I was already aware of these deficiencies when committing to this design for L to L upgrades), but I'm not sure I agree that we should have a different design for 5.3 to L upgrades, vs. L to L upgrades.

Further, I still don't like the idea of having this script duplicated in multiple repositories (which is why I didn't do this for L to L upgrades). So far, I think we're going in circles, and none of my concerns have been addressed.

But I won't block this, if you want to push this through; but I'm still not convinced this is a good design.

pzakha:
Thanks for the review Prakash.

Keep in mind that this change is only going to be there for the duration of migration, and that all those files will be cleaned up later, so the decision of where to put them is not as critical. Currently for consistency sake there is an easy rule to follow: we use scripts in `/var/dlpx-update` when we are in Illumos and then `/opt/delphix` when we are in Linux.

pzakha:
cc'ing @shartse @sdimitro for review

prakashsurya:
> Keep in mind that this change is only going to be there for the duration of migration

i.e. years, right?

PR #342, opened 2019-08-01
Store engine uuid before migration

pcd1193182:
This change works with an associated app-gate change that uses the cached UUID in place of the system's detected UUID if the cached version and the system's version are potentially byteswapped versions of each other.

Tested by migrating aws 5.3.5.0 VM, verified that reported UUID remained the same. In addition, ab-pre-push testing is in progress.

### DIFF
--- a/live-build/misc/migration-scripts/dx_apply
+++ b/live-build/misc/migration-scripts/dx_apply
@@ -195,6 +195,17 @@ cat <<-EOF >"$TMP_ROOT/etc/fstab" || die "Failed to setup /etc/fstab"
 EOF
 
 #
+# On some platforms (currently AWS and new enough versions of ESX), the
+# engine's UUID as reported by the os-level tools changes after
+# migration. This is because of a confusion over the proper endianness of the
+# first three fields of the UUID. To avoid having engine UUIDs change after
+# migration, we store the current UUID before we finish the migration, and
+# then use that on Linux.
+#
+/opt/delphix/server/bin/bos_mgmt get_system_uuid appliance >"$TMP_ROOT/etc/engine-uuid" ||
+	die "failed to copy engine uuid"
+
+#
 # Set things up for the FreeBSD bootloader to provide an option for
 # booting into the Linux environment This happens in two steps:
 # [1] We place vmlinuz and initrd from our migration image into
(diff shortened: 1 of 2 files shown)


live-build/misc/migration-scripts/dx_apply
```diff
@@ -194,6 +194,16 @@ cat <<-EOF >"$TMP_ROOT/etc/fstab" || die "Failed to setup /etc/fstab"
...
 
+#
+# On some platforms (currently AWS and new enough versions of ESX), the
+# engine's UUID as reported by the os-level tools changes after
+# migration. This is because of a confusion over the proper endianness of the
+# first three fields of the UUID. To avoid having engine UUIDs change after
+# migration, we store the current UUID before we finish the migration, and
+# then use that on Linux.
+#
+sudo /opt/delphix/server/bin/bos_mgmt get_system_uuid appliance 2>&1 $TMP_ROOT/etc/engine-uuid
```
pzakha: We'd probably want not to pipe stderr into the file.
Also, we should add a `die` statement for if this fails.

grwilson:
You will need to also migrate the `/etc/engine-uuid` file for the not-in-place upgrade case (see https://github.com/delphix/appliance-build/blob/master/upgrade/upgrade-scripts/upgrade-container#L600)

pzakha:
Good point George. We should do this while we are there since it's a change to the same repository.

prakashsurya:
Is CRA going to be updated to use the `/etc/engine-uuid` file? I think it was just updated to not use it; IIRC, that's why we just recently made the change to stop migrating this file on upgrade (see [here](https://github.com/delphix/appliance-build/commit/f614a05b464cd3f5c626106c903831efd62a258f)).

pcd1193182:
@prakashsurya It looks like I need to augment my app-gate change to tweak challenge_response/challenge_response.c to invoke the script, rather than run the system commands directly. So yes, it'll be updated to use the file indirectly, the same way the app-stack will.

pzakha:
@pcd1193182 It would be good to have the same script that can be called by:
 - app-gate
 - qa-gate tests that test the functionality of the API for getting system info
 - challenge_response

pcd1193182:
I've talked to @rasantel , the plan is to update CRA once the changes for this issue land.

pzakha:
bors delegate+

PR #343, opened 2019-08-02
add java to dcenter variant

grwilson:


grwilson:
[ab-pre-push](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1840/)

I have also deployed the dcenter group and verified that java is now installed.

prakashsurya:
Can you comment on why we’d decide on this approach, rather than adding java as a dependency of the platform package?

grwilson:
Today only the virtualization platform pulls java in as part of its build requirements and it doesn't get added in `delphix-platform`. We also have the `unittest-internal` which explicitly adds the jdk in`appliance-build`. Do we want to clean this up and just make `adoptopenjdk` a foundation package for all platforms?

prakashsurya:
I don’t think we have any plans to add java to the platform package, but I’m starting to think that we should, since we’ll now have two non-product variants that use it.

I’ll approve this as-is, and then maybe we can revisit this later.

PR #344, opened 2019-08-05
WIP: Use "delphix.target" to restart services on upgrade

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1858/)

pzakha:
Would this also attempt to restart `delphix-mgmt` inside a container?
Also, is this still a Work In Prgoress?

prakashsurya:
> Would this also attempt to restart delphix-mgmt inside a container?

No; "delphix-mgmt.service" doesn't run in the upgrade container.

> Also, is this still a Work In Prgoress?

Yes; I still need to test and do manual verification.

pzakha:
> No; "delphix-mgmt.service" doesn't run in the upgrade container.

Right, just wanted to make sure that this didn't affect that behaviour.
Might be useful to compare the output of running services before this change and after this change.

This LGTM, but I'll take another look once this is not WIP anymore.

prakashsurya:
> Might be useful to compare the output of running services before this change and after this change.

Can you be more specific about what you're requesting? I'm not sure exactly what information you want me to gather...

pzakha:
> Can you be more specific about what you're requesting? I'm not sure exactly what information you want me to gather...

Sorry, was a bit vague. We should just check that this restarts (or starts) all the Delphix services that we expect.

PR #345, opened 2019-08-06
DLPX-65387 Upgrade rebooting screen is no longer observed upgrading from 5.3.5 to L

pzakha:
Adds upgrade reboot screen to migration upgrades, using dx_execute code from 5.3.

## TESTING
- manually verified that the reboot screen appears in the UI.
- git-ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1889/

### DIFF
--- a/live-build/misc/migration-scripts/dx_execute
+++ b/live-build/misc/migration-scripts/dx_execute
@@ -164,6 +164,12 @@ MAIN_MENU_LINUX_CMD=$(grep -F 'mainmenu_command[8]' /boot/menu.rc.local |
 	cut -d = -f 2-)
 echo "set menu_timeout_command=$MAIN_MENU_LINUX_CMD" >>/boot/menu.rc.local
 
+#
+# Notify the UI that the system is rebooting.
+# This could hang if the JVM is in a bad state, so we use a timeout.
+#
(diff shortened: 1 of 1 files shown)


pzakha:
cc: @sebroy @sdimitro

grwilson:
Do we need similar logic in our linux upgrade scripts?

pzakha:
@grwilson, Not according to Michael, looking in the bug report.

> Appears to be specific to migration, since the correct screen is seen when upgrading L to L.

live-build/misc/migration-scripts/dx_execute
```diff
@@ -51,6 +51,23 @@ EOM
...
 }
 
+function upgrade_alert() {
+	local pid
+
+	#
+	# This could hang if the JVM is in a bad state, so we spawn it in the
+	# background and give it 5 seconds to complete before giving up.
+	#
+	/opt/delphix/server/bin/jmxtool boot upgrade server &>/dev/null &
```
grwilson: Can we just use `timeout` instead? It would simplify this logic.

pzakha: Yeah, much simpler indeed.

PR #346, opened 2019-08-07
DLPX-65345 Applying L early release migration image results in two upgrade versions being created

pzakha:
In the past the version of the `delphix-virtualization` package would be used to determine the version of the appliance. This was changed, but we forgot to update the logic when generating the migration image.

## Testing

- manually tested building a variant with `DELPHIX_APPLIANCE_VERSION` set and unset on a `bootstrap-18-04` system.
- ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1896/
- migration: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1895/
- Check that after performing a migration we do not see 2 versions anymore: **TODO**

### DIFF
--- a/live-build/build.gradle
+++ b/live-build/build.gradle
@@ -113,6 +113,17 @@ for (variant in allVariants) {
                     inputs.property(envVar, System.getenv(envVar)).optional(true)
                 }
 
+                if (!System.getenv('DELPHIX_APPLIANCE_VERSION')) {
+                    /*
+                     * If we are running the build manually and
+                     * DELPHIX_APPLIANCE_VERSION is not set explicitly,
+                     * just use an arbitrarily high value for the
+                     * appliance version to allow upgrading to the
+                     * produced image.
+                     */
+                    environment 'DELPHIX_APPLIANCE_VERSION', '99.9.9.9'
+                }
+
                 switch (runType) {
                     case upgradeArtifactsRunType:
                         outputs.file "${buildDir}/artifacts/${variant}-${platform}.debs.tar.gz"
--- a/live-build/config/hooks/configuration/81-upgrade-repository.binary
+++ b/live-build/config/hooks/configuration/81-upgrade-repository.binary
@@ -92,25 +92,9 @@ chroot binary dpkg-query -Wf '${Package}=${Version}\n' | sort >packages.list
 
 echo "$APPLIANCE_VARIANT" >variant
 
-if [[ -n "${DELPHIX_APPLIANCE_VERSION:-}" ]]; then
-	#
-	# If the build specified a version for the delphix-entire package,
-	# then use it.
-	#
-	VERSION=$DELPHIX_APPLIANCE_VERSION
-else
-	#
-	# Otherwise, just use a default value for the version, based on when
-	# the build occurs.
-	#
-	# Note that we want to generate the timestamp without any leading
-	# zeroes, because the Delphix API handles versions as integers.
-	# Leading zeroes will be stripped and cause mismatches.
-	#
-	VERSION=$(date '+%Y.%-m.%-d.%-H')
-fi
+test -n "$DELPHIX_APPLIANCE_VERSION"
 
-sed -i "s/@@VERSION@@/$VERSION/" delphix-entire.ctl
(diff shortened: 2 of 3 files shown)


pzakha:
cc: @jgallag88, @prakashsurya

pzakha:
Updated this code in preparation for the version schema changes.
Note that since `DELPHIX_APPLIANCE_VERSION` is always passed when built with Jenkins, the fictional version `99.9.9.9` will only be used when the image is build manually using the steps in the README.

## Retesting
- ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1909/
- migration: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1908/
- manual build

live-build/build.gradle
```diff
@@ -113,6 +113,19 @@ for (variant in allVariants) {
...
+                doFirst {
+                    if (!System.getenv('DELPHIX_APPLIANCE_VERSION')) {
+                        /*
+                         * If we are running the build manually and
+                         * DELPHIX_APPLIANCE_VERSION is not set explicitly,
+                         * just use an arbitrarily high value for the
+                         * appliance version to allow upgrading to the
+                         * produced image.
+                         */
+                        environment 'DELPHIX_APPLIANCE_VERSION', '99.9.9.9'
```
jgallag88: I think it should be possible to set this in the configuration phase instead of the execution phase, so I don't think this needs to be in a `doFirst` block.

live-build/config/hooks/configuration/81-upgrade-repository.binary
```diff
@@ -92,25 +92,9 @@ chroot binary dpkg-query -Wf '${Package}=${Version}\n' | sort >packages.list
...
-	# Otherwise, just use a default value for the version, based on when
-	# the build occurs.
-	#
-	# Note that we want to generate the timestamp without any leading
-	# zeroes, because the Delphix API handles versions as integers.
-	# Leading zeroes will be stripped and cause mismatches.
-	#
-	VERSION=$(date '+%Y.%-m.%-d.%-H')
-fi
+test -n "$DELPHIX_APPLIANCE_VERSION"
```
sebroy: We're not doing anything with the exit code of `test` here. Is there some magic at play that I'm not familiar with?

live-build/build.gradle
pzakha: Yes, you are probably right. I did this in the execution phase before because it depended on a task being executed. I'll fix that.

live-build/config/hooks/vm-artifacts/90-linux-migration-artifact.binary
```diff
@@ -85,6 +85,8 @@ mkdir $DEPOT_DIRECTORY
 #
 cp migration-scripts/* $DEPOT_DIRECTORY
 
+test -n "$DELPHIX_APPLIANCE_VERSION"
```
sebroy: Same here

live-build/config/hooks/configuration/81-upgrade-repository.binary
pzakha: Yeah, since we are running bash with `-e` set, test will return 1 and so the script will fail.

live-build/build.gradle
prakashsurya: Can we use the same version scheme as previously used, and what's used by the Jenkins logic? i.e. `"yyyy.M.d.H"`

pzakha: We can, that's what I did in the original review, but the code was more complex. The `"yyyy.M.d.H"` versioning will be deprecated after we transition to the new versioning schema, which will be some time next week.

prakashsurya: OK. I'm fine with this as-is. Can you point me to any documentation or discussion about this new versioning schema? I don't know anything about that.

pzakha:
verified that only a single version appears after applying the migration image.

PR #348, opened 2019-08-13
DLPX-64429 not-in-place upgrade should copy files related to cloud-init

pzakha:
- We copy all cloud-init directories that are related to cloud init cache, that is:
`/var/lib/cloud/{data, instance, instances, sem}`
We do not copy the whole `/var/lib/cloud` directory as it contains directories/files managed by other packages (e.g. `/var/lib/cloud/seed`)

- We copy the auto-generated network configuration:
`/etc/netplan/50-cloud-init.yaml`
Note that this file may not be present if the default network configuration was modified. In that case `/etc/netplan/10-delphix.yaml` and `/etc/cloud/cloud.cfg.d/99-disable-network-config.cfg` should be present instead.

- We copy the `/etc/hosts` template file:
`/etc/cloud/templates/hosts.debian.tmpl`
This is the file that is used by cloud-init to auto-generate `/etc/hosts`, so any changes that support wants to make to `/etc/hosts` should be made there or they will be lost.

## Testing
- ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1904/
- manually tested not-in-place upgrade and made sure that the system comes back up without any errors. Did a sanity check on cloud-init's `/var/lib/cloud`.

### DIFF
--- a/upgrade/upgrade-scripts/upgrade-container
+++ b/upgrade/upgrade-scripts/upgrade-container
@@ -581,6 +581,7 @@ function migrate_configuration() {
 	while read -r file; do
 		migrate_file "$file"
 	done <<-EOF
+		/etc/cloud/templates/hosts.debian.tmpl
 		/etc/default/nfs-kernel-server
 		/etc/hostid
 		/etc/hostname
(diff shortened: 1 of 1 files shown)


pzakha:
cc: @prakashsurya @grwilson @sdimitro

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -648,6 +649,28 @@ function migrate_configuration() {
 		/var/target/pr
 	EOF
 
+	#
+	# These directories and files are created by cloud-init. When
```
prakashsurya: Can you also explain, somewhere in this comment, why we don't want cloud-init to re-generate this information after a not-in-place upgrade.

After reading this comment, the PR description, and the bug report, I still don't understand why we can't rely on cloud-init to properly regenerate this information after the upgrade.

pzakha: I've added more comment.

PR #349, opened 2019-08-13
DLPX-65276 Create new dataset that will be shared across versions to store rollback and verify logs

vimalraj-anbarasu:
Removed the code where `/var/tmp/delphix-upgrade` folder is created and shared with container. 
This is not necessary anymore as going forward we will have that folder mounting the dataset which will be shared across versions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,9 @@
 .gradle/
 .gradleUserHome/
 build/
+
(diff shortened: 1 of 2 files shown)


prakashsurya:
There was some DCOA issue with the prior build... retrying...

bors r+

PR #351, opened 2019-08-16
Fixes for LX-1780 and DLPX-65365 related to migration

pzakha:
## Fixes
LX-1780 migration: snapshot should be taken before doing changes to MDS
DLPX-65365 After upgrading from 5.3.5 to L, the delphix user's home directory was wiped clean

## Changes
- Snapshots taken for migration are now called `@migration` to make it easier to clean them up later.
- Take a snapshot of MDS before reboot to make rollback easier/safer.
- Make a clone of the illumos /export/home dataset to make it easier for support to retrieve old scripts.

## Testing
- migration: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1932/
- After migration, manually test mounting `rpool/illumos-home`, and cleanup of the old illumos datasets after using `zfs promote`.

### DIFF
--- a/live-build/misc/migration-scripts/dx_execute
+++ b/live-build/misc/migration-scripts/dx_execute
@@ -51,6 +51,18 @@ EOM
 	exit 2
 }
 
+#
+# cleanup a leftover dataset from a previous run of dx_execute if needed.
+#
+function cleanup_leftover_dataset() {
+	local dataset=$1
+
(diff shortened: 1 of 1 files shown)


pzakha:
cc: @prakashsurya @sdimitro @shartse

PR #352, opened 2019-08-23
Require DELPHIX_SIGNATURE_VERSIONS to be passed in

prakashsurya:
This changes which keys are used when signing the upgrade image.

Previously we used a hardcoded value of 5.3, and thus we always signed
the upgrade image with the 5.3 key. Now, we require the user to specify
a space seperated list of versions, and we sign the upgrade image with
the key corresponding to each version in the list.

This is intended to work inconjunction with some Jenkins changes, where
Jenkins will build the correct list of signature versions, based the
specific version of the upgrade image being built.

### DIFF
--- a/scripts/build-upgrade-image.sh
+++ b/scripts/build-upgrade-image.sh
@@ -104,11 +104,10 @@ if [[ -n "${DELPHIX_SIGNATURE_TOKEN:-}" ]] && [[ -n "${DELPHIX_SIGNATURE_URL:-}"
 	#
 	# Here, we need to generate signature files for all of the appliance
 	# versions that'll be allowed to upgrade from, using this upgrade
-	# image. For now, we hardcode this to simply generate a signature for
-	# the "5.3" release, but eventually we'll need to update this code to
-	# support more versions (as more linux-based versions are released).
+	# image. We rely on the user of this script to pass in this list
+	# of versions; generally this will be some Jenkins automation.
 	#
-	for signature_version in "5.3"; do
+	for signature_version in $DELPHIX_SIGNATURE_VERSIONS; do
 		curl -s -S -f -H "Content-Type: application/json" \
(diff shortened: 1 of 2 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://collins.d.delphix.com:22812/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1/)

This depends on the following reviews: [devops-gate](http://reviews.delphix.com/r/51995/), [pipeline-shared](http://reviews.delphix.com/r/51990/)

prakashsurya:
cc: @basil @prashks

scripts/build-upgrade-image.sh
```diff
@@ -104,11 +104,10 @@ if [[ -n "${DELPHIX_SIGNATURE_TOKEN:-}" ]] && [[ -n "${DELPHIX_SIGNATURE_URL:-}"
...
 	# Here, we need to generate signature files for all of the appliance
 	# versions that'll be allowed to upgrade from, using this upgrade
-	# image. For now, we hardcode this to simply generate a signature for
-	# the "5.3" release, but eventually we'll need to update this code to
-	# support more versions (as more linux-based versions are released).
+	# image. We rely on the user of this script to pass in this list
+	# of versions; generally this will be some Jenkins automation.
 	#
-	for signature_version in "5.3"; do
+	for signature_version in ${DELPHIX_SIGNATURE_VERSIONS:-""}; do
```
pzakha: if `DELPHIX_SIGNATURE_VERSIONS` is unset, then the loop will iterate once with `signature_version=""`, which it doesn't look like it will have the desired outcome.

Why not just use `${DELPHIX_SIGNATURE_VERSIONS}`?

prakashsurya: Yes, I think you're right. I'll do that.

basil:
bors delegate+

PR #358, opened 2019-09-03
Use compression on rpool by default

pzakha:
Turning on compression on rpool makes sure it is inherited
by all the other datasets, so we do not need to set it
explicitly on each dataset we create.

## Motivation
Simplify our ZFS setup and reduce differences between Linux native and migrated systems.

## Testing
- ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2022/
- migration: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2023/
- Manually verified that compression was set on all datasets on a fresh Linux system and after a migration.


pzakha:
cc: @prakashsurya @sdimitro @shartse @grwilson

PR #360, opened 2019-09-04
Don't use "-q" when building Docker image

prakashsurya:
We've been seeing timeouts when building the Docker container used with
our TravisCI testing. If a command doesn't produce any output for 10
minutes, TravisCI will fail the build the command is running in. Thus,
if "docker build" takes longer than 10 minutes to build the image, our
builds would fail because we had been using the "-q" option when
building the image.

This change removes the "-q" option so that the "docker build" command
will output text as it builds the image, and this output will keep
TravisCI happy even if it takes more than 10 minutes to build the image.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ env:
   - TARGET=shfmtCheck
 
 install:
- - docker build -qt appliance-build:latest docker
+ - docker build -t appliance-build:latest docker
 
 script:
  - ./scripts/docker-run.sh gradle $TARGET


prakashsurya:
> I'm curious as to how did you debug that issue

The build log ([here](https://travis-ci.com/delphix/appliance-build/jobs/231122583)) shows this:
```
$ docker build -qt appliance-build:latest docker

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

The build has been terminated
```

I retried the build multiple times, and hit this same issue; additionally, some builds worked just fine. So, I'm guessing that "docker build" is proceeding normally, but it's just taking longer than 10 minutes in some cases. I also know that since we're using "-q" no output will be emitted while it runs; the only output line would be after it completes.

PR #363, opened 2019-09-17
DLPX-65501 uploading same upgrade image twice has "unexpected" error

vimalraj-anbarasu:
Instead of tags unpack script will accept different exit codes that will be translated in app-stack.
Corresponding app-gate change: http://reviews.delphix.com/r/52590

ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2193/

### DIFF
--- a/upgrade/prepare
+++ b/upgrade/prepare
@@ -18,8 +18,15 @@
 UPDATE_DIR=${UPDATE_DIR:-/var/dlpx-update}
 
 function die() {
+	exit_code=$1
+	# Use first argument as exit code, if provided.
+	if [[ -z "${exit_code//[0-9]/}" ]]; then
+		shift
+	else
+		exit_code=1
+	fi
 	echo "$(basename "$0"): $*" >&2
-	exit 1
+	exit ${exit_code}
 }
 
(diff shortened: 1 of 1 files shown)


upgrade/prepare
```diff
@@ -18,8 +18,11 @@
 UPDATE_DIR=${UPDATE_DIR:-/var/dlpx-update}
 
 function die() {
+	exit_code=$1
+	[[ -z "${exit_code//[0-9]/}" ]] && [[ -n "$exit_code" ]] || exit_code=1
```
pzakha: What you probably want is something like:
```
function die() {
        exit_code=$1
        # Use first argument as exit code, if provided.
        if [[ -z "${exit_code//[0-9]/}" ]]; then
                shift
        else
                exit_code=1
        fi
        echo "$(basename "$0"): $*" >&2
        exit ${exit_code}
}
```
This way if there is no number provided as first argument, it will print the whole string and exit with `1`

vimalraj-anbarasu: Done.

PR #365, opened 2019-09-20
LB-209 Want to consume new mirror for production appliance build

grodr:
Add functionality to appliance-build to infer the latest linux package mirror dataset for builds if no mirror url is passed in. 

Testing:
Ran git ab-pre-push 
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2212/parameters/

### DIFF
--- a/branch.config
+++ b/branch.config
@@ -1,0 +1,13 @@
+#
+# Copyright (c) 2019 by Delphix. All rights reserved.
+#
+
+#
+# The "BRANCH" parameter tracks the upstream branch of appliance-build. It is
+# used to determine which branch of the linux package mirror will be used for
+# the build if UPSTREAM_PRODUCT_BRANCH is not set. UPSTREAM_PRODUCT_BRANCH is
+# set when appliance build is built by the appliance-build Jenkins job. The
+# UPSTREAM_BRANCH parameter should be updated by the release lead on branching
+#
+
+UPSTREAM_BRANCH="master"
--- a/scripts/build-ancillary-repository.sh
+++ b/scripts/build-ancillary-repository.sh
@@ -142,27 +142,30 @@ function build_ancillary_repository() {
 #    environment variables, and the script will work as expected.
 #
 
-upstream_branch="${UPSTREAM_PRODUCT_BRANCH:-master}"
+if [[ -z "$UPSTREAM_BRANCH" ]]; then
+	echo "UPSTREAM_BRANCH is not set."
+	exit 1
+fi
 
 AWS_S3_URI_VIRTUALIZATION=$(resolve_s3_uri \
 	"$AWS_S3_URI_VIRTUALIZATION" \
 	"$AWS_S3_PREFIX_VIRTUALIZATION" \
-	"dlpx-app-gate/${upstream_branch}/build-package/post-push/latest")
+	"dlpx-app-gate/${UPSTREAM_BRANCH}/build-package/post-push/latest")
 
 AWS_S3_URI_MASKING=$(resolve_s3_uri \
 	"$AWS_S3_URI_MASKING" \
 	"$AWS_S3_PREFIX_MASKING" \
-	"dms-core-gate/${upstream_branch}/build-package/post-push/latest")
+	"dms-core-gate/${UPSTREAM_BRANCH}/build-package/post-push/latest")
 
 AWS_S3_URI_USERLAND_PKGS=$(resolve_s3_uri \
 	"$AWS_S3_URI_USERLAND_PKGS" \
 	"$AWS_S3_PREFIX_USERLAND_PKGS" \
-	"devops-gate/master/linux-pkg-build/${upstream_branch}/userland/post-push/latest")
+	"devops-gate/master/linux-pkg-build/${UPSTREAM_BRANCH}/userland/post-push/latest")
 
 AWS_S3_URI_KERNEL_PKGS=$(resolve_s3_uri \
 	"$AWS_S3_URI_KERNEL_PKGS" \
 	"$AWS_S3_PREFIX_KERNEL_PKGS" \
-	"devops-gate/master/linux-pkg-build/${upstream_branch}/kernel/post-push/latest")
+	"devops-gate/master/linux-pkg-build/${UPSTREAM_BRANCH}/kernel/post-push/latest")
 
 #
 # All package files will be placed into this temporary directory, such
--- a/scripts/docker-run.sh
+++ b/scripts/docker-run.sh
@@ -38,6 +38,29 @@ else
 	DOCKER_RUN="docker run"
 fi
 
+#
+# Set UPSTREAM_BRANCH. This will determine which version of the linux package
+# mirror is used.
+#
+if [[ -z "$UPSTREAM_PRODUCT_BRANCH" ]]; then
+	echo "UPSTREAM_PRODUCT_BRANCH is not set."
+	if ! source "$TOP/branch.config" 2>/dev/null; then
+		echo "No branch.config file found in repo root."
+		exit 1
+	fi
+
+	if [[ -z "$UPSTREAM_BRANCH" ]]; then
+		echo "UPSTREAM_BRANCH parameter was not sourced from branch.config." \
+			"Ensure branch.config is properly formatted with e.g." \
+			"UPSTREAM_BRANCH=\"<upstream-product-branch>\""
+		exit 1
+	fi
+	echo "Defaulting to branch $UPSTREAM_BRANCH set in branch.config."
+else
+	UPSTREAM_BRANCH="$UPSTREAM_PRODUCT_BRANCH"
+fi
+echo "Running with UPSTREAM_BRANCH set to ${UPSTREAM_BRANCH}"
+
 $DOCKER_RUN --rm \
 	--privileged \
 	--network host \
@@ -64,6 +87,7 @@ $DOCKER_RUN --rm \
 	--env DELPHIX_SIGNATURE_VERSIONS \
 	--env DELPHIX_UPGRADE_MINIMUM_VERSION \
 	--env DELPHIX_UPGRADE_MINIMUM_REBOOT_OPTIONAL_VERSION \
+	--env UPSTREAM_BRANCH="$UPSTREAM_BRANCH" \
 	--volume "$TOP:/opt/appliance-build" \
 	--workdir "/opt/appliance-build" \
 	appliance-build "$@"
--- a/scripts/run-live-build.sh
+++ b/scripts/run-live-build.sh
@@ -127,37 +127,72 @@ while ! curl --output /dev/null --silent --head --fail \
 
 	sleep 1
 done
+
+pkg_mirror_secondary=''
+if [[ -n "$DELPHIX_PACKAGE_MIRROR_SECONDARY" ]]; then
+	pkg_mirror_secondary="$DELPHIX_PACKAGE_MIRROR_SECONDARY"
+else
+	#
+	# If no secondary package mirror is provided, then pull in the latest
+	# mirror dataset for the build. If no latest dataset is found, then fail.
+	#
+	source_url="http://linux-package-mirror.delphix.com/$UPSTREAM_BRANCH/latest/"
+	if ! pkg_mirror_secondary=$(curl -fLSs -o /dev/null -w '%{url_effective}' "$source_url"); then
+		kill -9 $APTLY_SERVE_PID
+		echo "No URL found for PPA packages at ${source_url}."
+		exit 1
+	fi
+
+	# The mirror hosts secondary packages in the "ppas" subdirectory.
+	pkg_mirror_secondary+="ppas"
+fi
+
+sed "s|@@URL@@|$pkg_mirror_secondary|" \
+	<config/archives/delphix-secondary-mirror.list.in \
+	>config/archives/delphix-secondary-mirror.list
+
+pkg_mirror_main=''
+if [[ -n "$DELPHIX_PACKAGE_MIRROR_MAIN" ]]; then
+	pkg_mirror_main="$DELPHIX_PACKAGE_MIRROR_MAIN"
+else
+	#
+	# If no main package mirror is provided, then pull in the latest mirror
+	# dataset for the build. If no latest dataset is found, then fail.
+	#
+	source_url="http://linux-package-mirror.delphix.com/$UPSTREAM_BRANCH/latest/"
+	if ! pkg_mirror_main=$(curl -fLSs -o /dev/null -w '%{url_effective}' "$source_url"); then
+		kill -9 $APTLY_SERVE_PID
+		echo "No mirror URL found for ubuntu archive packages at ${source_url}."
+		exit 1
+	fi
+
+	#
+	# The internal mirror hosts the primary ubuntu package repository in the
+	# "ubuntu" subdirectory.
+	#
+	pkg_mirror_main+="ubuntu"
+fi
 set -o errexit
 
-if [[ -n "$DELPHIX_PACKAGE_MIRROR_SECONDARY" ]]; then
-	sed "s|@@URL@@|$DELPHIX_PACKAGE_MIRROR_SECONDARY|" \
-		<config/archives/delphix-secondary-mirror.list.in \
-		>config/archives/delphix-secondary-mirror.list
-fi
-
-if [[ -n "$DELPHIX_PACKAGE_MIRROR_MAIN" ]]; then
-	lb config \
-		--parent-mirror-bootstrap "$DELPHIX_PACKAGE_MIRROR_MAIN" \
-		--parent-mirror-chroot "$DELPHIX_PACKAGE_MIRROR_MAIN" \
-		--parent-mirror-chroot-security "$DELPHIX_PACKAGE_MIRROR_MAIN" \
-		--parent-mirror-chroot-volatile "$DELPHIX_PACKAGE_MIRROR_MAIN" \
-		--parent-mirror-chroot-backports "$DELPHIX_PACKAGE_MIRROR_MAIN" \
-		--parent-mirror-binary "$DELPHIX_PACKAGE_MIRROR_MAIN" \
-		--parent-mirror-binary-security "$DELPHIX_PACKAGE_MIRROR_MAIN" \
-		--parent-mirror-binary-volatile "$DELPHIX_PACKAGE_MIRROR_MAIN" \
-		--parent-mirror-binary-backports "$DELPHIX_PACKAGE_MIRROR_MAIN" \
-		--mirror-bootstrap "$DELPHIX_PACKAGE_MIRROR_MAIN" \
-		--mirror-chroot "$DELPHIX_PACKAGE_MIRROR_MAIN" \
-		--mirror-chroot-security "$DELPHIX_PACKAGE_MIRROR_MAIN" \
-		--mirror-chroot-volatile "$DELPHIX_PACKAGE_MIRROR_MAIN" \
-		--mirror-chroot-backports "$DELPHIX_PACKAGE_MIRROR_MAIN" \
-		--mirror-binary "$DELPHIX_PACKAGE_MIRROR_MAIN" \
-		--mirror-binary-security "$DELPHIX_PACKAGE_MIRROR_MAIN" \
-		--mirror-binary-volatile "$DELPHIX_PACKAGE_MIRROR_MAIN" \
-		--mirror-binary-backports "$DELPHIX_PACKAGE_MIRROR_MAIN"
-else
-	lb config
-fi
+lb config \
+	--parent-mirror-bootstrap "$pkg_mirror_main" \
+	--parent-mirror-chroot "$pkg_mirror_main" \
+	--parent-mirror-chroot-security "$pkg_mirror_main" \
+	--parent-mirror-chroot-volatile "$pkg_mirror_main" \
+	--parent-mirror-chroot-backports "$pkg_mirror_main" \
+	--parent-mirror-binary "$pkg_mirror_main" \
+	--parent-mirror-binary-security "$pkg_mirror_main" \
+	--parent-mirror-binary-volatile "$pkg_mirror_main" \
+	--parent-mirror-binary-backports "$pkg_mirror_main" \
+	--mirror-bootstrap "$pkg_mirror_main" \
+	--mirror-chroot "$pkg_mirror_main" \
+	--mirror-chroot-security "$pkg_mirror_main" \
+	--mirror-chroot-volatile "$pkg_mirror_main" \
+	--mirror-chroot-backports "$pkg_mirror_main" \
+	--mirror-binary "$pkg_mirror_main" \
+	--mirror-binary-security "$pkg_mirror_main" \
+	--mirror-binary-volatile "$pkg_mirror_main" \
+	--mirror-binary-backports "$pkg_mirror_main"
 
 lb build
 


scripts/run-live-build.sh
```diff
@@ -129,36 +129,64 @@ while ! curl --output /dev/null --silent --head --fail \
...
-	sed "s|@@URL@@|$DELPHIX_PACKAGE_MIRROR_SECONDARY|" \
-		<config/archives/delphix-secondary-mirror.list.in \
-		>config/archives/delphix-secondary-mirror.list
+    pkg_mirror_secondary="$DELPHIX_PACKAGE_MIRROR_SECONDARY"
+else
+    #
+    # If no secondary package mirror is provided, then pull in the latest
+    # mirror dataset for the build. If no latest dataset is found, then fail.
+    #
+    source_url="http://linux-package-mirror.delphix.com/$UPSTREAM_PRODUCT_BRANCH/latest/"
```
sebroy: Similar the comment I left in the `linux-pkg` review, I wonder if it would be cleaner to keep the Delphix infrastructure-specific details of which mirror URL gets used when building a particular branch to the Delphix-specific jenkins jobs that could then pass in the necessary inputs into this code.

grodr: I wanted this logic at the appliance-build and linux-pkg level rather than the jenkins level since these workflows can be ran locally without jenkins, so I wanted to support that workflow. I think it would be ideal if the local workflow is as close to prod as possible. However, this logic could potentially be done at the jenkins. Thoughts?

pzakha: One down-side I see of doing it this way is that it makes manual builds without Jenkins harder as we'd have to provide this argument manually. This is especially true for Linux-pkg which is built manually much more often.

sebroy: I can understand that argument. At the same time, we'd be hard coding bits of Delphix-internal infrastructure details within the business logic, which isn't ideal (especially in a public open-source code repo). It's your call, I don't feel that strongly about it.

scripts/run-live-build.sh
```diff
@@ -129,36 +129,64 @@ while ! curl --output /dev/null --silent --head --fail \
...
-		<config/archives/delphix-secondary-mirror.list.in \
-		>config/archives/delphix-secondary-mirror.list
+	pkg_mirror_secondary="$DELPHIX_PACKAGE_MIRROR_SECONDARY"
+else
+	#
+	# If no secondary package mirror is provided, then pull in the latest
+	# mirror dataset for the build. If no latest dataset is found, then fail.
+	#
+	source_url="http://linux-package-mirror.delphix.com/$UPSTREAM_PRODUCT_BRANCH/latest/"
+	pkg_mirror_secondary="$(curl -Ls -o /dev/null -w '%{url_effective}' $source_url)ppas"
```
prakashsurya: Can you explain the design for this, and/or point me to any existing written documentation for how this is supposed to work? What is the "public interface" that we're consuming here? I'm not questioning the validity, I would just like to better understand the full solution; e.g. how is this "service" managed, how will this be updated, what is the "interface" that we're adhering too, etc.

Is there a design doc that I can look over?

scripts/run-live-build.sh
```diff
@@ -129,36 +129,64 @@ while ! curl --output /dev/null --silent --head --fail \
...
-	sed "s|@@URL@@|$DELPHIX_PACKAGE_MIRROR_SECONDARY|" \
-		<config/archives/delphix-secondary-mirror.list.in \
-		>config/archives/delphix-secondary-mirror.list
+	pkg_mirror_secondary="$DELPHIX_PACKAGE_MIRROR_SECONDARY"
+else
+	#
+	# If no secondary package mirror is provided, then pull in the latest
+	# mirror dataset for the build. If no latest dataset is found, then fail.
+	#
+	source_url="http://linux-package-mirror.delphix.com/$UPSTREAM_PRODUCT_BRANCH/latest/"
```
prakashsurya: How will the `UPSTREAM_PRODUCE_BRANCH` be specified?

scripts/run-live-build.sh
```diff
@@ -129,36 +129,64 @@ while ! curl --output /dev/null --silent --head --fail \
...
+	pkg_mirror_secondary="$(curl -Ls -o /dev/null -w '%{url_effective}' $source_url)ppas"
+	if [[ "$pkg_mirror_secondary" == "${source_url}ppas" ]]; then
```
prakashsurya: Under what circumstances may this condition return true? e.g. what guarantees do we have about what the above `curl` command will return; when will it return the "correct" value, and when may it return an "incorrect" value? Is the only way to determine "correct vs. incorrect" by doing a comparison like this, with the original "source url"?

scripts/run-live-build.sh
basil: As an operator, I'm biased against hard-coding bits of Delphix-internal infrastructure details within the business logic. But as a developer, I understand that overly onerous environmental setup requirements inhibit code changes. Another option is to have this type of logic in both places: passed-in from `devops-gate` in the general case (which is most friendly to operators) while still making a best effort to infer it outside of `devops-gate` when it is not provided (which is friendly to developers). The more I think about it, the more I'm a fan of this option. What do others think?

pzakha: We already have similar logic in places like [this](https://github.com/delphix/appliance-build/blob/master/scripts/build-ancillary-repository.sh#L45), where we hardcode Delphix-internal infrastructure. I personally don't think that there is sensitive info in there, but adding this will not make things much more coupled to Delphix infrastructure than they are already.

pzakha: I'm fine with @basil's proposal. We already do something like this for the S3 urls in the example I provided above.

prakashsurya: > At the same time, we'd be hard coding bits of Delphix-internal infrastructure details within the business logic

@sebroy we already do that for certain parts of this repository; e.g. how we download our S3 package artifacts.

Unless we want to couple the build more tightly with Jenkins (e.g. only support building via Jenkins), then this is inevitable due to how we manage some of the things this build depends on. As a design principle, should we only support building via Jenkins? or should we support building outside of Jenkins, and take the approach that Jenkins is just one of the many potential users of the build?

prakashsurya: > Another option is to have this type of logic in both places: passed-in from devops-gate in the general case (which is most friendly to operators) while still making a best effort to infer it outside of devops-gate when it is not provided (which is friendly to developers).

@basil can you explain how this is different than what we've previously done, and what I think this change is doing? I had thought this was already the approach we should take, in general.

scripts/run-live-build.sh
grodr: We have nginx running a web server on our internal package mirror. On every successful update to the mirror we update an nginx config file to redirect  `http://linux-package-mirror.delphix.com/<delphix-branch>/latest/` to the latest updated mirror dataset, which can be dereferenced with a curl command as seen here. There is a redirect rule per branch. This mirror update and rewrite change are both done through this jenkins job `http://ops.jenkins.delphix.com/job/devops-gate/job/master/job/linux-package-mirror-sync/job/master/job/mirror1-sync/` (which is currently disabled to freeze the updates). The rewrite rule update in the job is here `https://gitlab.delphix.com/devops/devops-gate/blob/master/jenkins/jobs/pipelines/linux_package_mirror_sync.groovy#L402`. I am not sure what you mean by "public interface", let me know if that answers your questions, or if I can clarify more.

We do not have a design doc yet, but there should be one soon.

scripts/run-live-build.sh
grodr: This is set by the environment variable and should be the corresponding Delphix branch, e.g. `master` or `6.0/stage`.

pzakha: @grodr Coming back to the issue of building manually, we should probably use a similar logic in here as we do in the other place that consumes UPSTREAM_PRODUCT_BRANCH: https://github.com/delphix/appliance-build/blob/master/scripts/build-ancillary-repository.sh#L145.

prakashsurya: > This is set by the environment variable

Where does this environment variable get set?

scripts/run-live-build.sh
prakashsurya: Thanks for the details.

After the "latest" redirect has changed, how long will the previous URL work? I presume, if it changes in the middle of a running build, the old build URL(s) will still work, right? For how long will they work? and what will be the cadence for updating the redirect?

scripts/run-live-build.sh
prakashsurya: > This is set by the environment variable

Where does this environment variable get set?

re:
```
upstream_branch="${UPSTREAM_PRODUCT_BRANCH:-master}"
```
This code will break once we branch, so I don't think this is something we should duplicate.

pzakha: The alternative is to fail the manual build if this is not provided.

prakashsurya: @pzakha I'm sure I can work with @grodr to come up with a solution that works for our requirements; which is supporting our branching model, and enables manual builds without having to manually specify this variable.

With that said, I don't know if this has already been solved or not, which is why I was trying to ask @grodr for more details.

scripts/run-live-build.sh
grodr: Since I am using the -s (--silent) flag (since I just want the resulting URL) this returns `true` if the server responds to the HEAD request. Looking into this more, I think I should add the `--show-error` option as well so that the curl will fail if the URL doesn't exist. Then I won't have to do the comparison, and can just check the error message of the curl call. I will update to do that.

scripts/run-live-build.sh
grodr: Since we are dereferencing the redirect and we are not removing the previous dataset, the previous URL will continue to work indefinitely at the moment. We will eventually add clean up for old datasets, the plan being to remove old datasets after a few days so that we do not break any in flight builds.

scripts/run-live-build.sh
basil: Previously we placed direct references to `npm.delphix.com` or `artifactory.delphix.com` in builds. In fact, we have over 5,0000 such references in `dlpx-app-gate` today:

```
[basil:~/work/app/dlpx-app-gate]$ git grep artifactory.delphix.com | wc -l
5642
```

This is friendly to developers but not operators. This PR (as it stands today) continues the same pattern. I am now proposing that we adopt a two-pronged approach where such operational concerns are centralized in `devops-gate` while still making a best effort to infer a sensible value outside of `devops-gate` for convenience during interactive development.

prakashsurya: In the proposed two-pronged approach (for this particular case), what would the devops-gate code look like? would it effectively be duplicating the logic that's in this PR, of interacting with the "latest" redirect? or something different?

I'm not necessarily against a two-pronged approach in general, but I'm not sure I completely understand the proposed solution without seeing code changes implementing the suggestion.

basil: > In the proposed two-pronged approach (for this particular case), what would the devops-gate code look like? would it effectively be duplicating the logic that's in this PR, of interacting with the "latest" redirect?

Yeah, that's exactly what I'm proposing.

prakashsurya: What is the benefit of duplicating that logic in two repositories?

basil: The benefit of keeping the main copy of the logic in `devops-gate` is that it is the most friendly to operators. For example, if the mirror is renamed or the URL scheme changes, the logic in `devops-gate` could be updated and that update would apply to all product branches (including older branches and release tags). If this logic wasn't centralized in `devops-gate`, then there would be increased burden on operators, who would have to maintain the old API contract (e.g. preserving old DNS names and URL schemes) or risk completely breaking builds.

The above is an argument for moving the logic to `devops-gate`, but not for keeping a second copy in product repositories. The argument for doing that _in addition to_ keeping the main copy in `devops-gate` is to facilitate more convenienct interactive builds by making a best effort to infer whatever can be inferred. This best effort should always succeed as long as there are no operational changes. If there are operational changes in the future (such as a change in the DNS name or URL scheme), this best effort might fail (most likely on older branches or tags that wouldn't be updated). But this wouldn't matter so much because the best effort was only intended for convenience, and the primary logic (in `devops-gate`) wouldn't fail.

prakashsurya: > the logic in devops-gate could be updated and that update would apply to all product branches (including older branches and release tags).

Ah, right; I wasn't thinking about older releases, which can't be updated as we make changes. With that in mind, I definitely agree.

grodr: Taking the above into account - would it make sense to add this change here, but remove the `exit` if no latest URL is found. Then also add this logic in the devops-gate?

scripts/run-live-build.sh
grodr: It is only set in the Jenkins job. So this won't actually work for a developer workflow without that set, I didn't realize that.

grodr: We are planning on branching appliance-build at the same cadence as the product though right? So could we use the appliance-build branch for this?

prakashsurya: > We are planning on branching appliance-build at the same cadence as the product though right?

That's my understanding, yes.

> So could we use the appliance-build branch for this?

Given any "local" branch or commit, how would you determine the "product branch" from that? e.g. will this work if we're building for a project branch? or one a clone without all git history and branch information?

That general idea is correct, but I'd have to think more about the implementation of it, to know if it'd work for all cases that might arise.

pzakha: Generally, trying to figure out the product branch from a git tree is not particularly trivial logic. One thing we could do is to keep defaulting to "master", but when a build starts, we should display the variables that are normally set by Jenkins so that the user is aware of what is being used.

prakashsurya: > One thing we could do is to keep defaulting to "master",

How will this get changed/updated/fixed for each release branch? Are you proposing we use "master" as the default for non-master branches?

> but when a build starts, we should display the variables that are normally set by Jenkins so that the user is aware of what is being used.

How do we get the values that are normally set by Jenkins when building outside of Jenkins?

pzakha: With my proposal, building manually will default to master on non-master branches as well, which will naturally not do the expected thing. One alternative would be to not use any default, and instead put some checks in the topmost `build.gradle` that will fail if some variables are not set and give a suggestion to resolve this. This would mean a few extra steps for building manually but would avoid unexpected behaviour when building older product branches.

grodr: What if we add a `BRANCH` variable that represents the appliance-build upstream branch. That variable can be updated by the release lead on branching, and pulled in to determine which branch we should pull in from the mirror.

prakashsurya: > which will naturally not do the expected thing.

If we can't do the correct thing, I'd much rather explicitly fail with an error, than silently do the wrong thing; this sounds like the worst way to handle this issue, IMO.

prakashsurya: Before suggesting any possible solutions, I need to better understand the scenarios that are most common to occur.

E.g. I wonder if we can have automatic product release branch resolution for most cases (e.g. a "normal" git clone operation when building manually), but then throw an error when that fails, and rely on the user to specific the environment variable correctly on failure.

I might not have time to investigate this immediately, though.

pzakha: I'd propose to go for this approach for now:
> One alternative would be to not use any default, and instead put some checks in the topmost build.gradle that will fail if some variables are not set and give a suggestion to resolve this. This would mean a few extra steps for building manually but would avoid unexpected behaviour when building older product branches.

This should be easy to implement and doesn't cause any unexpected behaviour.

If we want to add more means to automatically determine the product branch, then we can do that later.

grodr: @pzakha Any reason not to add the default branch as a variable, and have it updated on branching per branch?

pzakha: This is one more manual step that we'd need to do each time we are branching. Given that this variable would not be consumed by the regular code paths (Jenkins will provide it normally), I don't want to impose extra steps into the release process.

prakashsurya: If we force the user to pass in the correct value, how will we communicate to them which value to use (and when)?

pzakha: "Error, UPSTREAM_PRODUCT_BRANCH is empty. Please provide the upstream branch for the version of Delphix you are building (e.g. master, 6.0/stage) by running 'export UPSTREAM_PRODUCT_BRANCH=<value>".

prakashsurya: I don't think it's reasonable to assume folks will know what value to use, based only on that error message.

Sure, the folks in the PR thread might know what to set it to, but we're "experts"; I wouldn't expect a user from another team to know, without any other information.

As a design principle, folks from other teams should be able to build this by just looking at the instructions from the README. I also don't think it's good design to communicate build requirements and input parameters in the form of error messages.

pzakha: I think folks should know which version of the product they are building, and you can add this to the README... In any case, I feel like we are overthinking this right now. Manual build is mainly for doing changes to appliance-build, which would most likely involve the platform/devops team anyway. If you guys wanna implement some other ideas, feel free to go ahead with it, I'm not gonna block this. However, if you want to add steps to the release process, make sure to inform the release team.

prakashsurya: > I think folks should know which version of the product they are building, and you can add this to the README...

The onus to update the README isn't on me, it's on the person making the change to add the new requirement.

> if you want to add steps to the release process

I don't know if you're referring to me specifically here, but to be clear, I never suggested making any changes to the release process.

pzakha: By "you" I was referring to "whoever is working" on this change, not to you personally. In this case, this would be Garret.

grodr: I will be sure that the release lead is informed of the change. I think adding the BRANCH variable would solve the problems that have been raised.
I could also add something in the README if we think it is necessary. It definitely would be if we require environment variables be set. I am not sure if it needs to be updated if I am just adding the BRANCH variable.

prakashsurya: > I am not sure if it needs to be updated if I am just adding the BRANCH variable.

If the value has a proper default value, then I don't see any reason for a user to change it, so there's no need to discuss it in the build instruction in the README.

With that said, as an alternative to a hard-coded value, I think it might be reasonable to rely on the "upstream tracking branch" for this information. We already rely on this for some of our "git-utils" commands, so it's generally a requirement we have for other repositories and workflows. Additionally, it should "just work" give the build instructions in the README.

For example, we could use something like:
```
$ git for-each-ref --format='%(upstream:short)' "$(git symbolic-ref -q HEAD)"
```

Here's an example of what this returns for all of our current branches:
```
$ for branch in master staging trying projects/6.0-ea; do
> git clone https://github.com/delphix/appliance-build.git --branch "$branch" &>/dev/null
> cd appliance-build
> git for-each-ref --format='%(upstream:short)' "$(git symbolic-ref -q HEAD)"
> cd ..
> rm -rf appliance-build
> done
origin/master
origin/staging
origin/trying
origin/projects/6.0-ea
```

Obviously this wouldn't work when the "upstream tracking branch" isn't set; but in that case, I'd be fine throwing an error because I doubt folks will actually run into that problem.

Additionally, for the Jenkins case where it might not have a "upstream tracking branch", we'll already be passing that value into the build, rather than relying on this "best effort" logic. I also think this falls squarely into the two-pronged design principle that Basil mentioned before; i.e. Jenkins (or any consumer) can directly specify the variable to use when it knows best, and when it's not specified, we attempt to infer the correct value using the "upstream tracking branch" information.

Would this address all cases, without having to have a hard coded value that needs to be updated for each release branch, and without forcing users to manually specify the value?

scripts/docker-run.sh
```diff
@@ -64,6 +64,7 @@ $DOCKER_RUN --rm \
 	--env DELPHIX_SIGNATURE_VERSIONS \
 	--env DELPHIX_UPGRADE_MINIMUM_VERSION \
 	--env DELPHIX_UPGRADE_MINIMUM_REBOOT_OPTIONAL_VERSION \
+	--env UPSTREAM_PRODUCT_BRANCH \
```
prakashsurya: Are you still planning to use this variable?

Based on your recent changes to `run-live-build.sh`, this isn't being used.

scripts/run-live-build.sh
```diff
@@ -15,6 +15,12 @@
 # limitations under the License.
 #
 
+#
+# The appliance-build upstream branch. This should be updated by the release
+# lead on branching.
+#
+BRANCH="master"
```
prakashsurya: We also will need to update the line in `build-ancillary-repository.sh`.. I'd be great to consolidate both uses into a single place, if possible.. if not, please also update any release docs to also reference that other script (or ask Pavel to do that, since he originally added that other line), so that it gets updated in addition to this line.

scripts/run-live-build.sh
pzakha: I'm OK with that. So the logic would be:
1. Check if `UPSTREAM_PRODUCT_BRANCH` is set
2. If not, try to get it from the git upstream tracking branch (making sure to sanitize it properly).
3. If the above fails, ask user to either set upstream branch or to manually set UPSTREAM_PRODUCT_BRANCH.

There's still a chance that the user will set the tracking upstream branch to a branch that isn't a product branch (e.g. some project branch), but that's the best we can do without manually entering the branch. It might be useful to display the value UPSTREAM_PRODUCT_BRANCH when starting the build.

grodr: This seems fine to me, especially if we require users to set an upstream branch in their set up. I will work on implementing this.

prakashsurya: After thinking about this a little more, this may not work exactly like I was thinking, since this repository will be on the bootstrap system, rather than on a developer's workstation.

A workflow I use often is to clone my personal git repository onto the bootstrap system, and use a feature branch to build/test. In that example, the upstream tracking branch will be my personal git repository and feature branch name (which won't match our product release branch name). So, for this workflow, I don't think the automatic branch name detection using the upstream tracking branch will work.

With that said, I'm OK with any of the possible designs proposed so far. I'm having a hard time deciding which approach I prefer, since they all come with unique benefits and downsides, so I'm OK with whatever y'all think is best.

scripts/docker-run.sh
```diff
@@ -38,6 +38,27 @@ else
 	DOCKER_RUN="docker run"
 fi
 
+#
+# Ensure that either UPSTREAM_PRODUCT_BRANCH or DELPHIX_PACKAGE_MIRROR_MAIN
+# is set. If neither is set, attempt to use git to determine the upstream
+# branch. If that also fails, then exit.
+#
+UPSTREAM_BRANCH="$UPSTREAM_PRODUCT_BRANCH"
```
pzakha: Note that we are also using `UPSTREAM_PRODUCT_BRANCH` in `scripts/build-ancillary-repository.sh`. I think it would be better to set a default variable for the same variable if it is empty instead of creating a new variable.

scripts/docker-run.sh
```diff
@@ -38,6 +38,27 @@ else
 	DOCKER_RUN="docker run"
 fi
 
+#
+# Ensure that either UPSTREAM_PRODUCT_BRANCH or DELPHIX_PACKAGE_MIRROR_MAIN
```
pzakha: This comment refers to DELPHIX_PACKAGE_MIRROR_MAIN but I don't see that variable used in the logic below.

scripts/docker-run.sh
```diff
@@ -38,6 +38,27 @@ else
...
+# is set. If neither is set, attempt to use git to determine the upstream
+# branch. If that also fails, then exit.
+#
+UPSTREAM_BRANCH="$UPSTREAM_PRODUCT_BRANCH"
+if [[ -z "$UPSTREAM_BRANCH" ]]; then
+    echo "UPSTREAM_PRODUCT_BRANCH is not set."
+    echo "Attempting to infer UPSTREAM_BRANCH."
+    UPSTREAM_BRANCH=$(git for-each-ref --format='%(upstream:short)' \
```
pzakha: I think @prakashsurya mentioned that since this will not be run on a developer laptop, the upstream branch will likely not be set. Also, another reason why using an upstream is likely problematic: let's say I clone my private appliance-build repository, which has the branch "testing" to which I just pushed from my laptop. If I then do `git checkout testing`, it will create a new branch named "testing" which tracks "origin/testing", and so UPSTREAM_PRODUCT_BRANCH will be set to "testing", which is definitely not what we want.

I'd propose we just fail and mention a message such as:
```please set UPSTREAM_PRODUCT_BRANCH to the version of the product branch you are trying to build (e.g. "master", "6.0/stage", "release/6.0.0.0")```. By giving concrete examples of product branches, the developer is much more likely to provide the correct value.

File: scripts/run-live-build.sh (hunk 3 above)
Comment (grodr) [reply scripts/run-live-build.sh]: If that is a common workflow then I lean towards hard coding the `BRANCH` parameter in the repo and having the release lead change the value on branching. I think that will give the smallest chance for a developer having a misconfiguration.

Comment (pzakha) [reply scripts/run-live-build.sh]: This is not the way I'd go for it, but if the release team is OK with taking this extra burden then I'm fine with this. Note that this would mean that when we are branching we need to make 3 PRs for changing this variable in 3 different branches.

Alternatively, if you decide to go for the approach where the user has to provide the upstream product branch when building manually, then we can always switch back to the BRANCH variable hardcoding later on if users of this job script this too cumbersome or confusing.

File: scripts/docker-run.sh (hunk 9 above)
Comment (grodr) [reply scripts/docker-run.sh]: Given what @prakashsurya mentioned I am leaning more towards including the hard coded BRANCH parameter in the repository itself, and having that updated by the release lead. I think that will make it easy to use for developers, and difficult for them to screw up the configuration.

Comment (grodr) [reply scripts/docker-run.sh]: After talking with @pzakha offline I opted to ensure that UPSTREAM_PRODUCT_BRANCH is always set. If this ends up being too cumbersome for developers I will change this in the future.

Comment (prakashsurya) [reply scripts/docker-run.sh]: Can a summary of that discussion be posted here, so that all parties involved can have a say in the decision.

Given that we’ve been going back and forth on this issue, I think it’s only appropriate to have that discussion in the the open on this PR.

Additionally it’s good to have that discussion captured here in case the decision comes into scrutiny at some later time.

I think it’s perfectly reasonable to use a hard-coded value. Especially given, when we create a new release branch(es) we’ll have to create the new mirrors. So it’s a natural extension that after we create the mirror for the new release, we explicitly update the code to start using that new mirror.

If we allow the value to be passed it, it opens the door to be incorrectly specified, which I’m not sure how that error would get handled; e.g. if the user specifies “master” when building an older release.

File: scripts/docker-run.sh [hunk 10]
```diff
@@ -38,6 +38,19 @@ else
 	DOCKER_RUN="docker run"
 fi
 
+#
+# Ensure that UPSTREAM_PRODUCT_BRANCH is set.
+#
+if [[ -z "$UPSTREAM_PRODUCT_BRANCH" ]]; then
```
prakashsurya: What is expected when an incorrect value is specified? If “master” is set, but we’re building a non-master branch, will the build fail, or silently do the wrong thing?

Also, please update the build instructions in the README to document this new requirement.

scripts/docker-run.sh
pzakha: The summary is that this is the easiest way to go forward for now to get this work integrated, but that we will most likely want to get the branches hard-coded in the longer term. However, we should spend some time to do it properly, both for this repository and for linux-pkg. I've created this issue to track the future work: https://jira.delphix.com/browse/LB-221.

Another question is once we hardcode this, what do we do with the value currently passed in UPSTREAM_PRODUCT_BRANCH through Jenkins? Ideally I think we should remove it as there will be 2 sources of truth which would be confusing.

prakashsurya: What do you mean by do it properly? How was the prior solution proposed not proper? Since the discussion where you swayed @grodr was not done here in the open, it’s hard for me to understand what you mean. 

I do not agree that this is the easiest way to go forward, and I do not agree that this is the best solution in general.

But clearly you are not going to try to drive to a consensus here, but rather you want your approach taken and opt for offline conversations to keep me out of the loop to achieve that goal.

I don’t have the patience, so do what you will, but please don’t count me as a reviewer.

scripts/docker-run.sh
grodr: If the wrong values is specified then the job will silently do the wrong thing. It will attempt the build, but pull in the incorrect packages for that version. I think the build will succeed, but with those wrong package versions.

prakashsurya: I want to point out where this is likely to do the wrong thing...

In Jenkins, we configure the jobs that is used by "bors" like this:
```
createApplianceBuildJob('post-push', 'staging', 'master', false, false, false, false) 
```
i.e. we assume that anytime we're building the "staging" branch, the "upstream branch" is "master".

This assuming is not true if we're attempting to "bors r+" a pull request that is not based on master. For example, a backport PR will be based on the "6.0/release" branch, and when "bors r+" is used, the "staging" branch will become the 6.0/release branch merged with the pull request.

Based on this logic in this change, coupled with how our Jenkins logic incorrectly assumes "staging" always maps to "master", we'd end up doing a build of the 6.0/release branch, but wind up using the "master" branch's package repository.

Arguably, this is a bug in the Jenkins logic, and this change would simply be a victim here; but regardless, I think it's worth pointing out this issue, and being aware of it before landing this change.

prakashsurya: I've opened https://jira.delphix.com/browse/TOOL-8444 to track this issue.

grodr: I think with this concern and the concern that if a developer specifies an incorrect UPSTREAM_PRODUCT_BRANCH the build will silently be done with an incorrect mirror dataset I am going to reintroduce the hard coded BRANCH parameter. I think I am going to add this at a file in the root of the repo called something like `branch.config` to make it easy for the release lead to update this parameter, and make it easier to possibly automate this change later.

With this concern with bors I think the UPSTREAM_PRODUCT_BRANCH parameter should just be pulled out, and appliance build should always just pull in the BRANCH from the new branch config file. Can you think of a situation where a developer would want to override this? They could do that by changing the value in the config file, but I am trying to decide if this should be a documented workflow.

pzakha: I think given the issues this seems reasonable. 

I'd also like to have feedback from other release-team folks like @shartse and @basil .

prakashsurya: I'd be fine using some sort of "metadata file", where we specify this sort of information.

We do something similar to this with the `.hooksconfig` file in some of our other repositories. There's also some precedence for this kind of approach in the upstream ZFS on Linux code base; see [here](https://github.com/delphix/zfs/blob/master/META).

So, as a general design, I think it could work. If we go down this route, I think it'd be best to try and design a solution that could generally be used in all of our various repositories, so we have a common solution across them all (potentially, at least; but only if the specific repositories requires the complexity).

This is something that'd need to be updated when we create new release branches; but if we're consistent, we could probably write some automation to create the branch and update the file at the same time.

prakashsurya: > They could do that by changing the value in the config file, but I am trying to decide if this should be a documented workflow.

IMO, I wouldn't document it, nor officially support changing the value.

branch.config
```diff
@@ -0,0 +1,13 @@
...
+
+#
+# The "BRANCH" parameter tracks the upstream branch of appliance-build. It is
+# used to determine which branch of the linux package mirror will be used for
+# the build if UPSTREAM_PRODUCT_BRANCH is not set. UPSTREAM_PRODUCT_BRANCH is
+# set when appliance build is built by the appliance-build Jenkins job. This
+# parameter should be updated by the release lead on branching
+#
+
+UPSTREAM_BRANCH="master"
```
pzakha: missing new line

scripts/build-ancillary-repository.sh
```diff
@@ -142,7 +142,7 @@ function build_ancillary_repository() {
 #    environment variables, and the script will work as expected.
 #
 
-upstream_branch="${UPSTREAM_PRODUCT_BRANCH:-master}"
+upstream_branch="${UPSTREAM_BRANCH:-master}"
```
pzakha: We do not need this anymore since `UPSTREAM_BRANCH` should always be set.
Can you please get rid of `upstream_branch` and do a fatal check that `UPSTREAM_BRANCH` is set?

scripts/docker-run.sh
```diff
@@ -38,6 +38,29 @@ else
...
+        exit 1
+    fi
+
+    if [[ -z "$UPSTREAM_BRANCH" ]]; then
+        echo "UPSTREAM_BRANCH parameter was not sourced from branch.config." \
+             "Ensure branch.config is properly formatted with e.g." \
+             "UPSTREAM_BRANCH=\"<upstream-product-branch>\""
+        exit 1
+    fi
+    echo "Defaulting to branch $UPSTREAM_BRANCH."
```
pzakha: nit: "Defaulting to branch $UPSTREAM_BRANCH set in branch.config".

grodr:
I am waiting to merge this until the mirror update job finishes http://ops.jenkins.delphix.com/job/devops-gate/job/master/job/linux-package-mirror-sync/job/master/job/mirror1-sync/23/console (this failed last night due to unstable dx-integration tests). This is required since we have been building linux-pkg against newer kernel versions than our mirror currently has, so we need to update the mirror to not regress the kernel version.

scripts/docker-run.sh
grodr: updated

scripts/build-ancillary-repository.sh
grodr: shellcheck failed for me when I did a check of UPSTREAM_BRANCH (e.g. `if [[ -z UPSTREAM_BRANCH ]];...`) without previously assigning it in the script, so I opted to reassign to `upstream_branch` defaulting to nothing, and then check `upstream_branch`.

branch.config
grodr: Added.

scripts/docker-run.sh
grodr: 👍

pzakha:
bors delegate+

grodr:
bors +r

PR #368, opened 2019-10-08
Ensure that we are using secondary mirror

jgallag88:
This removes uses of non-mirror apt repositories. In particular, we were using the ansible repository to fetch ansible packages and the upstream ddebs repository to fetch the kernel debug symbols. By removing those repositories, we ensure that we are pulling these packages from our mirrors.

This also adds to live-build the public key for the secondary mirror, allowing us to remove `[trusted=yes]` from the repository configuration.


jgallag88:
### Testing:

Checked the appliance-build output to confirm that the ansible and debug symbol packages were being fetched from our mirror.

ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2291/console

jgallag88:
@grodr I've left off the `bionic-security` archive for the secondary mirror for now. I tried to add that, but the build failed with the error:
```
16:41:46  E: The repository 'http://linux-package-mirror.****.com/master/415f972b/ppas bionic-security Release' does not have a Release file.
```
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2291/console
Not sure if that is expected or not.

grodr:
> Not sure if that is expected or not.

@jgallag88 that is expected. The ddebs repository doesn't have a a `security` sub repo (see here http://ddebs.ubuntu.com/dists/) and it doesn't look like the ansible ppa has a `security` version either.

prakashsurya:
bors delegate+

live-build/auto/config
```diff
@@ -31,24 +31,6 @@
 # delphix-platform.
 #
 lb config noauto \
```
pzakha: How does this `lb config` step interact with the `lb config` done in `scripts/run-live-build.sh`?

jgallag88: The `auto` script here is basically a wrapper around the real `lb config` command, and `run-live-build.sh` is invoking this script. The invocation here provides the default options, which are added to / overridden by the options passed in from `run-live-build.sh`. This has some details on how it works: https://live-team.pages.debian.net/live-manual/html/live-manual/managing-a-configuration.en.html

PR #370, opened 2019-10-08
[Backport of LB-209 to 6.0/stage] Want to consume new mirror for prod…

grodr:
Backport of commit 4aa9fc996e5c589545f507eb8d7d309f68e96aed on master to allow appliance-build to use the package mirror for 6.0/stage builds.

Testing link (kicked off appliance-build-stage1 without UPSTREAM_PRODUCT_BRANCH set and verified `Defaulting to branch 6.0/stage set in branch.config` was present in jenkins log )
http://collins.d.delphix.com:35379/job/devops-gate/job/master/job/appliance-build-stage1/job/master/job/post-push/354/consoleFull

### DIFF
--- a/branch.config
+++ b/branch.config
@@ -1,0 +1,13 @@
+#
+# Copyright (c) 2019 by Delphix. All rights reserved.
+#
+
+#
+# The "BRANCH" parameter tracks the upstream branch of appliance-build. It is
+# used to determine which branch of the linux package mirror will be used for
+# the build if UPSTREAM_PRODUCT_BRANCH is not set. UPSTREAM_PRODUCT_BRANCH is
+# set when appliance build is built by the appliance-build Jenkins job. The
+# UPSTREAM_BRANCH parameter should be updated by the release lead on branching
+#
+
+UPSTREAM_BRANCH="6.0/stage"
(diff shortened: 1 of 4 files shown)


prakashsurya:
I want to point out, in this build, the `UPSTREAM_PRODUCT_BRANCH` is set to "master".. see [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-stage1/job/master/job/pre-push/6833/parameters) for the build parameters. I don't know for sure, since I haven't tried to root cause why it's using "master", but it might be a result of [this](https://jira.delphix.com/browse/TOOL-8444) bug.

pzakha:
We will probably need to just push this manually without bors for now then.

prakashsurya:
> We will probably need to just push this manually without bors for now then.

I'd like a root cause with details about why we can't (or, don't want to) fix the build, before we stop enforcing testing on our release branches. I don't currently support such a decision to disable the testing.

pzakha:
I'm just proposing that we push this particular work so that we can leverage Delphix mirrors on the Delphix/staging branch. This particular PR did get tested already (see link in description).

This does look like TOOL-8444, and I do think we should fix it, I'm just proposing to not block this PR until we fix it.

PR #381, opened 2019-10-25
DLPX-66286 Interrupting a running upgrade verify can leave the engine not-upgradable

shartse:
Extending `dx_delete` for migration to clean up the state of verify, even when it was cancelled in the middle of an operation. This includes: undoing changes to the bootloader, removing the MDS clone service and datasets and destroying `rpool/ROOT` using `-f`, which force unmounts its children as well. 

This was all based on what `dx_apply` and `dx_verify` do to clean up after themselves and replicating that logic in `dx_delete` so we can guarantee ending up in a "clean" state if we're cancelled or fail unexpectedly. 

I manually tested this by running a verify and cancelling it at various stages (roughly during `dx_apply` and `dx_verify`), ensuring that the cancel succeeded and then checking that a subsequent verify can run successfully as well. 

Automated testing: 
`git-ab-pre-push --test-upgrade-from 5.3.6.0` (diff 1) http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2375/
`git-ab-pre-push --test-upgrade-from 5.3.6.0` (diff 2) http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2385/
`git-ab-pre-push --test-upgrade-from 5.3.6.0` (diff 3) http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2391/

### DIFF
--- a/live-build/misc/migration-scripts/dx_delete
+++ b/live-build/misc/migration-scripts/dx_delete
@@ -39,10 +39,84 @@ function die() {
 # names are randomly generated.
 #
 
-linux_dataset="rpool/ROOT"
-if ! zfs list ${linux_dataset} &>/dev/null; then
-	echo "Linux dataset '${linux_dataset}' is not installed."
+LX_DATASET="rpool/ROOT"
+if ! zfs list ${LX_DATASET} &>/dev/null; then
+	echo "Linux dataset '${LX_DATASET}' is not installed."
 	exit 0
 fi
 
-zfs destroy -r ${linux_dataset} || die "Failed to destroy Linux dataset '${linux_dataset}'"
+LX_RDS_PARENT=$(zfs list -o name -H -d 1 "$LX_DATASET" | tail -n 1)
+[[ -n $LX_RDS_PARENT ]] || die "could not find Linux RDS parent dataset"
+LX_CONTAINER="${LX_RDS_PARENT##*/}"
+LX_RDS_TMP_ROOT_MOUNT="/tmp/$LX_CONTAINER/root"
+
+MDS_SNAPNAME="MDS-CLONE-upgradeverify"
+MDS_CLONE=domain0/$MDS_SNAPNAME
+SVC=svc:/system/delphix/postgres:$MDS_SNAPNAME
+PG_DATA=/$MDS_CLONE/db
+
+BOOT_MENU=/boot/menu.rc.local
+BOOT_MENU_COPY=$BOOT_MENU.copy
+
+function dx_apply_cleanup() {
+	#
+	# Restore original version of the bootloader
+	#
+	if [[ -e $BOOT_MENU_COPY ]]; then
+		mv $BOOT_MENU_COPY $BOOT_MENU ||
+			die "failed to restore copy of $BOOT_MENU"
+	fi
+
+	rm -f /boot/vmlinuz-* /boot/initrd.img-* ||
+		die "failed to destroy previously copied Linux kernel data"
+}
+
+function dx_verify_cleanup() {
+	#
+	# dx_verify creates a clone of the MDS dataset and the postgres
+	# service as well as a clone of /var/delphix for masking validation.
+	# If any are left, clean them up. (This logic is based on dx_manage_pg
+	# stop and cleanup.)
+	#
+	local sta
+	sta=$(svcs -Ho sta $SVC)
+	if [[ -n "$sta" ]]; then
+		/usr/sbin/svcadm disable -s $SVC ||
+			die "unable to disable SMF service: $SVC"
+		/usr/sbin/svccfg delete -f $SVC ||
+			die "unable to delete SMF service: $SVC"
+	fi
+
+	#
+	# Clean up the postmaster.pid leftover from the main MDS in the
+	# snapshot. This needs to be done after disabling the SMF service in
+	# case the postmaster.pid was instead from the active clone.
+	#
+	if [[ -e $PG_DATA/postmaster.pid ]]; then
+		echo "renaming $PG_DATA/postmaster.pid to $PG_DATA/postmaster.pid.original"
+		mv $PG_DATA/postmaster.pid $PG_DATA/postmaster.pid.original ||
+			echo "failed to move postmaster.pid file"
+	fi
+
+	[[ $(zfs list domain0/mds@$MDS_SNAPNAME) ]] &&
+		zfs destroy -R "domain0/mds@$MDS_SNAPNAME"
+	[[ $(zfs list domain0/mds@$MDS_SNAPNAME) ]] &&
+		die "unable to cleanup domain0/mds@$MDS_SNAPNAME"
+
+	runningVar=$(mount | awk '/^\/var\/delphix /{ print $3 }')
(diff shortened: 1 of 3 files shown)


live-build/misc/migration-scripts/dx_delete
```diff
@@ -39,10 +39,94 @@ function die() {
...
+	#
+}
+
+function dx_apply_cleanup() {
+	#
+	# Remove new bootloader entry
+	#
+	sed -i '/set lxcmdline/d' /boot/menu.rc.local
+	[[ "$(grep -cF 'set lxcmdline' /boot/menu.rc.local)" -eq 0 ]] ||
+		die "failed to cleanup lxcmdline from previous run"
```
sdimitro: This will only remove the command executed by the new 8th option that we added to the bootloader (i.e. the option to boot linux). The actual option will not be gone from the menu, so I think we probably want to add the following lines too:
```
sed -i '/set mainmenu_caption\[8\]/d' /boot/menu.rc.local
[[ "$(grep -cF 'mainmenu_caption[8]' /boot/menu.rc.local)" -eq 0 ]] ||
	die "failed to cleanup mainmenu_caption from previous run"
sed -i '/set mainansi_caption\[8\]/d' /boot/menu.rc.local
[[ "$(grep -cF 'mainansi_caption[8]' /boot/menu.rc.local)" -eq 0 ]] ||
	die "failed to cleanup mainansi_caption from previous run"
sed -i '/set mainmenu_keycode\[8\]/d' /boot/menu.rc.local
[[ "$(grep -cF 'mainmenu_keycode[8]' /boot/menu.rc.local)" -eq 0 ]] ||
	die "failed to cleanup mainmenu_keycode from previous run"
sed -i '/set mainmenu_command\[8\]/d' /boot/menu.rc.local
[[ "$(grep -cF 'mainmenu_command[8]' /boot/menu.rc.local)" -eq 0 ]] ||
	die "failed to cleanup mainmenu_command from previous run"
```

In addition we leave some leftover state in the `/boot` directory of illumos that could be cleaned:
```
rm -f /boot/vmlinuz-* /boot/initrd.img-* ||
	die "failed to destroy previously copied Linux kernel data"
```

sebroy: Another approach to consider would be to save the original file so that it can be copied back on cleanup from failure.

live-build/misc/migration-scripts/dx_delete
```diff
@@ -39,10 +39,94 @@ function die() {
...
+function dx_verify_cleanup() {
+	#
+	# dx_verify creates a clone of the MDS dataset and the postgres
+	# service as well as a clone of /var/delphix for masking validation.
+	# If any are left, clean them up. (This logic is based on dx_manage_pg
+	# stop and cleanup.)
+	#
+	local sta
+	sta=$(svcs -Ho sta $SVC)
+	if [[ -n "$sta" && ("$sta" == ON || "$sta" == DGD) ]]; then
```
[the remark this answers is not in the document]
shartse: Yeah, I don't see why dx_manage_pg was only deleting the service in those states - it makes to me to delete fully. I found from experimenting that even if the service is `OFF`, it still causes the next round of verification to fail since you can't create a new one.

live-build/misc/migration-scripts/dx_delete
shartse: Yeah, I could see that being pretty clean. At the beginning of `dx_apply` we could save the file in `/boot/menu.rc.local.copy` and the restore it in `dx_delete`. What do you think of this approach @sdimitro ?

sdimitro: Yeah that makes sense to me!

live-build/misc/migration-scripts/dx_delete
```diff
@@ -39,10 +39,79 @@ function die() {
...
+MDS_SNAPNAME="MDS-CLONE-upgradeverify"
+MDS_CLONE=domain0/$MDS_SNAPNAME
+SVC=svc:/system/delphix/postgres:$MDS_SNAPNAME
+PG_DATA=/$MDS_CLONE/db
+
+function dx_apply_cleanup() {
+	#
+	# Restore original version of the bootloader
+	#
+	mv "/boot/menu.rc.local.copy" "/boot/menu.rc.local" ||
```
sebroy: One thing to consider here is that it's possible for `dx_delete` to be called even if `dx_apply` was never called. For example, I upload a migration image, decide not to attempt to install it, and the click `Delete` on that version in the GUI. This goes through the `UpgradeManagerImpl#uninstallVersion` method which unconditionally calls `dx_delete` regardless of the state of the version.

shartse: Ah, good point. I forgot about that path. I'll check it out.

live-build/misc/migration-scripts/dx_delete
```diff
@@ -39,10 +39,79 @@ function die() {
...
+	[[ $(zfs list domain0/mds@$MDS_SNAPNAME) ]] &&
+		zfs destroy -R "domain0/mds@$MDS_SNAPNAME"
+	[[ $(zfs list domain0/mds@$MDS_SNAPNAME) ]] &&
+		die "unable to cleanup domain0/mds@$MDS_SNAPNAME"
+
+	runningVar=$(mount | awk '/^\/var\/delphix /{ print $3 }')
+	[[ $(zfs list "$runningVar/$MDS_SNAPNAME") ]] &&
+		zfs destroy -f "$runningVar/$MDS_SNAPNAME"
+	[[ $(zfs list "$runningVar/$MDS_SNAPNAME") ]] &&
+		die "unable to cleanup $runningVar/$MDS_SNAPNAME"
```
pzakha: This dataset doesn't look familiar, where is it coming from?

live-build/misc/migration-scripts/dx_delete
pzakha: Yeah, we should restore it only if it exists. `dx_delete` should be idempotent.

live-build/misc/migration-scripts/dx_delete
shartse: We do a deferred destroy on it after making the clone on line 107:
`zfs destroy -d "$runningVar@$MDS_SNAPNAME"`

It's a kind of odd part in my opinion. Do you think I should change it to explicitly cleanup both snapshot and clone?

pzakha: Ah, I didn't know such an option existed! I'm okay either way.

live-build/misc/migration-scripts/dx_delete
shartse: Ok, added an existence check for restoring the file.

live-build/misc/migration-scripts/dx_delete
shartse: Since this is the second time I've seen confused by this (the first time was me a few months ago), I changed it to recursively delete the snapshot instead in both files.

sebroy:
bors delegate+

PR #388, opened 2019-11-11
DLPX-67184 migration: dx_apply should call dx_delete to cleanup previous state

pzakha:
`dx-delete` will restore the old bootloader menu before the Linux dataset was dx-applied, effectively removing all the new linux entries.

Regarding the cleanup of `rpool/ROOT`, dx-delete takes care of it, so we can remove the now redundant logic.

## Testing 
ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2414/

### DIFF
--- a/live-build/misc/migration-scripts/dx_apply
+++ b/live-build/misc/migration-scripts/dx_apply
@@ -106,13 +106,8 @@ RPOOL=${RDS%%/*}
 #
 # Cleanup any previous intermediate state.
 #
-rm -rf /tmp/delphix.* ||
-	die "failed to destroy old delphix temporary directories"
-rm -f /boot/vmlinuz-* /boot/initrd.img-* ||
-	die "failed to destroy previously copied Linux kernel data"
-zfs destroy -r "$RPOOL/ROOT" 2>/dev/null
-zfs list "$RPOOL/ROOT" 2>/dev/null &&
-	die "could not destroy linux root dataset from previous run"
+"${BASH_SOURCE%/*}/dx_delete" ||
+	die "failed to cleanup previous state with dx_delete"
 
 #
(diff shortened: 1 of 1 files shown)


pzakha:
cc: @shartse @sdimitro

live-build/misc/migration-scripts/dx_apply
```diff
@@ -106,13 +106,8 @@ RPOOL=${RDS%%/*}
...
 # Cleanup any previous intermediate state.
 #
-rm -rf /tmp/delphix.* ||
-	die "failed to destroy old delphix temporary directories"
-rm -f /boot/vmlinuz-* /boot/initrd.img-* ||
-	die "failed to destroy previously copied Linux kernel data"
-zfs destroy -r "$RPOOL/ROOT" 2>/dev/null
-zfs list "$RPOOL/ROOT" 2>/dev/null &&
-	die "could not destroy linux root dataset from previous run"
+"${BASH_SOURCE%/*}/dx_delete" ||
```
sebroy: dx_delete only appears to destroy rpool/ROOT, and nothing else. We didn't need to do the other steps that were removed?

shartse: Which version of dx_delete are you looking at? I updated it recently to clean things up more comprehensively.

pzakha: I've updated the description to explain this.

sebroy: @shartse ha! I had a super outdated version of appliance-build that I was looking at. Indeed, fetching the latest one makes everything clear to me. Thanks for pointing this out. I even reviewed your code (<embarrassed>)...!

PR #393, opened 2019-11-14
DLPX-67392 azure marketplace requires a 1mb unallocated region

grwilson:
This is a new requirement from Microsoft for the Azure marketplace. This creates an empty 1MB unallocated region at the beginning of the disk which Microsoft can use. 

Testing:
ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2450/

We are deploying this image in the azure marketplace and have confirmed it has passed their certification.

### DIFF
--- a/live-build/config/hooks/vm-artifacts/90-raw-disk-image.binary
+++ b/live-build/config/hooks/vm-artifacts/90-raw-disk-image.binary
@@ -64,10 +64,12 @@ sgdisk --zap-all "$ARTIFACT_NAME.img"
 # partition will be used and automatically detected by "grub-install"
 # based on the partitions typecode. This partition is required since we're
 # partitioning using GPT; if we used MBR, an explicit boot partition
-# wouldn't be required.
+# wouldn't be required. Also we leave the first 1MB unallocated since
+# some clouds (i.e. Azure)  may require space for their own internal
+# purposes.
 #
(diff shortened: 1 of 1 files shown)


prakashsurya:
bors delegate+

pzakha:
> LGTM.
> 
> As a tangent, I wonder how/if these disk formatting differences between a fresh linux install and a migrated system will become problematic at some point...

I think this might be a real issue.
Regarding Azure, we may need to talk to Microsoft to see if that could become an issue for existing images that do not have that 1MB of free space.

grwilson:
> LGTM.
> 
> As a tangent, I wonder how/if these disk formatting differences between a fresh linux install and a migrated system will become problematic at some point...

This shouldn't be an issue for existing customers since MSFT won't use this region except during their internal testing. So it only impacts new submissions to the marketplace.

prakashsurya:
> This shouldn't be an issue for existing customers since MSFT won't use this region except during their internal testing. So it only impacts new submissions to the marketplace.

Sure.. my concern was more about "in general", rather than any specific platform.. i.e. all of our systems (migrated vs. new install) will have these differences, regardless of the platform.. and how that might affect the behaviour of the system.

I wasn't trying to spark a discussion, though, just a passing thought.

Issue #380, opened 2019-10-18
build failure fetching from internal mirror


grwilson:
We have seen several builds fail with the following error:

```
00:34:40  Get:641 http://linux-package-mirror.****.com/master/415f972b/ppas bionic-updates/main amd64 openssh-server-dbgsym amd64 1:7.6p1-4ubuntu0.3 [884 kB]
00:34:40  Get:642 http://linux-package-mirror.****.com/master/415f972b/ppas bionic-updates/main amd64 systemd-dbgsym amd64 237-3ubuntu10.29 [7677 kB]
00:34:40  Fetched 1379 MB in 18s (77.6 MB/s)
00:34:40  W: Download is performed unsandboxed as root as file '/packages/acl_2.2.52-3build1_amd64.deb' couldn't be accessed by user '_apt'. - pkgAcquire::Run (13: Permission denied)
00:34:40  E: Failed to fetch http://linux-package-mirror.****.com/master/415f972b/ubuntu/pool/universe/t/trace-cmd/trace-cmd_2.6.1-0.1_amd64.deb  Undetermined Error [IP: 10.110.20.241 80]
00:34:40  E: config/hooks/81-upgrade-repository.binary failed (exit non-zero). You should check for errors.
```

Looking earlier in the build we see the `acl` package was successfully accessed:
```00:27:21  Get:221 http://linux-package-mirror.****.com/master/415f972b/ubuntu bionic/main amd64 acl amd64 2.2.52-3build1 [38.5 kB]
00:27:40  Selecting previously unselected package acl.
00:27:40  Preparing to unpack .../126-acl_2.2.52-3build1_amd64.deb ...
00:27:40  Unpacking acl (2.2.52-3build1) ..

00:28:32  Setting up acl (2.2.52-3build1) ...
```

And as we start to run the upgrade build image we see the following:
```
00:34:22  Get:1 http://linux-package-mirror.****.com/master/415f972b/ubuntu bionic/main amd64 acl amd64 2.2.52-3build1 [38.5 kB]
```

But it fails later:
```
00:34:40  W: Download is performed unsandboxed as root as file '/packages/acl_2.2.52-3build1_amd64.deb' couldn't be accessed by user '_apt'. - pkgAcquire::Run (13: Permission denied)
[...]

PR #395, opened 2019-11-15
Retry downloading packages on failure

pzakha:
This is a proposed workaround for #380 

## Testing
relevant sub-job from ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-stage1/job/master/job/pre-push/7356/consoleFull
ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2506/

### DIFF
--- a/live-build/config/hooks/configuration/download-packages.sh
+++ b/live-build/config/hooks/configuration/download-packages.sh
@@ -1,0 +1,23 @@
+#!/bin/bash -eu
+
+set -o pipefail
+
+function download_packages() {
+	echo "Downloading: $*"
+	for _ in {1..5}; do
+		apt-get download "$@" && return
+		echo "Download failed, retrying."
+		sleep 5
+	done
+	echo "Failed to download packages after 5 attempts. Aborting." >&2
+	return 1
+}
+
+mkdir -p /packages
+cd /packages
+
+apt-get update
+
+export -f download_packages
+dpkg-query -Wf '${Package}=${Version}\n' |
+	xargs -n 20 bash -c 'download_packages "$@"' _
--- a/live-build/config/hooks/configuration/81-upgrade-repository.binary
+++ b/live-build/config/hooks/configuration/81-upgrade-repository.binary
@@ -59,16 +59,8 @@ fi
 # This directory of download packages will later be passed to Aptly to
 # build our "upgrade repository".
 #
-chroot binary bash -eux <<'EOF'
-set -o pipefail
-
-mkdir -p /packages
-cd /packages
-
-apt-get update
-
-dpkg-query -Wf '${Package}=${Version}\n' | xargs apt-get download
-EOF
+cp config/hooks/download-packages.sh binary/tmp/
+chroot binary /tmp/download-packages.sh
 
 #
 # After downloading the packages, the package filenames any have the
--- a/live-build/misc/migration-scripts/dx_apply
+++ b/live-build/misc/migration-scripts/dx_apply
@@ -99,7 +99,7 @@ MIGRATION_MINOR_VERSION=$(echo "$MIN_MIGRATION_VERSION" | cut -d. -f3)
 
 DX_UPG_STRESS=$ARCHIVE_DIR/dx_upg_stress_options
 # shellcheck source=/dev/null
-. $DX_UPG_STRESS --source
+. "$DX_UPG_STRESS" --source
 __trigger_unset_stress_option "STRESS_DX_APPLY_FAIL_AFTER_VERSION_CHECK"
 
 #
--- a/live-build/misc/migration-scripts/dx_execute
+++ b/live-build/misc/migration-scripts/dx_execute
@@ -35,7 +35,7 @@ set -o pipefail
 DX_UPG_PAUSE="${BASH_SOURCE%/*}/dx_upg_pause_options"
 DX_UPG_STRESS="${BASH_SOURCE%/*}/dx_upg_stress_options"
 # shellcheck source=/dev/null
-. $DX_UPG_STRESS --source
+. "$DX_UPG_STRESS" --source
 
 function die() {
 	echo "$(basename "$0"): $*" >&2
--- a/live-build/misc/migration-scripts/dx_verify
+++ b/live-build/misc/migration-scripts/dx_verify
@@ -201,7 +201,7 @@ run_upgrade_verify "$output" "$format" "$locale" 20 95
 
 DX_UPG_STRESS="${BASH_SOURCE%/*}/dx_upg_stress_options"
 # shellcheck source=/dev/null
-. $DX_UPG_STRESS --source
+. "$DX_UPG_STRESS" --source
 __trigger_unset_stress_option "STRESS_DX_VERIFY_FAIL_AFTER_TEST_MIGRATION"
 
 test_masking


prakashsurya:
Can you please add information to help explain why we'd want to implement this suggested solution, vs. other alternative solutions? Can you also add information about any alternative solutions that were investigated (if any), and the benefits/drawbacks of these alternatives when compared to what you're proposing here.

prakashsurya:
Can you also add information to help explain why the download failures occur? I think we should fully understand why the failures occur, especially given that we're using infrastructure that's under our control, before we attempt to workaround the failures.

pzakha:
Most of the failures in #380 are in the stage where we download packages.

We've spend some time investigating why the failures are occurring but did not have much luck so far. I think we should spend more time root causing this issue after we ship 6.0.

In the meanwhile, this can help with:
a) Reducing the number of failures and so reducing the burden to monitor failed builds.
b) If this does not reduce failures and we see the download of a package fail for several attempts in a row, this could give us more insight into the problem.

prakashsurya:
Can you also respond to my first comment, about potential alternative solutions, and why you believe this is the best path to take.

pzakha:
A few alternatives:
 1. Re-use packages downloaded previously during the build, so "apt download" can get them from cache. This may not work for all packages, but should work for most.
2. Spend more time investigating the root cause for the failure. One idea was to monitor the workload on the server.

Both alternatives would most likely take a good amount of time, and would probably have to be done after we ship 6.0.

pzakha:
I think implementing this workaround now will help us to get us a better success rate on our builds.

prakashsurya:
Why is shipping 6.0 a factor here? Is this blocking the release?

Unless I'm mistaken, the build is currently flaky, but we can re-run it when it fails, and continue to re-run until we get a successful build right? and then we use that successful build to ship the release?

It's not clear why we'd actively pursue a short term solution due to the 6.0 release, even though this issue isn't blocking the release?

pzakha:
The main idea here is to reduce the maintenance burden of monitoring failed builds, so that we can focus on real failures instead of infrastructure flakiness. If we think dealing with flakiness is fine for the next few weeks, until we get more time to root cause this, I'm fine with dropping this.

live-build/config/hooks/configuration/download-packages.sh
```diff
@@ -0,0 +1,32 @@
...
+PACKAGES=()
+while read -r package; do
+	PACKAGES+=("$package")
+	if [[ "${#PACKAGES[@]}" -ge 20 ]]; then
+		download_packages "${PACKAGES[@]}"
+		PACKAGES=()
+	fi
+done </tmp/packages
+
+[[ "${#PACKAGES[@]}" -ne 0 ]] && download_packages "${PACKAGES[@]}"
```
prakashsurya: could this be replaced with something like this:
```
export -f download_packages
dpkg-query -Wf '${Package}=${Version}\n' | xargs -n 20 bash -c 'download_packages "$@"' _
```
?

pzakha: That's much more concise. It seems like it doesn't abort if `download_packages` fails though, not sure why.

prakashsurya: What do you mean? Here's what I see in my testing...

- All succeed, it returns success:
```
± function dostuff() { exit 0; }
± echo 1 2 3 4 | xargs -n2 bash -c 'dostuff "$@"' _
± echo $?
0
```

- All fail, it returns failure:
```
± function dostuff() { exit 1; }
± echo 1 2 3 4 | xargs -n2 bash -c 'dostuff "$@"' _
± echo $?
123
```

- one fails, it returns failure:
```
± function dostuff() { local R=$((RANDOM % 2)); echo $R; exit $R; }
± echo 1 2 3 4 | xargs -n1 bash -c 'dostuff "$@"' _
0
1
0
0
± echo $?
123
```

live-build/config/hooks/configuration/download-packages.sh
```diff
@@ -0,0 +1,32 @@
...
+
+function download_packages() {
+	echo "Downloading: $*"
+	for _ in {1..5}; do
+		apt-get download "$@" && return
+		echo "Download failed, retrying."
+		sleep 5
+	done
+	echo "Failed to download packages after 5 attempts. Aborting." >&2
+	exit 1
```
prakashsurya: we use `exit` here, but `return` earlier in this function.. we should try to be consistent and either use `exit` for both, or `return` for both.

live-build/config/hooks/configuration/download-packages.sh
pzakha: Yeah, in my quick testing I was running something very similar to your 3rd case, where we can see it doesn't stop on failure. It does return an error code at the end though, so it should be fine. I've updated the review.

pzakha:
re-testing ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2507/

pzakha:
@prakashsurya I think I've addressed your feedback, can you take another look?

prakashsurya:
As others have asked offline, how are we going to monitor and/or track the effectiveness of the retry loop? e.g. how will we track how many times we have to retry? how many retries are needed until it works? which packages required and retry, and which dont? how do we intend to use the data we get from this retry loop to help address the flakiness of the package server?

Also, if this is a short term solution, what criteria will be using to know when we should remove the retry loop, in favor of whatever long term solution we adopt?

I still feel like rushing into a short term solution might not be the right approach, since if this does alleviate build failures, I'd bet this will become the long term solution (e.g. due to other work taking priority). I'd like to get others to weigh in before approving this.

Tangentially, who is in charge of maintaining the package server? What kind of SLA does the server have? Is this flakiness expected? Have we opened a dialog/ticket with the folks that maintain the service?

sebroy:
@prakashsurya I filed internal tool issue TOOL-8699 against the internal package mirror as a result of these failures.

prakashsurya:
@sebroy Thanks! Are you good with this approach?

sebroy:
@prakashsurya I'm fine with this approach, although to your point, it would be great to include whether retries were needed, against what package, and how many, as part of the report of the job that's included in our slack notifications (e.g. similar to how we include some details in the `linux-pkg-update` job) so that I or the monitor can identify a pattern with little effort.

pzakha:
> I still feel like rushing into a short term solution might not be the right approach, since if this does alleviate build failures, I'd bet this will become the long term solution (e.g. due to other work taking priority).

I agree there's definitely a risk of this becoming a long term solution if things work. I think if this workaround actually fixes the issue, then it's not a big deal that we just leave it in place. The bigger risk is that we'd just mask the issue temporarily, and it would come back later and fail even with the workaround in place.

> it would be great to include whether retries were needed, against what package

That would definitely require quite a bit more work. If we want to go this route, I'd propose we hold off on this for now.

sebroy:
@pzakha would there be another way to fetch and report that same information, perhaps via post-processing? I'd like to maybe run such a report every week to see what the state of the job is, whether it had to retry the package updates, etc. etc. without having to spend a lot of time clicking through all of the successful runs.

pzakha:
There's definitely ways of doing that via post-processing, such as parsing the output of each `appliance-build-stage1` job and looking for specific strings. This may put quite some load on the Jenkins server though as the failure rate is pretty low and the script will have to parse several hundreds of jobs each week.

pzakha:
Superseded by https://github.com/delphix/appliance-build/pull/415

PR #396, opened 2019-11-15
remove crash-python role from appliance-build

sdimitro:
# Description

Now that we can run crash-python outside the git repo
from any path and we are halting future development on
it, its role on the appliance build is no longer needed
for developer VMs.

Note that deleting that role uncovered a different bug
where the minimal-common role in internal-dev is needed
by the ldap role and it would always run before it
because it was listed as a dependency under the crash
python role. Now that the crash-python role is removed
we re-add an old version of that configuration making
the minimal-common role run before the ldap role.

# Testing

ab-pre-push:
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/delphix-build-and-snapshots/job/ami-snapshots/1204/

Created an internal-dev image from the OVA generated above and ensured that the crash-python repo doesn't show up anymore and also that we still have crash-python in the system:
```
delphix@sd-cp:~$ ls
dlpx-app-gate  dms-core-gate  zfs
delphix@sd-cp:~$ crash-kcore.sh
Must be run as root
delphix@sd-cp:~$ sudo crash-kcore.sh
GNU gdb (GDB) 8.3.50.20190416-git
...
```


sdimitro:
cc @prakashsurya

PR #398, opened 2019-12-04
Leverage Github Actions in favor of TravisCI

prakashsurya:
This change adds the necessary files and logic to take advantage of
GitHub Actions for running our automated tests. Currently we rely on
TravisCI to do this, but we'd like to transition over to GitHub Actions,
for the following reasons:

    1. The Actions UI is more integrated with GitHub PRs
    2. Best practices at Delphix suggests using Actions


.github/workflows/main.yml
```diff
@@ -0,0 +1,21 @@
+on: [push, pull_request]
```
pzakha: Q1: Does `on: push` mean this is this going to run when pushing to any branch, i.e. this would also work for bors' `staging` branch?

Q2: Should we modify `bors.toml` to remove mentions of travis? Do we need to add something to bors to check the status of the actions?

prakashsurya: Q1: Yes, AFAIK it'll run on all branches containing this file.

Q2: Yes, you're right. But, rather than updating to use the actions' status, I think we should remove `bors.toml` to more closely adhere to our other repositories and the upcoming Delphix best practices. Currently, bors doesn't buy us much, since it's no longer verifying the build. I've pushed an updated version to remove/disable "bors".

PR #399, opened 2019-12-12
Add new "finalize" subcommand to upgrade script

prakashsurya:
This change adds the "finalize" subcommand to the upgrade script, which
is intended to be used to clean up any filesystems, snapshots, etc. that
may have been generated during the upgrade process (e.g. for rollback).

This is intended to be used to mark the finalization or completion of
the upgrade, after which rollback will no longer be possible; e.g.

    $ sudo /var/dlpx-update/latest/upgrade deferred


prakashsurya:
cc: @sumedhbala-delphix

prakashsurya:
I believe Sumedh is planning to pick this up and finish it, so I'll close this PR..

PR #402, opened 2020-01-06
Modify propagation type of root dataset mount before mounting children

jgallag88:
The change is to unblock a future change along the lines of https://github.com/delphix/appliance-build/pull/247.

The issue is that when we run the build outside of a container, there are mnt namespaces which are peers to the one in which the build is running. Currently, the following happens

- The root dataset `rpool/ROOT/delphix.v8IGiyn/root` is mounted on `/mnt/delphix.v8IGiyn`. This mount is propagated to the peer mnt namespaces.
- Child datasets `data`, `home`, and `log` are mounted beneath `/mnt/delphix.v8IGiyn`. These mounts are propagated to the peer mnt namespaces.
- The propagation type of `/mnt/delphix.v8IGiyn` is changed to 'slave'.
- We mount and unmount some other things which aren't directly relevant.
- We unmount  `data`, `home`, and `log`. These unmount events are _not_ propagated, because the parent mount is now no longer 'shared'. These datasets remain mounted in the other namespaces.
- We unmount `root`. This is also not propagated, because the mount is busy in the other namespaces, because of the children still mounted.
- We try to export the pool, but the pool is busy because of the mounts which still exist in the other namespaces.

We can resolve this by changing the propagation type of `/mnt/delphix.v8IGiyn` before mounting the other datasets on top of it. This way the child mounts are never propagated to the other namespaces.

### DIFF
--- a/live-build/config/hooks/vm-artifacts/90-raw-disk-image.binary
+++ b/live-build/config/hooks/vm-artifacts/90-raw-disk-image.binary
@@ -138,6 +138,37 @@ zfs create \
 
 zfs mount "rpool/ROOT/$FSNAME/root"
 
+#
+# We are later going to recursively bind mount /proc/, /sys/, and /dev/
+# beneath the root dataset. Before doing that, we need to change the root
+# dataset's mount so that it has type 'slave'. If were to leave it 'shared',
+# the following would happen:
+#
+#  - The mount of $DIRECTORY/sys/ would be propagated to other existing
+#    namespaces.
+#  - We would recursively change the propagation type of $DIRECTORY/sys/ to
+#    be 'slave', for reasons explained below.
+#  - After finishing our work, we would 'umount -R $DIRECTORY/sys/'. For the
+#    children operated on recursively, this event would _not_ be propagated
+#    because they are not longer shared.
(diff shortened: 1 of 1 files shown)


jgallag88:
ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2634/console

live-build/config/hooks/vm-artifacts/90-raw-disk-image.binary
```diff
@@ -137,6 +137,7 @@ zfs create \
 	"rpool/ROOT/$FSNAME/root"
 
 zfs mount "rpool/ROOT/$FSNAME/root"
+mount --make-slave "$DIRECTORY"
```
prakashsurya: Since the placement of this call is critical, can you perhaps add a comment to explain why it _needs_ to be called here? (e.g. in contrast to calling it later like before this change).

I'd be OK more-or-less copying the PR description into the form of a comment here, if needed.

jgallag88:
> Have you verified this will unblock #247 (or an equivalent change)?

I did test that; I should have mentioned that originally.

PR #404, opened 2020-01-08
Stop using Docker when executing the build

prakashsurya:
This change modifies the Ansible bootstrap logic to install all of the
necessary build dependencies onto the host that's running Ansible. This
way, we can execute appliance-build directly on the host, rather than
relying on the Docker contianer. Additionally, the documentation has
been updated to no longer reference the use of Docker.

With that said, we still build the Docker image and support building in
the Docker contianer just as before, so we can temporarily maintain
backwards compatibility with our automation, and avoid flag days.


prakashsurya:
This is a continuation of #247.

prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2656/)

gradlew
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env sh
```
pzakha: Are we expecting to be running this script on other systems aside from the bootstrap VM?

If not, it might be cleaner to remove the cygwin/msys/darwin/nonstop stuff.

The only time that I'd think we could use this script is for running the style checks locally. However that can also be done by installing gradle locally.

scripts/build-ancillary-repository.sh
```diff
@@ -142,10 +142,28 @@ function build_ancillary_repository() {
...
+
+	if [[ -z "$UPSTREAM_BRANCH" ]]; then
+		echo "UPSTREAM_BRANCH parameter was not sourced from branch.config." \
+			"Ensure branch.config is properly formatted with e.g." \
+			"UPSTREAM_BRANCH=\"<upstream-product-branch>\""
+		exit 1
+	fi
+	echo "Defaulting to branch $UPSTREAM_BRANCH set in branch.config."
+else
+	UPSTREAM_BRANCH="$UPSTREAM_PRODUCT_BRANCH"
```
pzakha: It might be cleaner we could add this logic to the "gradlew" wrapper. This way it will be set for all gradle Tasks and we do not need to worry about inconsistencies if other tasks/scripts end up needing this environment variable.

gradlew
prakashsurya: I recommend we don't touch this script at all; it comes from running `gradle wrapper`.

pzakha: I see, disregard this then.

scripts/build-ancillary-repository.sh
prakashsurya: We should not be modifying `gradlew`.

IMO, it would have been best to have had this logic added to the gradle build logic to begin with, but for whatever reason it was added to `docker-run.sh` (which we're no longer using). If you'd like to open a separate PR to move this into the gradle build logic, then I can rebase on to that PR, and remove this.. Otherwise, I'd like to move forward with what I have here.

pzakha: The reason this was added to `docker-run.sh` instead of the gradle logic is because there was no straight forward way to add logic in gradle to set environment variables for all sub-tasks. 

Do you think adding a wrapper on top of gradlew that would be responsible of setting the Environment the way we need would be a viable option?

prakashsurya: > Do you think adding a wrapper on top of gradlew that would be responsible of setting the Environment the way we need would be a viable option?

To me, that seems awfully complicated and awkward.. given that we're already leveraging a complicated and powerful build system (Gradle).

If I understand you correctly, we'd call into a shell script, which would then call into Gradle, which would then call into another shell script. This just seems unnecessarily complicated and convoluted to me.

But if you want to do the work to change this, then, by all means...

pzakha: Given that our build system was designed to depend on configuration to be passed through the environment, this wrapper script would just make sure that the configuration is sane and fill-in defaults (such as with the UPSTREAM_BRANCH parameter), so while it would add one more layer of complexity, the layer is fairly thin and simple.

That said, given that we don't have a consensus on this, I'm going to drop this issue for now so that we can proceed with this change.

prakashsurya: Thanks.

If we have time, maybe we can discuss this at our next team meeting.. or set up another meeting with @jgallag88 (and anybody else that's interested) to come up with a more maintainable solution than what I'm doing here.

jgallag88: We have done something like what Pavel is suggesting in the past (see `appliance/java-home-helper.diff` and `appliance/gradle/java-home-helper.sh` in the app-gate). It's not the end of the world, but it is a bit of a burden to maintain because we have to keep applying our diff to the wrapper every time we upgrade Gradle.

I think the simplest thing to do would be to move the logic to choose the correct value into a gradle function, and then use that to set the environment variable in the appropriate `Exec` tasks.

That said, I'm on board with pushing this as is. We can do a follow up if we want to.

README.md
```diff
@@ -98,15 +87,7 @@ correcting any deficencies that may exist. This is easily done like so:
...
-    $ ./scripts/docker-run.sh gradle ...
-
-This will create a new container based on the image we previously
-created, and then execute "gradle" inside of that container.
-
-The "./scripts/docker-run" script can also be run without any arguments,
-which will provide an interactive shell running in the container
-environment, with the appliance-build git repository mounted inside of
-the container; this can be useful for debugging and/or experimenting.
+    $ sudo -E ./gradlew ...
```
jgallag88: What environment variables do we need to preserve here?

.github/workflows/main.yml
```diff
@@ -5,17 +5,17 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v1
-      - run: docker build -t appliance-build:latest docker
-      - run: ./scripts/docker-run.sh gradle ansibleCheck
+      - run: sudo ./.github/scripts/install-ansible-lint.sh
+      - run: gradle ansibleCheck
```
jgallag88: Are we using a preinstalled version of Gradle? If so, we might be better of installing Gradle explicitly, in case the version of the preinstalled one changes.

README.md
prakashsurya: In general (e.g. when building manually), none are required. But it's possible to specify variables that cause the build to do different things; which is functionality used by Jenkins to pass in different values.

I could remove the suggestion of using `-E` here, since those reading this doc likely won't be specifying any custom values.

Should I remove it from here?

.github/workflows/main.yml
prakashsurya: Yes, it is.

I opted for this, since I was having trouble properly setting the `PATH` variable to enable access to both `ansible-lint` and `gradle` installed in `/opt`.

But if this is a requirement (I'll assume it is), I'll give it another shot, and see if I can come up with something relatively clean.

README.md
jgallag88: I think it makes sense to remove it. It makes the instructions a little simpler, and if people do want to do more advanced things, they can always `sudo WHATEVER=whatever ./gradlew ...`

.github/workflows/main.yml
jgallag88: Is invoking it by absolute path an option?

prakashsurya: Yea, that seems reasonable. Pushed a commit to do this.

PR #413, opened 2020-01-23
Add new "finalize" subcommand to upgrade script

prakashsurya:
This change adds the "finalize" subcommand to the upgrade script, which
is intended to be used to clean up any filesystems, snapshots, etc. that
may have been generated during the upgrade process (e.g. for rollback).

This is intended to be used to mark the finalization or completion of
the upgrade, after which rollback will no longer be possible; e.g.

    $ sudo /var/dlpx-update/latest/upgrade deferred

### DIFF
--- a/upgrade/upgrade-scripts/common.sh
+++ b/upgrade/upgrade-scripts/common.sh
@@ -197,6 +197,11 @@ function source_upgrade_properties() {
 		die "failed to source: '$UPDATE_DIR/upgrade.properties'"
 }
 
+function remove_upgrade_properties() {
+	rm "$UPDATE_DIR/upgrade.properties" ||
+		die "failed to remove: '$UPDATE_DIR/upgrade.properties'"
+}
+
 function set_upgrade_property() {
 	[[ -n "$1" ]] || die "upgrade property key is missing"
 	[[ -n "$2" ]] || die "upgrade property value is missing"
--- a/upgrade/upgrade-scripts/upgrade
+++ b/upgrade/upgrade-scripts/upgrade
@@ -54,6 +54,7 @@ function usage() {
 	echo "$PREFIX_STRING [-n] [-v] deferred"
 	echo "$PREFIX_SPACES [-n] [-v] full"
 	echo "$PREFIX_SPACES rollback"
+	echo "$PREFIX_SPACES finalize"
 
 	exit 2
 }
@@ -296,7 +297,9 @@ function rollback() {
 
 	source_upgrade_properties
 
-	[[ -n "$UPGRADE_TYPE" ]] || die "variable UPGRADE_TYPE is not set"
+	[[ -n "$UPGRADE_TYPE" ]] ||
+		die "variable UPGRADE_TYPE is not set; is upgrade in progress?"
+
 	[[ -n "$UPGRADE_BASE_CONTAINER" ]] ||
 		die "variable UPGRADE_BASE_CONTAINER is not set"
 	[[ -n "$UPGRADE_BASE_VERSION" ]] ||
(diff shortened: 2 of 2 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2743/)

prakashsurya:
```
$ sudo /var/dlpx-update/latest/upgrade -v deferred
$ sudo zfs list -t all -r rpool
NAME                                                      USED  AVAIL     REFER  MOUNTPOINT
rpool                                                    26.3G  41.0G       64K  none
rpool/ROOT                                               21.2G  41.0G       64K  none
rpool/ROOT/delphix.OL73l4e                               21.2G  41.0G       64K  none
rpool/ROOT/delphix.OL73l4e@execute-upgrade.r66hh4G          0B      -       64K  -
rpool/ROOT/delphix.OL73l4e/data                          71.5M  41.0G     67.2M  legacy
rpool/ROOT/delphix.OL73l4e/data@execute-upgrade.r66hh4G  4.35M      -     67.0M  -
rpool/ROOT/delphix.OL73l4e/home                          11.7G  41.0G     11.7G  legacy
rpool/ROOT/delphix.OL73l4e/home@execute-upgrade.r66hh4G     0B      -     11.7G  -
rpool/ROOT/delphix.OL73l4e/log                            124M  41.0G      121M  legacy
rpool/ROOT/delphix.OL73l4e/log@execute-upgrade.r66hh4G   3.45M      -      110M  -
rpool/ROOT/delphix.OL73l4e/root                          9.28G  41.0G     7.08G  /
rpool/ROOT/delphix.OL73l4e/root@execute-upgrade.r66hh4G  2.20G      -     7.60G  -
rpool/crashdump                                          65.5K  34.7G     65.5K  legacy
rpool/grub                                               3.03M  41.0G     3.03M  legacy
rpool/public                                               64K  41.0G       64K  /public
rpool/update                                             5.16G  24.8G     5.16G  /var/dlpx-update
rpool/upgrade-logs                                         65K  41.0G       65K  /var/tmp/delphix-upgrade

$ sudo zfs list -t all -r rpool
rpool                            24.1G  43.2G       64K  none
rpool/ROOT                       18.9G  43.2G       64K  none
rpool/ROOT/delphix.OL73l4e       18.9G  43.2G       64K  none
rpool/ROOT/delphix.OL73l4e/data  67.2M  43.2G     67.2M  legacy
rpool/ROOT/delphix.OL73l4e/home  11.7G  43.2G     11.7G  legacy
rpool/ROOT/delphix.OL73l4e/log    121M  43.2G      121M  legacy
rpool/ROOT/delphix.OL73l4e/root  7.08G  43.2G     7.08G  /
rpool/crashdump                  65.5K  34.7G     65.5K  legacy
rpool/grub                       3.03M  43.2G     3.03M  legacy
rpool/public                       64K  43.2G       64K  /public
rpool/update                     5.16G  24.8G     5.16G  /var/dlpx-update
rpool/upgrade-logs                 65K  43.2G       65K  /var/tmp/delphix-upgrade
```

prakashsurya:
This change isn't meant to help address the cleanup of snapshots and datasets via the product upgrade logic in the app-gate, it's only meant to address the issues we've seen when updating "dc1"; hence the reason why this change only supports "finalize" for deferred upgrades.

The work to handle this issue in the product upgrade logic will be undertaken in [DLPX-64438](https://jira.delphix.com/browse/DLPX-64438), and that work likely will be tangential to what I'm proposing in this PR.

prakashsurya:
> Edit: I've just seen your last comment. Is the idea to have this logic for the time being to alleviate current issues with dc1 and then reworking it for DLPX-64438?

Yes, that's right.

For now, I wanted to keep this limited to what I know is useful for "dc1", and then we can expand on this later for DLPX-64438, or address that bug in a different way.

pzakha:
Gotcha.

Do you think it would make sense to call it "cleanup" instead of "finalize". I'd image this could be called in a similar fashion to `dx_delete` to cleanup all the stuff that wasn't cleaned up earlier, rather than only the last version (for instance if previous invocations of this script wasn't called or has failed for some reason and didn't perform the full cleanup)?

prakashsurya:
> Do you think it would make sense to call it "cleanup" instead of "finalize".

I don't think so.. I'd recommend we write a different script to do the cleanup like you're thinking, and put that in a package, since this script may not be available when we want to perform the cleanup that you're thinking of.

For example, I could imagine this "finalize" logic also removing the directory in `/var/dlpx-update` where this script lives (perhaps I should add this?), since at that point the upgrade is done and that directory isn't required anymore. Thus, after that, we couldn't call this a second time, since the script will no longer exist.

Additionally, as it stands, this new sub-command cannot be used after it's invoked the first time:
```
/var/dlpx-update/latest/common.sh: line 196: /var/dlpx-update/upgrade.properties: No such file or directory
upgrade: failed to source: '/var/dlpx-update/upgrade.properties'
```

pzakha:
Yeah that makes sense. So we've had some similar concerns recently with dx_delete, where we would always call the script from the OS (equivalent to putting it in packages on Linux), but with the introduction of the new Linux dataset layout, the 5.3 dx_delete script would not know how to cleanup those new datasets (see DLPX-64068).

The solution was to call the script from /var/dlpx-update/ when it existed, or from the OS as the backup. I'm not sure if we want to do something similar here (although taking care of not duplicating code between appliance-build and whatever package we put the new dx_delete into).

In any case, if we are not planning to use the "finalize" on the Delphix Engine outside of dc1, I'm fine with this. As a side note, IIRC some of the upgrade scripts are not used on the Delphix Engine, but I'm not sure which. It might be good to make it more clear in the comments/usage.

prakashsurya:
> As a side note, IIRC some of the upgrade scripts are not used on the Delphix Engine

It depends on what you mean by "not used on the Delphix Engine".. I think all can be used on the engine, e.g. manually run by support, it's just that not all are used by the product logic in the app-gate.

IIRC, currently only `upgrade` is not used by the product logic. The design being, `upgrade` is the script equivalent of the product logic that drives the upgrade (e.g. the script equivalent of `UpgradeManagerImpl`).

> So we've had some similar concerns recently with dx_delete, where we would always call the script from the OS

I may remember incorrectly.. but, I think what you're thinking of is calling `dx_delete` from the "old" root filesystem, to delete the datasets of the "new" root filesystem (e.g. on failure). I don't think this problem exists today, since the upgrade scripts for linux take care of deleting any datasets it creates on failure. So, I believe we're already doing what you suggest.

When we do any cleanup of "old" datasets, e.g. after a sucessful upgrade, we should use the cleanup script from the new version, which we can always ensure is compatible with any prior dataset layout.

I agree we should not be using a cleanup script from the old version, to cleanup the dataset layout of a new version (since the new layout may differ from what the old cleanup script expects); but I don't think we do this on linux.

prakashsurya:
> IIRC, currently only upgrade is not used by the product logic. 

And to clarify, I don't think it's that these scripts _shouldn't_ be used by the product logic.. it's just that we've chosen not to, for various reasons; e.g. the product wants to have more control than what the scripts enable. In the places where the script is sufficient, it is used; e.g. `upgrade rollback` is used by the product's `svc-rollback` script.

And in this case, the product may eventually end up using `finalize` to do it's cleanup; it's just not clear to me if that's the best way to do it yet, or not. So, I'd rather implement only the logic that I _know_ will be useful right now, and leave the rest until we have a better idea of what the product logic will need.

prakashsurya:
> For example, I could imagine this "finalize" logic also removing the directory in /var/dlpx-update where this script lives (perhaps I should add this?)

I've gone ahead and added logic to do this.. so now the directory for the upgrade in `/var/dlpx-update` will be removed after running this new "finalize" sub-command.

PR #419, opened 2020-02-05
Add run-pty subcommand to upgrade-container script

prakashsurya:
This adds a new "run-pty" subcommand to the "ugprade-container" script
which is intended to be used when running interactive commands in a
container; e.g. running a shell. Currently, the "run" subcommand can be
used, but since that does not use the "--pty" option with "systemd-run",
the terminal may not work correctly (e.g. keyboard input, pager output,
etc). Now, when folks want to run interactive commands within a
container, they should use "run-pty" rather than "run", to avoid these
potential issues.

### DIFF
--- a/upgrade/upgrade-scripts/upgrade-container
+++ b/upgrade/upgrade-scripts/upgrade-container
@@ -476,6 +476,13 @@ function run() {
 		--quiet --wait --pipe -- "$@"
 }
 
+function run_pty() {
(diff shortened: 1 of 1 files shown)


prakashsurya:
cc: @ahrens

prakashsurya:
Closing in favor of https://github.com/delphix/appliance-build/pull/420

PR #420, opened 2020-02-05
Add "shell" subcommand to upgrade-container script

prakashsurya:
This adds a new "shell" subcommand to the "upgrade-container" script,
which is intended to enable users to obtain an interactive shell session
within the specified container.

Currently, the "run" subcommand can be used, while specifying the shell
to run (e.g. /bin/bash), but since that does not specify the "--pty"
option with "systemd-run", the terminal may not work correctly (e.g.
keyboard input, pager output, etc).

Now, when folks want to obtain an interactive shell, they should use
this new "shell" subcommand, rather than "run".

### DIFF
--- a/upgrade/upgrade-scripts/upgrade-container
+++ b/upgrade/upgrade-scripts/upgrade-container
@@ -476,6 +476,10 @@ function run() {
 		--quiet --wait --pipe -- "$@"
 }
 
+function shell() {
+	machinectl shell "$CONTAINER"
+}
+
 function convert_to_rootfs() {
 	#
 	# We're relying on the "mountpoint" property for the "data" and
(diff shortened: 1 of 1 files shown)


prakashsurya:
This is an alternative to #419; from the `systemd-run` manpage, it recommends we use this approach, rather than using `--pty` with `systemd-run`:
```
           Note that machinectl(1)'s shell command is usually a better alternative for requesting a new, interactive login session on the local host or a
           local container.
```

prakashsurya:
It looks like it still gives me a root shell.. I ran this starting as the `delphix` use on the host:
```
$ machinectl shell delphix.Ho8LNEi
==== AUTHENTICATING FOR org.freedesktop.machine1.shell ===
Authentication is required to acquire a shell in a local container.
Authenticating as: root
Password: 
==== AUTHENTICATION COMPLETE ===
Connected to machine delphix.Ho8LNEi. Press ^] three times within 1s to exit session.
root@localhost:~# systemd-detect-virt
systemd-nspawn
root@localhost:~# whoami
root
```

prakashsurya:
I did have to enter the `delphix` user's password, though.. it looks like that's just to get authentication to run `machinectl` as root.. similar to how you could run `systemctl restart ...` as non-root, and it'll ask for a password.

PR #422, opened 2020-02-10
DLPX-68440 remove server.conf and demo image from appliance-build

grwilson:


grwilson:
This is appliance-build changes required for `DLPX-68182 All images need lockdown functionality`.  The [ab-pre-push](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2848/) combines both changes for testing purposes.

grwilson:
@prakashsurya I've already removed the demo image from the devops logic and that change has been pushed. We should no longer build the demo image and the release/stage logic is no longer expecting them. Are there other jobs you're thinking of?

prakashsurya:
Nope, that's all I was thinking of. Sounds like we're good, thank you!

live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
```diff
@@ -23,29 +23,6 @@
     path: /var/opt/delphix
     state: directory
 
-#
-# The "product", "platform", and "cr_auth" fields are intentionally left
-# out of this file, as those fields should get filled in by other roles
-# that will run after this role.
-#
-- copy:
-    dest: /var/opt/delphix/server.conf
```
pzakha: Do we want to also cleanup this file if it already exists to avoid confusing people in the future? We can do this in another PR though.

PR #423, opened 2020-02-11
WIP: DLPX-66313 files previously installed into /etc by packages may not be removed on upgrade

grwilson:


grwilson:
This is a WIP and I'm looking for feedback on this approach for dealing with configuration files. This doesn't do anything to change the debiain package info (i.e. the file will show up even after it's removed).

grwilson:
I tested the obsolete listing logic on my 6.0.0.0 to 6.0.1.0 upgrade and it detects these files:
```
/etc/systemd/system/systemd-resolved.service.d/override.conf
/etc/vmware-tools/guestproxy-ssl.conf
```
On dc1 it detects:
```
/etc/systemd/system/systemd-resolved.service.d/override.conf
/etc/systemd/system/nfs-mountd.service.d/override.conf
/etc/vmware-tools/guestproxy-ssl.conf
```

grwilson:
Another option would be to try to only fix this for `delphix-platform` and follow something like what is outlined here: https://wiki.debian.org/DpkgConffileHandling

prakashsurya:
We've addressed this in #425 so I'll close this.

PR #425, opened 2020-02-11
Remove obsolete configuration files on upgrade

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2866/)

prakashsurya:
This is a continuation of @grwilson's work in #423

upgrade/upgrade-scripts/execute
```diff
@@ -323,6 +323,56 @@ zcat "/usr/share/doc/delphix-entire-$platform/packages.list.gz" |
...
+		# Since we run this each time a file is removed, if a
+		# package has multiple obsolete files, we'll end up
+		# reinstalling the package multiple times. Currently
+		# this isn't a problem, since we haven't hit a case
+		# where a package has multiple obsolete files; but
+		# perhaps this will change in the future, at which point
+		# we can optimize this better to handle that case.
+		#
+		apt_get install -y --reinstall "$package" ||
+			die "failed to reinstall package '$package'"
```
pzakha: Does `--reinstall` require the package deb to be available? While most of our package debs should be available during upgrade this would not be the case for packages pinned by `delphix-kernel`.

From my understanding, we "reinstall" the package just to remove those files from the "obsolete" list, as such should we just make this a warning instead of a failure?

upgrade/upgrade-scripts/execute
```diff
@@ -323,6 +323,56 @@ zcat "/usr/share/doc/delphix-entire-$platform/packages.list.gz" |
...
+		#
+		# We should only be operating on currently installed
+		# packages. So, to verify the package retrieved above,
+		# we double check that the package is installed.
+		#
+		dpkg-query -W "$package" &>/dev/null ||
+			die "package '$package' is not installed"
+
+		rm -f "$file" ||
+			die "failed to remove file '$file' of package '$package'"
```
pzakha: Could we log each file that is being removed?

upgrade/upgrade-scripts/execute
```diff
@@ -323,6 +323,56 @@ zcat "/usr/share/doc/delphix-entire-$platform/packages.list.gz" |
...
+
+		#
+		# Since we run this each time a file is removed, if a
+		# package has multiple obsolete files, we'll end up
+		# reinstalling the package multiple times. Currently
+		# this isn't a problem, since we haven't hit a case
+		# where a package has multiple obsolete files; but
```
pzakha: This could potentially be a big performance hit if we have to reinstall the delphix-virtualization package several times. That said, right now it only has 2 etc files:
```
/etc/sudoers.d/delphix-cli
/etc/fluent
```

upgrade/upgrade-scripts/execute
prakashsurya: We're already doing that; bash's `xtrace` output is captured in syslog, which will contain this command (containing the file being removed).

upgrade/upgrade-scripts/execute
prakashsurya: > Does --reinstall require the package deb to be available? While most of our package debs should be available during upgrade this would not be the case for packages pinned by delphix-kernel.

Hm, yea, I think this is a good point. I'll have to think about this case a little more...

> From my understanding, we "reinstall" the package just to remove those files from the "obsolete" list, as such should we just make this a warning instead of a failure?

Can you give me an example where we'd expect this to fail?

If there aren't examples where we expect this to fail, I'd rather not silently ignore the failure, without a good understanding about why it's safe/correct to ignore the failure. Just because we _can_ ignore the failure, doesn't mean  we _should_.

prakashsurya: I think the point about not being able to `--reinstall` packages that aren't in the upgrade image is valid, but I'm not sure if this will ever be a practical concern.. I think that a conffile will only become obsolete after a package is upgraded, and in order for the package to be upgraded it has to be contained in the upgrade image.. so, I think, any package that we find with an obsoleted confile _must_ be in the upgrade image, so we _must_ be able to `--reinstall` it.

By that logic, for the "old packages" pinned by "delphix-kernel", we will not have upgraded those packages, so I don't think we'll find any of their conffiles as being obsolete.. and thus, we won't attempt to `--reinstall` any of them.

What do you think? Am I mistaken, or overlooking anything?

pzakha: Following our conversation offline, I think it's fine to leave it as an error as chances that we would be trying to reinstall packages related to delphix-kernel are very small and a failure to reinstall could have potentially bad consequences, so it would be better that we catch it as an error in the upgrade container.

prakashsurya: > chances that we would be trying to reinstall packages related to delphix-kernel are very small

For posterity and just to be clear, I'm claiming the chances are _zero_ rather than _small_, and I think the distinction is important.

PR #433, opened 2020-02-20
Fix "upgrade-container stop" return code on success

prakashsurya:
This change fixes a regression introduced in acb60f4c. Without this
change, calling "upgrade-container stop" will return an error code, even
when the call successfully stops the container. With this change, when
the container is stopped successfully, we'll now properly return 0.

### DIFF
--- a/upgrade/upgrade-scripts/upgrade-container
+++ b/upgrade/upgrade-scripts/upgrade-container
@@ -420,6 +420,13 @@ function stop() {
 
 	machinectl status "$CONTAINER" &>/dev/null &&
 		die "timeout waiting for container termination: '$CONTAINER'"
+
+	#
+	# We don't want to return the error code from the call to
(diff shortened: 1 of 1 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2931/)

prakashsurya:
Without this change:
```
$ sudo /var/dlpx-update/latest/upgrade-container start delphix.MkW1ax9
$ sudo /var/dlpx-update/latest/upgrade-container stop delphix.MkW1ax9
$ echo $?
1
```

With this change:
```
$ sudo /var/dlpx-update/latest/upgrade-container start delphix.MkW1ax9
$ sudo /var/dlpx-update/latest/upgrade-container stop delphix.MkW1ax9
$ echo $?
0
```

PR #434, opened 2020-02-20
Fix "upgrade-container stop" return code on success (#433)

prakashsurya:
This change fixes a regression introduced in acb60f4c. Without this
change, calling "upgrade-container stop" will return an error code, even
when the call successfully stops the container. With this change, when
the container is stopped successfully, we'll now properly return 0.

### DIFF
--- a/upgrade/upgrade-scripts/upgrade-container
+++ b/upgrade/upgrade-scripts/upgrade-container
@@ -420,6 +420,13 @@ function stop() {
 
 	machinectl status "$CONTAINER" &>/dev/null &&
 		die "timeout waiting for container termination: '$CONTAINER'"
+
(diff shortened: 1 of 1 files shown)


prakashsurya:
This is a clean cherry-pick of #433.

`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2932/)

prakashsurya:
This isn't critical, since the app-gate will not fail when "upgrade-container stop" returns an error; rather, it will simply emit a line in the log, and continue on. Later, after "stop" fails, it will still attempt to "destroy" the container which will succeed.

PR #442, opened 2020-03-05
Incorrect handling for certain boot devices names

prakashsurya:
The "get_bootloader_devices" function attempts to list all devices
attached to "rpool" that are valid devices for storing the bootloader.
Since the devices used for "rpool" are partitions, it has to convert the
partion name (as returned by "zpool list") to the underlying device
name.

Unfortunately, the logic to do that mapping did not properly support
device partiations with "p" in the name; e.g. "nvme0n1p1" should map to
"nvme0n1", and "sda1" to "sda". This change updates this logic to better
support partitions names, regardless of it containing a "p" in the
partiation name, or not.

### DIFF
--- a/upgrade/upgrade-scripts/rootfs-container
+++ b/upgrade/upgrade-scripts/rootfs-container
@@ -77,7 +77,7 @@ function get_bootloader_devices() {
 	#
 	zpool list -vH rpool |
 		awk '! /rpool|mirror|replacing|spare/ {print $1}' |
-		sed 's/[0-9]*$//'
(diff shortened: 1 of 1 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3014/)

prakashsurya:
@pzakha I opened a bug pointing to your suggestion: https://jira.delphix.com/browse/DLPX-68846

PR #445, opened 2020-03-10
DLPX-68157 Update 'finalize' logic to clean up snapshots older than c…

AdityaPrasad-dlpx:
Wanted to get some reviews for this change. This change modifies Prakash's implementation of `upgrade finalize` in the upgrade scripts. With the new logic, only snapshots older than the previous snapshot (this being the snapshot of the version we just upgraded from) are deleted. Currently the `finalize` logic still only supports deferred upgrades. The logic for post-rollback cleanup will be developed following this.

JIRA: https://jira.delphix.com/browse/DLPX-68157

Dependent review for app-gate changes: http://reviews.delphix.com/r/55544/

git ab-pre-push (SUCCESS): http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3099/console

---------------------------------
Manual Testing:
---------------------------------
```
Before deferred upgrade 
$ sudo zfs list -t all -r rpool
NAME                                                      USED  AVAIL     REFER  MOUNTPOINT
rpool                                                    33.9G  33.4G       64K  none
rpool/ROOT                                               25.7G  33.4G       64K  none
rpool/ROOT/delphix.PLOfKjK                               25.7G  33.4G       65K  none
rpool/ROOT/delphix.PLOfKjK@execute-upgrade.pPc3ac3          0B      -       65K  -
rpool/ROOT/delphix.PLOfKjK/data                          9.77M  33.4G     9.30M  legacy
rpool/ROOT/delphix.PLOfKjK/data@execute-upgrade.pPc3ac3   476K      -     7.75M  -
rpool/ROOT/delphix.PLOfKjK/home                          14.8G  33.4G     14.8G  legacy
rpool/ROOT/delphix.PLOfKjK/home@execute-upgrade.pPc3ac3   167K      -     11.9G  -
rpool/ROOT/delphix.PLOfKjK/log                           51.3M  33.4G     50.2M  legacy
rpool/ROOT/delphix.PLOfKjK/log@execute-upgrade.pPc3ac3   1.05M      -     32.8M  -
rpool/ROOT/delphix.PLOfKjK/root                          10.8G  33.4G     5.99G  /
rpool/ROOT/delphix.PLOfKjK/root@execute-upgrade.pPc3ac3  4.82G      -     6.02G  -
rpool/crashdump                                          65.5K  33.4G     65.5K  legacy
rpool/grub                                               3.03M  33.4G     3.03M  legacy
rpool/public                                               64K  33.4G       64K  /public
rpool/update                                             8.23G  21.8G     8.23G  /var/dlpx-update
rpool/upgrade-logs                                         65K  33.4G       65K  /var/tmp/delphix-upgrade
———————————————————————
$ sudo /var/dlpx-update/latest/upgrade -v deferred
$ sudo zfs list -t all -r rpool
NAME                                                      USED  AVAIL     REFER  MOUNTPOINT
rpool                                                    38.7G  28.6G       64K  none
rpool/ROOT                                               30.4G  28.6G       64K  none
rpool/ROOT/delphix.PLOfKjK                               30.4G  28.6G       65K  none
rpool/ROOT/delphix.PLOfKjK@execute-upgrade.pPc3ac3          0B      -       65K  -
rpool/ROOT/delphix.PLOfKjK@execute-upgrade.RNNPqan          0B      -       65K  -
rpool/ROOT/delphix.PLOfKjK/data                          10.4M  28.6G     9.45M  legacy
rpool/ROOT/delphix.PLOfKjK/data@execute-upgrade.pPc3ac3   476K      -     7.75M  -
rpool/ROOT/delphix.PLOfKjK/data@execute-upgrade.RNNPqan   458K      -     9.38M  -
rpool/ROOT/delphix.PLOfKjK/home                          14.8G  28.6G     14.8G  legacy
rpool/ROOT/delphix.PLOfKjK/home@execute-upgrade.pPc3ac3   167K      -     11.9G  -
rpool/ROOT/delphix.PLOfKjK/home@execute-upgrade.RNNPqan     0B      -     14.8G  -
rpool/ROOT/delphix.PLOfKjK/log                           81.7M  28.6G     79.4M  legacy
rpool/ROOT/delphix.PLOfKjK/log@execute-upgrade.pPc3ac3   1.05M      -     32.8M  -
rpool/ROOT/delphix.PLOfKjK/log@execute-upgrade.RNNPqan   1.27M      -     63.6M  -
rpool/ROOT/delphix.PLOfKjK/root                          15.5G  28.6G     6.10G  /
rpool/ROOT/delphix.PLOfKjK/root@execute-upgrade.pPc3ac3  4.82G      -     6.02G  -
rpool/ROOT/delphix.PLOfKjK/root@execute-upgrade.RNNPqan  4.42G      -     5.99G  -
rpool/crashdump                                          65.5K  28.6G     65.5K  legacy
rpool/grub                                               3.03M  28.6G     3.03M  legacy
rpool/public                                               64K  28.6G       64K  /public
rpool/update                                             8.30G  21.7G     8.30G  /var/dlpx-update
rpool/upgrade-logs                                         65K  28.6G       65K  /var/tmp/delphix-upgrade
———————————————————————
$ sudo zfs list -t all -r rpool
NAME                                                      USED  AVAIL     REFER  MOUNTPOINT
rpool                                                    25.9G  41.4G       64K  none
rpool/ROOT                                               25.6G  41.4G       64K  none
rpool/ROOT/delphix.PLOfKjK                               25.6G  41.4G       65K  none
rpool/ROOT/delphix.PLOfKjK@execute-upgrade.RNNPqan          0B      -       65K  -
rpool/ROOT/delphix.PLOfKjK/data                          10.3M  41.4G     9.75M  legacy
rpool/ROOT/delphix.PLOfKjK/data@execute-upgrade.RNNPqan   574K      -     9.38M  -
rpool/ROOT/delphix.PLOfKjK/home                          14.8G  41.4G     14.8G  legacy
rpool/ROOT/delphix.PLOfKjK/home@execute-upgrade.RNNPqan   147K      -     14.8G  -
rpool/ROOT/delphix.PLOfKjK/log                           81.0M  41.4G     79.8M  legacy
rpool/ROOT/delphix.PLOfKjK/log@execute-upgrade.RNNPqan   1.27M      -     63.6M  -
rpool/ROOT/delphix.PLOfKjK/root                          10.7G  41.4G     6.10G  /
rpool/ROOT/delphix.PLOfKjK/root@execute-upgrade.RNNPqan  4.59G      -     5.99G  -
rpool/crashdump                                           299M  34.5G      299M  legacy
rpool/grub                                               3.03M  41.4G     3.03M  legacy
rpool/public                                               64K  41.4G       64K  /public
rpool/update                                               88K  30.0G       88K  /var/dlpx-update
rpool/upgrade-logs                                         65K  41.4G       65K  /var/tmp/delphix-upgrade

```

### DIFF
--- a/upgrade/upgrade-scripts/upgrade
+++ b/upgrade/upgrade-scripts/upgrade
@@ -379,32 +379,69 @@ function finalize() {
 
 	[[ -n "$UPGRADE_TYPE" ]] ||
 		die "variable UPGRADE_TYPE is not set; is upgrade in progress?"
+	[[ -n "$UPGRADE_BASE_VERSION" ]] ||
+		die "variable UPGRADE_BASE_VERSION is not set"
 
 	case "$UPGRADE_TYPE" in
-	DEFERRED)
-		#
-		# Currently, we've only implemented "finalize" for
-		# DEFERRED upgrades. This is to avoid adding the logic
-		# for "not-in-place" upgrades (DEFERRED is strictly
-		# "in-place") and rollback (which is not supported for
-		# DEFERRED), until we determine those are required.
-		#
-		;;
+	DEFERRED | FULL | ROLLBACK) ;;
 	*)
-		die "rollback is not supported for upgrade type: '$UPGRADE_TYPE'"
+		die "finalize is not supported for upgrade type: '$UPGRADE_TYPE'"
 		;;
 	esac
 
-	ROOTFS_DATASET=$(get_mounted_rootfs_container_dataset)
-	[[ -n "$ROOTFS_DATASET" ]] ||
-		die "unable to determine mounted rootfs container dataset"
+	#
+	# This first pass checks for available "execute.upgrade" snapshots
+	# and deletes snapshots that are older than the base upgrade version.
+	# This pass takes care of cleanup after in-place and deferred upgrades.
+	#
+	SNAPSHOT_LIST=$(zfs list -t snapshot -r rpool/ROOT -d 2 -Ho name) ||
+		die "unable to determine available rollback snapshots"
 
-	SNAPSHOT_NAME=$(get_dataset_rollback_snapshot_name "$ROOTFS_DATASET")
-	[[ -n "$SNAPSHOT_NAME" ]] ||
-		die "unable to determine rollback snapshot name"
+	for snapshot in $SNAPSHOT_LIST; do
+		local SNAPSHOT_NAME
+		SNAPSHOT_NAME=$(echo "$snapshot" | awk -F@ '{print $2}')
+		#
+		# We skip container-delphix instances since `rootfs-container delete`
+		# handles them in the second pass below.
+		#
+		grep -qE "^container-delphix.[[:alnum:]]{7}$" <(echo "$SNAPSHOT_NAME") &&
+			continue
+		grep -qE "^execute-upgrade.[[:alnum:]]{7}$" <(echo "$SNAPSHOT_NAME") ||
+			die "unexpected snapshot name: '$snapshot'"
+		local SNAPSHOT_VERSION
+		SNAPSHOT_VERSION=$(zfs get -Hpo value "$PROP_CURRENT_VERSION" "$snapshot") ||
+			die "failed to get snapshot version"
+		[[ "$SNAPSHOT_VERSION" != "-" ]] ||
+			die "failed to get current version for snapshot '$snapshot'"
+		if compare_versions \
+			"$SNAPSHOT_VERSION" "lt" "$UPGRADE_BASE_VERSION"; then
+			zfs destroy -r "$snapshot" ||
+				die "failed to destroy rollback snapshot"
+		fi
+	done
 
-	zfs destroy -r "$ROOTFS_DATASET@$SNAPSHOT_NAME" ||
-		die "failed to destroy rollback snapshots"
+	#
+	# This second pass checks for available filesystems and deletes those
+	# that are older than the base upgrade version. This pass takes care
+	# of cleanup following rollbacks and not-in-place upgrades.
+	#
+	FILESYSTEM_LIST=$(zfs list -t filesystem -r rpool/ROOT -d 1 -Ho name | tail -n +2) ||
+		die "unable to determine available filesystems"
+
+	for filesystem in $FILESYSTEM_LIST; do
+		local FILESYSTEM_VERSION
+		FILESYSTEM_VERSION=$(zfs get -Hpo value "$PROP_CURRENT_VERSION" "$filesystem") ||
+			die "failed to get filesystem version"
+		[[ "$FILESYSTEM_VERSION" != "-" ]] ||
+			die "failed to get current version for filesystem '$filesystem'"
+		if compare_versions \
+			"$FILESYSTEM_VERSION" "lt" "$UPGRADE_BASE_VERSION"; then
+			local FILESYSTEM_NAME
+			FILESYSTEM_NAME=$(echo "$filesystem" | awk -F/ '{print $3}')
+			"$IMAGE_PATH/rootfs-container" delete "$FILESYSTEM_NAME" ||
+				die "failed to delete filesystem '$filesystem"
+		fi
+	done
 
 	rm -rf "$IMAGE_PATH" || die "failed to remove unpacked upgrade image"
 
@@ -434,12 +471,14 @@ deferred)
 	shift 1
 	verify_upgrade_is_allowed
 	verify_upgrade_in_place_is_allowed
+	verify_upgrade_not_in_progress
 	upgrade_in_place "$@"
 	;;
 full)
 	UPGRADE_TYPE="FULL"
 	shift 1
 	verify_upgrade_is_allowed
+	verify_upgrade_not_in_progress
 
 	#
 	# FULL upgrade always perform a reboot but can take on two
--- a/upgrade/upgrade-scripts/common.sh
+++ b/upgrade/upgrade-scripts/common.sh
@@ -288,3 +288,9 @@ function set_upgrade_property() {
 	source_upgrade_properties ||
 		die "failed to read properties file after setting '$1=$2'"
 }
+
+function verify_upgrade_not_in_progress() {
+	. "$UPDATE_DIR/upgrade.properties"
+
+	[[ -z "$UPGRADE_TYPE" ]] || die "upgrade currently in-progress"
+}


upgrade/upgrade-scripts/upgrade
```diff
@@ -395,16 +397,21 @@ function finalize() {
...
-	[[ -n "$ROOTFS_DATASET" ]] ||
-		die "unable to determine mounted rootfs container dataset"
-
-	SNAPSHOT_NAME=$(get_dataset_rollback_snapshot_name "$ROOTFS_DATASET")
-	[[ -n "$SNAPSHOT_NAME" ]] ||
-		die "unable to determine rollback snapshot name"
-
-	zfs destroy -r "$ROOTFS_DATASET@$SNAPSHOT_NAME" ||
-		die "failed to destroy rollback snapshots"
+	SNAPSHOT_LIST=$(zfs list -t snapshot -r rpool/ROOT -d 2 -Ho name) ||
```
prakashsurya: We're not filtering the list of snapshots to _only_ the snapshot names that we expect (i.e. `execute-upgrade.XXXXXXX`). I think this is fine, since I don't believe we take any other snapshots on these datasets, but I just wanted to make a note of it.

Additionally, I think not having any filtering logic is preferable for now, since we wouldn't have any way to test the filtering (since we don't have any other snapshot names that'd we want to filter out).

upgrade/upgrade-scripts/upgrade
```diff
@@ -395,16 +397,21 @@ function finalize() {
...
+		die "unable to determine available rollback snapshots"
+
+	for snapshot in $SNAPSHOT_LIST; do
```
prakashsurya: Since we're assuming each snapshot in the list here, is one of the `execute-upgrade.XXXXXXX` snapshots.. would it be reasonable to add some verification of the snapshot name in the block below? e.g. something like this:
```
grep -qE "^execute-upgrade.[[:alnum:]]{7}$" <<< "$snapshot" || die "unexpected snapshot name: '$snapshot'"
```
This might help us catch the case where/if we start adding other snapshots to the container dataset(s), but forget to update this cleanup code.. in that case, this check would fail, so we'd (hopefully) more quickly notice the problem.

upgrade/upgrade-scripts/upgrade
```diff
@@ -395,16 +397,21 @@ function finalize() {
...
+		local SNAPSHOT_VERSION
+		SNAPSHOT_VERSION=$(zfs get -Hpo value "$PROP_CURRENT_VERSION" "$snapshot") ||
+			die "failed to get snapshot version"
+		compare_versions \
+			"$SNAPSHOT_VERSION" "lt" "$UPGRADE_BASE_VERSION"
+		local rc="$?"
+		if [[ "$rc" == 0 ]]; then
```
prakashsurya: Could we remove the `rc` variable here, and more simply do this:
```
if [[ "$?" == 0 ]]; then
```
it's not a big deal, but I don't see where `rc` is used elsewhere, so it might be reasonable to remove it..

Along the same lines, we could probably do this too:
```
if compare_versions \
    "$SNAPSHOT_VERSION" "lt" "$UPGRADE_BASE_VERSION"; then
```
IMO, this would be best, since it would remove our use of `$?` as well.. the use of `$?` can be problematic if code is mistakenly added (later) in between the call to `compare_versions` and the use of `$?`.

upgrade/upgrade-scripts/upgrade
AdityaPrasad-dlpx: That makes sense, I'll keep that in mind for future developments.

upgrade/upgrade-scripts/upgrade
AdityaPrasad-dlpx: That's a good idea, I can include that

upgrade/upgrade-scripts/upgrade
AdityaPrasad-dlpx: Will do, that's a good point

upgrade/upgrade-scripts/upgrade
```diff
@@ -395,16 +397,21 @@ function finalize() {
...
+		die "unable to determine available rollback snapshots"
+
+	for snapshot in $SNAPSHOT_LIST; do
+		grep -qE "^execute-upgrade.[[:alnum:]]{7}$" <(echo "$snapshot" | awk -F@ '{print $2}') ||
+			die "unexpected snapshot name: '$snapshot'"
+		local SNAPSHOT_VERSION
+		SNAPSHOT_VERSION=$(zfs get -Hpo value "$PROP_CURRENT_VERSION" "$snapshot") ||
+			die "failed to get snapshot version"
+		if compare_versions \
+			"$SNAPSHOT_VERSION" "lt" "$UPGRADE_BASE_VERSION"; then
```
prakashsurya: It's worth noting that.. since we're comparing versions here, if we end up "upgrading" to an older version, we won't end up removing the snapshots of that previous version..

e.g. I can simulate this by cloning a trunk VM, and then "upgrading" to a upgrade image that was generated earlier than that trunk VM.. after doing something like that, I wind up with datasets looking like this:
```
rpool/ROOT/delphix.SVehJOD                               com.delphix:current-version  6.1.0.0-snapshot.20200310170304292+jenkins-selfservice-devops-gate-master-appliance-build-master-pre-push-4447  local
rpool/ROOT/delphix.SVehJOD@execute-upgrade.NXkUlTF       com.delphix:current-version  6.1.0.0-snapshot.20200311085025156+jenkins-ops-devops-gate-master-appliance-build-master-post-push-2728         local
```

i.e. `6.1.0.0-snapshot.20200311085025156+jenkins-ops-devops-gate-master-appliance-build-master-post-push-2728` was the original version of the trunk VM before upgrade.. then I upgraded using an upgrade image containing version `6.1.0.0-snapshot.20200310170304292+jenkins-selfservice-devops-gate-master-appliance-build-master-pre-push-4447`

After performing the upgrade, I was now running version `6.1.0.0-snapshot.20200310170304292+jenkins-selfservice-devops-gate-master-appliance-build-master-pre-push-4447`, but since that version number is actually less than what I started with.. when I later call `finalize`, it won't remove the snapshot.
 
I'm not sure it's worth changing the code you have here to account for such a use case, since I don't think a customer system will ever run into this issue (i.e. I don't think we'll ever upgrade to a version that's less than what's currently running, on a customer's system), but I do think it's something to be aware of..

upgrade/upgrade-scripts/upgrade
sanbhara: May I suggest we warn instead of die here. Feels like an unexpected snapshot should not prevent us from completing the operation. Another concern is if we do die here, the stackStartup will hit a DFE.

upgrade/upgrade-scripts/upgrade
sanbhara: @prakashsurya I believe the API route of upgrade does not allow for upload and unpack of older version image. Was not aware that the scripts provide a way to do this which is good to know for some internal testing situations.

upgrade/upgrade-scripts/upgrade
prakashsurya: > May I suggest we warn instead of die here.

I disagree. IMO, warnings are generally useless, since they mostly go ignored. So, if we went that route, I'd opt to do nothing, rather than warn (we will already log the removal command in syslog).

With that said, the point of this check is to avoid removing a snapshots that we didn't not expect to remove; e.g. avoid "data corruption" where we accidentally remove a snapshot we should not have removed. So, in that case, a warning will not prevent us from removing that unexpected snapshot.

Additionally, we should never hit this case. If we do, it's a bug, and I'd rather halt the operation immediately, rather than assume it's OK to remove an arbitrary snapshot we weren't expecting and carry on.

> Another concern is if we do die here, the stackStartup will hit a DFE.

This is my intention.

upgrade/upgrade-scripts/upgrade
```diff
@@ -395,16 +397,21 @@ function finalize() {
...
-	zfs destroy -r "$ROOTFS_DATASET@$SNAPSHOT_NAME" ||
-		die "failed to destroy rollback snapshots"
+	SNAPSHOT_LIST=$(zfs list -t snapshot -r rpool/ROOT -d 2 -Ho name) ||
+		die "unable to determine available rollback snapshots"
+
+	for snapshot in $SNAPSHOT_LIST; do
+		grep -qE "^execute-upgrade.[[:alnum:]]{7}$" <(echo "$snapshot" | awk -F@ '{print $2}') ||
+			die "unexpected snapshot name: '$snapshot'"
+		local SNAPSHOT_VERSION
+		SNAPSHOT_VERSION=$(zfs get -Hpo value "$PROP_CURRENT_VERSION" "$snapshot") ||
```
prakashsurya: After setting `SNAPSHOT_VERSION`, we should probably verify a version was obtained, since by default it will successfully return `-`.. e.g.
```
[[ "$SNAPSHOT_VERSION" != "-" ]] || die "failed to get current version for snapshot '$snapshot'"
```

upgrade/upgrade-scripts/upgrade
prakashsurya: :+1: OK, good to know.

> Was not aware that the scripts provide a way to do this

Yea, the underlying logic (i.e. the scripts) try to allow "upgrading" to any upgrade image, no matter the "from" and "to" versions. This was intentional, so the logic to implement policy (e.g. one can only upgrade to a "newer" version) could be implemented at the higher layers; which sounds like exactly what we're doing in the API, so that's good to know.

upgrade/upgrade-scripts/upgrade
sanbhara: Hmm my intention was we would warn and skip that snapshot but I see your point about it being a bug from a product standpoint. Are you aware if support engineers ever take snapshots? say when installing hotfixes or doing some manual intervention on the engine.

prakashsurya: > Hmm my intention was we would warn and skip that snapshot

I'd be more OK with that approach, but I still think it's a little strange to skip the snapshot without having some prior knowledge about what the snapshot would contain, and why it'd be better to skip it rather than fail.

> Are you aware if support engineers ever take snapshots?

I am not, but I also haven't tried to ask around to find out.. this might be a good question for Anthony.

upgrade/upgrade-scripts/upgrade
AdityaPrasad-dlpx: I see, let me include that

upgrade/upgrade-scripts/upgrade
AdityaPrasad-dlpx: I just asked Anthony for some input, will update you on the response

AdityaPrasad-dlpx: Anthony's response
```
Hey Aditya, not normally. It's possible we could if we're looking into a specific issue just to preserve the state but we try not to do anything the stack wouldn't be aware of.
So it would probably be fairly short-lived if we did it.
```

sanbhara: Good to know. Thanks. Changes look good.

prakashsurya:
@AdityaPrasad-dlpx is this ready to land? I can do that, just let me know when you're ready.

AdityaPrasad-dlpx:
@prakashsurya I'm planning to do one final test that Sumedh suggested:
**A** --(full)-> **B** --(full)-> **C** --(rollback)-> **B**
just to check if there are any regressions.

upgrade/upgrade-scripts/upgrade
```diff
@@ -379,6 +379,8 @@ function finalize() {
 
 	[[ -n "$UPGRADE_TYPE" ]] ||
 		die "variable UPGRADE_TYPE is not set; is upgrade in progress?"
+	[[ -n "$UPGRADE_BASE_VERSION" ]] ||
+		die "variable UPGRADE_BASE_VERSION is not set"
 
 	case "$UPGRADE_TYPE" in
```
prakashsurya: Does this change work properly if a FULL upgrade (without a call to finalize) is performed first, then a DEFERRED upgrade (with a call to finalize) is performed next? How about a DEFERRED, then FULL, then DEFERRED? Just wondering if there's any potential issues that may occur, if a FULL upgrade was attempted at some point in the past upgrade history of the engine.

AdityaPrasad-dlpx: I can test these out as well

AdityaPrasad-dlpx:
I included changes to handle container/dataset cleanup (presumably would occur after rollbacks and not-in-place upgrades). 

Testing in progress

PR #447, opened 2020-03-13
DLPX-68846 logic for determining device path for grub operations duri…

pcd1193182:
…ng upgrade is fragile

## Pull request checklist

Please check if your PR fulfills the following requirements:
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [x] Build was run locally and any changes were pushed
- [x] Lint has passed locally and any fixes were made for failures

## Pull request type

 

 

Please check the type of change your PR introduces:
- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe): 

## What is the current behavior?
Currently, the logic used during upgrade to determine the disk that a partition is part of is fragile. We perform a purely regex-based manipulation, which is prone to failure on more complex device names.

Issue Number: DLPX-68846

## What is the new behavior?
A better approach is to leverage the lsblk command to directly determine the parent of the partition.

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

Tested manually on ESX, GCP, Azure, and AWS.
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3057/console

### DIFF
--- a/upgrade/upgrade-scripts/rootfs-container
+++ b/upgrade/upgrade-scripts/rootfs-container
@@ -77,7 +77,17 @@ function get_bootloader_devices() {
 	#
 	zpool list -vH rpool |
 		awk '! /rpool|mirror|replacing|spare/ {print $1}' |
-		sed 's/p\{0,1\}[0-9]*$//'
+		while read -r part; do
+			#
+			# If the rpool is not installed a parition, we throw
+			# an error. We expect this to never happen, and the
+			# calling code is likely untested in that case, so we
+			# throw an error rather than try to handle it.
+			#
+			[[ "$(lsblk --nodeps -no type "/dev/$part")" == "part" ]] ||
+				die "rpool installed on full disk \"$part\""
+			lsblk -no pkname "/dev/$part"
+		done
 }
 
 function set_bootfs_not_mounted_cleanup() {


upgrade/upgrade-scripts/rootfs-container
```diff
@@ -77,7 +77,14 @@ function get_bootloader_devices() {
 	#
 	zpool list -vH rpool |
 		awk '! /rpool|mirror|replacing|spare/ {print $1}' |
-		sed 's/p\{0,1\}[0-9]*$//'
+		while read -r part; do
+			# If it doesn't end with a number, it's already a full disk
+			if [[ "$part" =~ [a-z]$ ]]; then
+				echo "$part"
```
prakashsurya: presumably this is done because `lsblk` doesn't work when the input is not a partition?

IMO, rather than trying to support this case, I think it might be better to throw an error. I say that because for our (non-migrated) Linux systems, rpool will always use a partition, so this code path won't ever be tested/used. We'd need to verify this for migrated systems.

What do you think? I'm just hesitant to add a branch that'll never get exercised.

pcd1193182: I think that makes sense, I'll switch this over to throwing an error.

prakashsurya: Experimenting a little, my suggestion is to unconditionally call:
```
lsblk --nodeps -no pkname "/dev/$part"
```
and then throw an error if that returns an empty string..

For example:
```
$ lsblk --nodeps -no pkname /dev/sda

$ lsblk --nodeps -no pkname /dev/sda1
sda
```

I'm assuming a migrated system also uses partitions for rpool, just like a non-migrated system. If a migrated system does not use partitions for rpool, then perhaps we could instead use `--nodeps -no type` to determine if it's a partition or not (rather than simply checking the string name):
```
$ lsblk --nodeps -no type /dev/sda1
part
$ lsblk --nodeps -no type /dev/sda
disk
```

pcd1193182: I checked on a migrated VM, our root pool is also installed on a partition in that case.

upgrade/upgrade-scripts/rootfs-container
```diff
@@ -77,7 +77,17 @@ function get_bootloader_devices() {
...
 		awk '! /rpool|mirror|replacing|spare/ {print $1}' |
-		sed 's/p\{0,1\}[0-9]*$//'
+		while read -r part; do
+			#
+			# If it doesn't end with a number, it's already a full
+			# disk. Since we expect this to never happen, and the
+			# calling code is likely untested in that case, we
+			# throw an error rather than try to handle it.
+			#
+			[[ "$part" =~ [a-z]$ ]] &&
```
prakashsurya: I don't feel strongly, but I think this check might be a little more robust if we did something like:
```
[[ "$(lsblk --nodeps -no type "/dev/$part")" == "part" ]] ||
```

PR #450, opened 2020-03-18
DLPX-67726 com.delphix:current-version property is not updated for al…

sumedhbala-delphix:
…l snapshots
Current behavior: 
On Linux we have the "com.delphix:current-version" property set for datasets and snapshots under the "rpool/ROOT/delphix.#######" container. When we do in-place upgrade, we do not create new containers for each upgrade, but rather recursively snapshot the container and then apply the upgrade in-place. We then set the "com.delphix:current-version" property of the snapshot of the container to be the old version and then the same property for the container itself to be the new version. This way, we can find out which version of the product an older snapshot points to.
The problem is that we do not update this property for the snapshots for the container's children datasets.

Testing:
Attached chained upgrade results showing that the com.delphix:current-version is being set for recursive snapshots.
[chained_upgrade.txt](https://github.com/delphix/appliance-build/files/4347276/chained_upgrade.txt)
ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3064/

### DIFF
--- a/upgrade/upgrade-scripts/rootfs-container
+++ b/upgrade/upgrade-scripts/rootfs-container
@@ -31,16 +31,17 @@ function delete() {
 	local clonesnaps=()
 
 	#
-	# The "data" and "home" datasets of a rootfs container may have
-	# been cloned as part of a prior upgrade, and the "root" dataset
-	# may have been cloned as part of a prior rollback. Thus, in
-	# order to delete this specific rootfs container, we need to
-	# promote any clones that exist.
+	# The "data", "home", and "log" datasets of a rootfs container
+	# may have been cloned as part of a prior upgrade, and the
+	# "root" dataset may have been cloned as part of a prior
+	# rollback. Thus, in order to delete this specific rootfs
+	# container, we need to promote any clones that exist.
 	#
 	for snap in \
 		$(get_dataset_snapshots "rpool/ROOT/$CONTAINER/root") \
 		$(get_dataset_snapshots "rpool/ROOT/$CONTAINER/data") \
-		$(get_dataset_snapshots "rpool/ROOT/$CONTAINER/home"); do
+		$(get_dataset_snapshots "rpool/ROOT/$CONTAINER/home") \
+		$(get_dataset_snapshots "rpool/ROOT/$CONTAINER/log"); do
 		for clone in $(get_snapshot_clones "$snap"); do
 			zfs promote "$clone" ||
 				die "'zfs promote $clone' failed"


prakashsurya:
What is the motivation for fixing this? Just wondering if this bug is impacting some workflow, or not...

prakashsurya:
If I understand correctly, we're modifying the properly for the snapshots of the child datasets (i.e. snapshots of `log`, `data`, etc) such that the property will be "local" to that snapshot, and not inherited. The reason we _have_ to do that for snapshots, is otherwise the snapshot will inherit from the parent dataset, _not_ the corresponding snapshot of that parent dataset (which is what we want). Thus, since we _have_ to do this for the child dataset snapshots, I definitely think this is appropriate.

With that said, should we do this also for the child datasets themselves? I don't think we _have_ to do this for the datasets (not the snapshots of those datasets), because they'll automatically inherit the correct value; i.e. they'll inherit the value from their parent dataset, which is what we want.. but, I'm wondering if we should modify those datasets to keep a local version of the property, so that we have consistency between how the property is stored for the child dataset snapshots, as well as the child datasets themselves.

This way, no matter if one is looking at a child dataset, or a child dataset snapshot, the property will be stored locally; vs. having two modes, where the snapshots have a local property, and the datasets have an inherited property.

What do you think?

For example:
```
$ sudo zfs get -r com.delphix:current-version rpool/ROOT/delphix.jDHoOtU
NAME                                                       PROPERTY                     VALUE                        SOURCE
rpool/ROOT/delphix.jDHoOtU                                 com.delphix:current-version  6.0.1.0                      local
rpool/ROOT/delphix.jDHoOtU@container-delphix.Bs4VTPo       com.delphix:current-version  6.0.1.0                      local
rpool/ROOT/delphix.jDHoOtU/data                            com.delphix:current-version  6.0.1.0                      inherited from rpool/ROOT/delphix.jDHoOtU
rpool/ROOT/delphix.jDHoOtU/data@container-delphix.Bs4VTPo  com.delphix:current-version  6.0.1.0                      inherited from rpool/ROOT/delphix.jDHoOtU
rpool/ROOT/delphix.jDHoOtU/home                            com.delphix:current-version  6.0.1.0                      inherited from rpool/ROOT/delphix.jDHoOtU
rpool/ROOT/delphix.jDHoOtU/home@container-delphix.Bs4VTPo  com.delphix:current-version  6.0.1.0                      inherited from rpool/ROOT/delphix.jDHoOtU
rpool/ROOT/delphix.jDHoOtU/log                             com.delphix:current-version  6.0.1.0                      inherited from rpool/ROOT/delphix.jDHoOtU
rpool/ROOT/delphix.jDHoOtU/log@container-delphix.Bs4VTPo   com.delphix:current-version  6.0.1.0                      inherited from rpool/ROOT/delphix.jDHoOtU
rpool/ROOT/delphix.jDHoOtU/root                            com.delphix:current-version  6.0.1.0                      inherited from rpool/ROOT/delphix.jDHoOtU
rpool/ROOT/delphix.jDHoOtU/root@container-delphix.Bs4VTPo  com.delphix:current-version  6.0.1.0                      inherited from rpool/ROOT/delphix.jDHoOtU
```

This change will make it so all the snapshots have a local property (it's currently inherited, today):
```
rpool/ROOT/delphix.jDHoOtU/data@container-delphix.Bs4VTPo  com.delphix:current-version  6.0.1.0                      inherited from rpool/ROOT/delphix.jDHoOtU
rpool/ROOT/delphix.jDHoOtU/home@container-delphix.Bs4VTPo  com.delphix:current-version  6.0.1.0                      inherited from rpool/ROOT/delphix.jDHoOtU
rpool/ROOT/delphix.jDHoOtU/log@container-delphix.Bs4VTPo   com.delphix:current-version  6.0.1.0                      inherited from rpool/ROOT/delphix.jDHoOtU
rpool/ROOT/delphix.jDHoOtU/root@container-delphix.Bs4VTPo  com.delphix:current-version  6.0.1.0                      inherited from rpool/ROOT/delphix.jDHoOtU
```

And what I'm proposing is we also make it so the datasets themselves also have a local property (again, today, these are all inherited):
```
rpool/ROOT/delphix.jDHoOtU/data                            com.delphix:current-version  6.0.1.0                      inherited from rpool/ROOT/delphix.jDHoOtU
rpool/ROOT/delphix.jDHoOtU/home                            com.delphix:current-version  6.0.1.0                      inherited from rpool/ROOT/delphix.jDHoOtU
rpool/ROOT/delphix.jDHoOtU/log                             com.delphix:current-version  6.0.1.0                      inherited from rpool/ROOT/delphix.jDHoOtU
rpool/ROOT/delphix.jDHoOtU/root                            com.delphix:current-version  6.0.1.0                      inherited from rpool/ROOT/delphix.jDHoOtU
```

Would it be beneficial to always rely on local properties (e.g. perhaps less complexity), and ensure we properly set these properties for all snapshots and datasets? Or would it be better to rely on inheritance for datasets (since that works properly for our use-case with datasets), and local properties for snapshots (since inheritance does not work properly for our use-case with snapshots)? I believe you're proposing the latter with the changes you have in this PR so far.

sumedhbala-delphix:
> What is the motivation for fixing this? Just wondering if this bug is impacting some workflow, or not...

Just technical debt that I thought I can fix. Customers do not have visibility into this. It could potentially confuse support or engineering who are looking at these values.

sumedhbala-delphix:
> Would it be beneficial to always rely on local properties (e.g. perhaps less complexity), and ensure we properly set these properties for all snapshots and datasets? Or would it be better to rely on inheritance for datasets (since that works properly for our use-case with datasets), and local properties for snapshots (since inheritance does not work properly for our use-case with snapshots)? I believe you're proposing the latter with the changes you have in this PR so far.

I did not think about setting local properties on child datasets since inheritance gives the correct value. I could go be swayed wither way: setting the local value on child datasets by adding additional code OR using inheritance to return the correct value. I was going for the later since the code changes involved is lesser and it works.

If you think using inheritance is acceptable, I would prefer it since it reduces the number of code changes needed and it has already been tested.

prakashsurya:
> If you think using inheritance is acceptable, I would prefer it since it reduces the number of code changes needed and it has already been tested.

OK, sounds good. I'm good moving forward with this approach.

upgrade/upgrade-scripts/common.sh
```diff
@@ -193,6 +193,17 @@ function copy_dataset_property() {
 		die "failed to set property '$PROP_NAME=$PROP_VALUE' for '$DST_DATASET'"
 }
 
+function copy_dataset_property_to_snapshots_recursive() {
```
prakashsurya: I don't feel strongly, but I would prefer to inline this function into the callers. By doing so, we'd duplicate a small amount of code (a loop, and one call to `zfs list`), but I think that'll make it more clear to any readers/maintainers of these scripts exactly what is happening.

What do you think?

I'll approve as-is in case you'd rather stick with this function.

upgrade/upgrade-scripts/common.sh
```diff
@@ -193,6 +193,17 @@ function copy_dataset_property() {
...
+	local PROP_NAME="$1"
+	local SRC_DATASET="$2"
+	local SNAPSHOT_NAME="$3"
+
+	for dir in $(zfs list -rH -o name "$SRC_DATASET"); do
```
prakashsurya: I'd prefer `dataset` (or `ds` for short) to `dir`, since it's not really a "directory" (which is what I think of when I see "dir"); but it doesn't really matter so this is fine too.

PR #455, opened 2020-03-25
Use "delphix-upgrade-verification" package in "verify-jar" script

sanbhara:
DLPX-68919 Use New upgrade-verify.jar in appliance-build
DLPX-69031 Make sure that the delphix-verification package is not installed post upgrade
DLPX-64198 define and implement version info file for verification package

DLPX-68919: This PR adds in the new SUV based upgrade verify debian package as an input to the build-upgrade-image.sh script. This artifact is published to S3 from the https://gitlab.delphix.com/app/delphix-upgrade-verification repo and is intended to be packaged within the upgrade image for pre-upgrade verification.

DLPX-69031: This debian package is intended to be used only during pre-upgrade verification and hence is only installed for this purpose in verify-jar.sh and uninstalled once verification is completed.

DLPX-64198: The version of the upgrade verify artifact is independent of the delphix appliance version, hence it is stored in a new verification-version.info file within the upgrade image. This file is created in the build-upgrade-image script similar to the version.info file.

git ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3200/

### DIFF
--- a/scripts/build-upgrade-image.sh
+++ b/scripts/build-upgrade-image.sh
@@ -15,6 +15,8 @@
 # limitations under the License.
 #
 
+. "${BASH_SOURCE%/*}/common.sh"
+
 #
 # This script is intended to build an upgrade image that contains all of
 # the packages needed to upgrade a particular variant of the appliance,
@@ -25,8 +27,6 @@
 # combining all the deb.tar.gz tarballs produced by live build for this
 # variant (of which there will be one per supported platform).
 #
-
-TOP=$(git rev-parse --show-toplevel 2>/dev/null)
 
 if [[ -z "$TOP" ]]; then
 	echo "Must be run inside the git repsitory."
@@ -63,6 +63,20 @@ for deb_tarball in "$LIVE_BUILD_OUTPUT_DIR/$APPLIANCE_VARIANT"*.debs.tar.gz; do
 done
 
 #
+# Download the delphix upgrade verification debian package.
+# Note, we always pull from the "master" build of the verification
+# package, no matter what the UPSTREAM_BRANCH of the appliance is that
+# we're building; this is intentional.
+#
+
+AWS_S3_URI_UPGRADE_VERIFICATION=$(resolve_s3_uri \
+	"$AWS_S3_URI_UPGRADE_VERIFICATION" \
+	"$AWS_S3_PREFIX_UPGRADE_VERIFICATION" \
+	"devops-gate/master/upgrade-verify/master/post-push/release/latest")
+
+download_delphix_s3_debs "$TOP/upgrade/debs" "$AWS_S3_URI_UPGRADE_VERIFICATION"
+
+#
 # Generate an Aptly/APT repository
 #
 aptly repo create -distribution=bionic -component=delphix upgrade-repository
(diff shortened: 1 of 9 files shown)


matt-skinner-delphix:
Looks good other than the shell script formatter. Probably need to get @pzakha involved or some other person who can eventually merge it. 👍

sanbhara:
@vimalraj-anbarasu @sumedhbala-delphix @prakashsurya @matt-skinner-delphix @AdityaPrasad-dlpx

scripts/build-upgrade-image.sh
```diff
@@ -93,6 +106,12 @@ cp version.info.template version.info
 VERSION=$(dpkg -f "$(find debs/ -name 'delphix-entire-*' | head -n 1)" version)
 sed -i "s/@@VERSION@@/$VERSION/" version.info
 
+cp verification-version.info.template verification-version.info
```
prakashsurya: I think it'd be good to add a comment to explain why we're not using `version.info` for this info.

scripts/build-upgrade-image.sh
```diff
@@ -93,6 +106,12 @@ cp version.info.template version.info
...
+
+# Include version information about this image.
+VERIFICATION_VERSION=$(dpkg -f "$(find debs/ -name 'delphix-upgrade-verification*.deb' | head -n 1)" version)
+sed -i "s/@@VERIFICATION_VERSION@@/$VERIFICATION_VERSION/" verification-version.info
```
prakashsurya: Can we also put this information into `version.info` as well?

I understand that we have to use the `verification-version.info` file for existing release (i.e. 6.0.0 and 6.0.1), but I think we should adapt future releases to use the `version.info` file, and then we can eventually remove the `verification-version.info` file (once we no longer support upgrading from 6.0.0 and 6.0.1).

With that said, I don't feel strongly; IMO, we should have planned to use `version.info` from the start, but that's a tangent...

upgrade/upgrade-scripts/verify-jar
```diff
@@ -42,6 +44,11 @@ function verify_jar_verify_cleanup() {
 			die "failed to cleanup postgres for snapshot '${MDS_SNAPNAME}'."
 	fi
 
+	if [[ -d "$JAR_PATH" ]]; then
```
prakashsurya: This conditional is required in the event that we trigger the cleanup logic prior to actually installing the package, correct?

If so, I think it'd be better to check to see if the package is installed, rather than checking for that directory.. i.e.
```
if dpkg-query -W delphix-upgrade-verification &>/dev/null; then
    apt_get remove ...
fi
```

IMO, this way it's more clear that we only want to attempt to remove the package, if the package is currently installed.

After this change, we could inline the `JAR_PATH` variable since it'll only be used in one place within this script.

upgrade/upgrade-scripts/verify-jar
```diff
@@ -19,7 +19,9 @@
 . "${BASH_SOURCE%/*}/common.sh"
 
 IMAGE_VERSION=$(get_image_version)
+IMAGE_PATH=$(get_image_path)
```
prakashsurya: We should verify `IMAGE_PATH` is not empty, similar to how we verify `IMAGE_VERSION` is not empty below.

scripts/build-upgrade-image.sh
sanbhara: Added it to version.info file. But it should be noted that we do not expect to be able to remove the verification-version.info file till at least 6.3 - and maybe even later if we start loosening the restrictions on 2 major release upgrade as a maximum.

prakashsurya: Right, I understand.

This will also allow us to easily retrieve the verification version without unpacking the image. We may not need that functionality now, but it might be useful.. e.g. when unpacking, to determine if we're unpacking an already unpacked version, but with a new/different verification version..

upgrade/version.info.template
```diff
@@ -57,3 +57,9 @@ MINIMUM_VERSION=@@MINIMUM_VERSION@@
 # MINIMUM_VERSION above.
 #
 MINIMUM_REBOOT_OPTIONAL_VERSION=@@MINIMUM_REBOOT_OPTIONAL_VERSION@@
+
+#
+# The version of the upgrade verification package contained in the upgrade image.
+#
+VERIFICATION_VERSION=@@VERIFICATION_VERSION@@
+
```
prakashsurya: trailing newline..

PR #464, opened 2020-04-06
TOOL-9418 dcenter image should include tenacity and update nfs threads

grwilson:


grwilson:
ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3223/

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
```diff
@@ -87,4 +88,10 @@
     regexp: '^preserve_hostname: false'
     line: 'preserve_hostname: true'
 
+# Update number of NFS threads to 64
```
prakashsurya, line 91: it looks like we currently do this configuration for the product via some app-gate logic.. instead of duplicating that work here, should we instead move the product solution into delphix-platform? then, it'd apply to the dcenter variant as well.. and any future changes would automatically apply to the dcenter variant (rather than needing to duplicate the changes for the product and dcenter).

grwilson: I looked at that, but the appstack delivers 2 files and uses `bos_mgmt` to decide which one to install. So I didn't see an easy way to consolidate without a more extensive change to the appstack logic. I could deliver a default version in `delphix-platform` which has this change and just have the appstack override this with the file it wants to use. Thoughts?

prakashsurya: We could potentially move both files into "delphix-platform", and then also add a script to switch between the two files; similar to what the app-gate is doing. And then then app-gate code could use that script to make the config swap, and dcenter could use it too (assuming the default would not work for dcenter).

With that said, I don't feel strongly. I'm OK with the approach you have here.

prakashsurya, line 91: I'd suggest either removing the comment, or adding some commentary about why 64 is better than the default value of 8.

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
```diff
@@ -87,4 +88,10 @@
...
+- lineinfile:
+    path: /etc/default/nfs-kernel-server
+    regexp: '^RPCNFSDCOUNT=8'
```
prakashsurya, line 94: I'm OK with this, but just want to point out that this will not work if the default of `8` is ever changed. Perhaps we should change `regexp` to `^RPCNFSDCOUNT=`? This way, it's updated no matter what the default value is.

PR #467, opened 2020-06-02
copy static DNS configuration for not-in-place upgrades

sebroy:
http://reviews.delphix.com/r/58876/ changes how static DNS configuration is maintained. With this PR, we ensure that the static configuration file is persisted across not-in-place upgrades.

### DIFF
--- a/upgrade/upgrade-scripts/upgrade-container
+++ b/upgrade/upgrade-scripts/upgrade-container
@@ -639,6 +639,7 @@ function migrate_configuration() {
 		/etc/ssh/ssh_host_rsa_key
 		/etc/ssh/ssh_host_rsa_key.pub
 		/etc/sudoers.d/90-cloud-init-users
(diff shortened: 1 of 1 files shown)


upgrade/upgrade-scripts/upgrade-container
```diff
@@ -639,6 +639,7 @@ function migrate_configuration() {
 		/etc/ssh/ssh_host_rsa_key
 		/etc/ssh/ssh_host_rsa_key.pub
 		/etc/sudoers.d/90-cloud-init-users
+		/etc/systemd/resolved.conf.d/delphix-static.con
```
prakashsurya, line 642: I think this should be `.conf`, rather than `.con`

sebroy: Indeed, it should be. :-}

PR #472, opened 2020-07-01
Remove unnecessary files from upgrade payload

prakashsurya:
Currently, when we generate an upgrade image, we include the entire
contents of "\~/.aptly" in the image. It turns out, this causes us to
include two identical copies of each package, one under "\~/.aptly/pool"
and another under "\~/.aptly/public/pool".

For the upgrade image, we only need the contents of the "~/.aptly/public"
directory, so this change modifies the upgrade image generation logic,
as well as the upgrade image application logic, to take this into
account. As a result, this significantly reduces the size of the upgrade
image (roughly a 50% reduction in size).


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3720/)

PR #476, opened 2020-07-22
Automatically open "6.0/stage" backport PRs

prakashsurya:


prakashsurya:
I don't think this could leverage this as-is, since "6.0/stage" and "master" have some differences that would first need to be reconciled (e.g. migration images have been removed in master, and we wouldn't want to remove them from 6.0/stage).. but once we manually merge master with 6.0/stage, and ensure we do that properly w.r.t. the known differences, this PR may help us.

pzakha:
Looks interesting, I think we can experiment with this.

So from my understanding, this would essentially have us switch from cherry picks to merges, which would mean that we expect that master and 6.0/stage are going to be very close. If we have some bigger changes that happen on master but are not yet ready to be backported to 6.0/stage, how do you think this would affect the workflow?

Also, I'm assuming that with this change we are not planning to have backport tickets open in Jira, right?

prakashsurya:
> So from my understanding, this would essentially have us switch from cherry picks to merges

yea, so how this action works today, is it will open a PR between "master" and "6.0/stage".. and then the PR merge button can be used to land it.

> which would mean that we expect that master and 6.0/stage are going to be very close.

generally speaking, yes.. since this would continually open PRs with all commits from master, 6.0/stage will be kept in sync with master.

> If we have some bigger changes that happen on master but are not yet ready to be backported to 6.0/stage, how do you think this would affect the workflow?

As is, we'd have to either manually merge master with 6.0/stage and remember to leave out the set of commits we don't want to have on 6.0/stage (e.g. perhaps via git-reverts).. or we could let the auto-merge land, then fix 6.0/stage after the fact, via git-revert PRs..

I haven't thought too much about how to solve this situation, so I'm open to ideas. I also don't know if this is a problem yet; do you know of commits that have landed to master, that should not be on 6.0/stage? (ignoring the ones that are strictly master specific, like setting the value in the `branch.config` file)

There's other ideas I've had for automating backports, but I don't know if they'd be better or worse (and would require more work to implement).. For example, we could add labels to PRs, and then have an action that fired when a PR is landed, and do the backport based on what labels were on the original PR.. e.g. have a label to signal we should backport it.. or perhaps do backports by default, and then have a label signal we should not do the backport.

Since we're already using this action in other repositories (e.g. delphix/sdb), I thought I'd see if it could be used more widely (e.g. here, delphix-platform, etc).

> Also, I'm assuming that with this change we are not planning to have backport tickets open in Jira, right?

Correct. We'd land to master, and this would take care of merge master with 6.0/stage.. no JIRA involved (technically, JIRA isn't yet required to even land on master, so the whole process could be JIRA-less).

pzakha:
> do you know of commits that have landed to master, that should not be on 6.0/stage?

The one project I can think of is once we start work on Ubuntu 20.04. That said, there would likely be advantages to make those commits backwards-compatible.

But yeah, in general I do see more chances of appliance-build diverging between master and 6.0 compared to say sdb.

PR #479, opened 2020-07-23
Install recovery env

pcd1193182:
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3774/

We add a new role to install the recovery environment so that we can maintain a clearer dependency ordering between the recovery environment package and the app-stack; the recovery environment package currently relies on the app-stack, while the app-stack can function without the recovery environment.

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.recovery-environment/meta/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.recovery-environment/meta/main.yml
@@ -1,5 +1,5 @@
 #
-# Copyright 2018 Delphix
+# Copyright 2020 Delphix
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -15,14 +15,5 @@
 #
 
 ---
-- hosts: all
-  gather_facts: no
-  vars:
-    ansible_python_interpreter: /usr/bin/python3
-  roles:
-    - appliance-build.minimal-common
-    - appliance-build.minimal-internal
-    - appliance-build.masking-common
-    - appliance-build.qa-internal
-    - appliance-build.virtualization-common
-    - appliance-build.recovery-environment
+dependencies:
+  - role: appliance-build.virtualization-common
--- a/live-build/misc/ansible-roles/appliance-build.recovery-environment/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.recovery-environment/tasks/main.yml
@@ -1,5 +1,5 @@
 #
-# Copyright 2018 Delphix
+# Copyright 2020 Delphix
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -15,14 +15,7 @@
 #
 
 ---
-- hosts: all
-  gather_facts: no
-  vars:
-    ansible_python_interpreter: /usr/bin/python3
-  roles:
-    - appliance-build.minimal-common
-    - appliance-build.minimal-internal
-    - appliance-build.masking-common
-    - appliance-build.qa-internal
-    - appliance-build.virtualization-common
-    - appliance-build.recovery-environment
+- apt:
+    name: recovery-environment
+    state: present
+
--- a/live-build/variants/external-standard/ansible/playbook.yml
+++ b/live-build/variants/external-standard/ansible/playbook.yml
@@ -23,3 +23,4 @@
     - appliance-build.minimal-common
     - appliance-build.masking-common
     - appliance-build.virtualization-common
+    - appliance-build.recovery-environment
--- a/live-build/variants/internal-dev/ansible/playbook.yml
+++ b/live-build/variants/internal-dev/ansible/playbook.yml
@@ -48,3 +48,4 @@
     - appliance-build.virtualization-common
     - appliance-build.virtualization-development
     - appliance-build.zfsonlinux-development
+    - appliance-build.recovery-environment


prakashsurya:
> the recovery environment package currently relies on the app-stack

If some of my questions above can be answered by the fact that the recovery environment package required the virtualization package, then wouldn't it be more appropriate to install the recovery environment package within the existing `appliance-build.virtualization-common` role? Since that way, there's no way to install the recovery environment, without also installing the virtualization package.

Alternatively, we could have this new role, depend on the `appliance-build.virtualization-common` role; but I don't see any reason to go down that route, since it's slightly more complicated than just adding it to the existing virtualization role, without adding any additional benefit.

Also, out of curiosity, does the recovery environment package definition have a dependency on the virtualization package?

prakashsurya:
> then wouldn't it be more appropriate to install the recovery environment package within the existing appliance-build.virtualization-common role?

Further, if this is the case, then couldn't we add the recovery-environment as a dependency of the virtualization package? Similar to how we do it for lots of other packages?

Sorry for all the questions, I'm just trying to better understand the motivations directing us to go with this approach, when compared with many of the other possible options.

pzakha:
I can answer some of your questions Prakash since I've worked with Paul on this.

> If some of my questions above can be answered by the fact that the recovery environment package required the virtualization package, then wouldn't it be more appropriate to install the recovery environment package within the existing appliance-build.virtualization-common role? Since that way, there's no way to install the recovery environment, without also installing the virtualization package.

Yeah, the recovery environment depends on the app-stack, as it takes the nginx config from it.

> Alternatively, we could have this new role, depend on the appliance-build.virtualization-common role; but I don't see any reason to go down that route, since it's slightly more complicated than just adding it to the existing virtualization role, without adding any additional benefit.

Yeah the recovery-environment role should depend on the virtualization-common role.

I think having the `recovery-environment` role separate makes more sense conceptually, since virtualization doesn't depend on the recovery environment being there, and also it is easier for someone to see where we install the recovery environment when it has its own role. Bundling it together with virtualization-common is an option, but I'd say I'm more in favour of keeping them separate.

prakashsurya:
>  and also it is easier for someone to see where we install the recovery environment when it has its own role.

IMO, having a new task that installs the package is no better than using the existing task that installs the virtualization package. I might argue that it makes things more difficult and complicated to understand, since this adds another task/file that has to be audited to understand what's occurring.

> since virtualization doesn't depend on the recovery environment being there

Sure, but for all cases where we install the virtualization package, we also want to install the recovery environment package.

Having multiple packages being installed in a single role doesn't necessarily mean any one of those packages are dependent on another.. rather, it states that due to policy (and not technical dependency), when one is installed, we want our appliance to have the other(s) installed as well.. based on other roles, I think grouping makes sense for this case, and adding another role only makes things more complicated.

For example, it's now possible to install virtualization without installing the recovery environment, and we have to be careful to not make that mistake; would we ever want to install virtualization without the recovery environment?

There's still some unanswered questions:

> Further, if this is the case, then couldn't we add the recovery-environment as a dependency of the virtualization package? 

> Also, out of curiosity, does the recovery environment package definition have a dependency on the virtualization package?

> I see you've left out the dcenter variant, was that intentional? If so, why dont we want it installed?

Presumably, for that last question.. we _do_ want to have it installed on dcenter, but due to technical limitations (not having virtualization installed), we _can't_.. is that accurate?

If so, I think it'd be best if we could eventually remove these limitations from the recovery environment, such that it could be properly used with or without the virtualization package.. but that's a tangent w.r.t. this PR.

pzakha:
Unfortunately recovery-environment depends on delphix-virtualization, not the other way around. But I'm fine in moving the installation to the `recovery-environment` package to `appliance-build.virtualization-common` role if you think it's better.

prakashsurya:
> Unfortunately recovery-environment depends on delphix-virtualization, not the other way around.

Why is that unfortunate? Couldn't both depend on each other?

pzakha:
> Why is that unfortunate? Couldn't both depend on each other?

They could in theory, since recovery-environment's `postinst` script would run after files for both packages are installed, but I'm worrying about introducing circular package dependencies as that could potentially affect upgrade either now or in the future.

Since virtualization doesn't actually depend on the recovery environment to function it would also not make much sense to have the delphix-virtualization package depend on the recovery-environment package.

If we can avoid having this circular dependency from being created, I'd much rather avoid it.

prakashsurya:
> but I'm worrying about introducing circular package dependencies as that could potentially affect upgrade either now or in the future.

I don't have any such worry about circular dependencies; we should support that use case. If we don't today, I'd categorize that as a bug.

If you're making a policy decision that we don't want to support circular dependencies, I'd suggest we make that as a team. Further, even if we did commit to that policy decision for Delphix packages, I'm not sure we can guarantee that's upheld by external packages.

> If we can avoid having this circular dependency from being created, I'd much rather avoid it.

I don't share this broad aversion to circular dependencies.

> Since virtualization doesn't actually depend on the recovery environment to function it would also not make much sense to have the delphix-virtualization package depend on the recovery-environment package.

But "doesn't actually depend" isn't the criteria we've used for the virtualization package dependencies in the past. For example, we currently install "dbgsym" packages via virtualization package dependencies, even though the virtualization package "doesn't actually depend" on those packages.

If we want to change the criteria, such that virtualization doesn't have package dependencies for any packages that it "doesn't actually depend" on, I might be OK with that.. but we should be consistent, and perhaps retroactively move some of it's current dependencies into appliance-build like this PR is suggesting.

Currently, I believe the criteria for virtualization dependencies as been more aligned with the set of packages we want installed when the virtualization package is installed (of which, I believe the recovery environment fits), rather than the set of packages that the virtualization required to properly operate.

pcd1193182:
I have a general moral objection circular dependencies, but I don't think I have enough background or investment in this area to necessarily be able to provide a strong viewpoint; I'm happy to go with whatever people think would be best.

prakashsurya:
My prior approval still stands, so I'm not blocking this.

prakashsurya:
FWIW, I did some reading and found this about circular dependencies (from [here](https://www.debian.org/doc/debian-policy/ch-relationships.html)):

> If there is a circular dependency among packages being installed or removed, installation or removal order honoring the dependency order is impossible, requiring the dependency loop be broken at some point and the dependency requirements violated for at least one package. Packages involved in circular dependencies may not be able to rely on their dependencies being configured before they themselves are configured, depending on which side of the break of the circular dependency loop they happen to be on. If one of the packages in the loop has no postinst script, then the cycle will be broken at that package; this ensures that all postinst scripts are run with their dependencies properly configured if this is possible. Otherwise the breaking point is arbitrary. Packages should therefore avoid circular dependencies where possible, particularly if they have postinst scripts.

PR #481, opened 2020-07-31
Enable generation of "sparse" upgrade image

prakashsurya:
This change adds a new script that can be used to generate a "sparse"
upgrade image. For example, given two input upgrade images, we can
generate a sparse image like so:

    $ ./scripts/aptly-repo-from-image-diff.sh ../master/internal-dev.upgrade.tar ../6.0-release/internal-dev.upgrade.tar
    $ ./scripts/upgrade-image-from-aptly-repo.sh sparse-internal-dev
    $ ls -l build/artifacts/sparse-internal-dev.upgrade.tar
    -rw-r--r-- 1 delphix staff 5459210240 Jul 31 19:47 build/artifacts/sparse-internal-dev.upgrade.tar

Additionally, we can inspect this new artifact and verify the set of
packages it contains:

    $ tar -xf sparse-internal-dev.upgrade.tar
    $ tar -xf payload.tar.gz
    $ find pool -type f -name '*.deb' | wc -l
    199

And contrast this with the number of packages contained in both the
"master" and "6.0-release" images:

 - master:

       $ tar -xf internal-dev.upgrade.tar
       $ tar -xf payload.tar.gz
       $ find pool -type f -name '*.deb' | wc -l
       1042

 - 6.0/release:

       $ tar -xf internal-dev.upgrade.tar
       $ tar -xf payload.tar.gz
       $ find pool -type f -name '*.deb' | wc -l
       1080


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3807/)

prakashsurya:
> There's a bunch of typos in the original comments but that can be addressed separately.

If you want to point them out in this review, I'm happy to fix them in this PR.

scripts/upgrade-image-from-aptly-repo.sh
```diff
@@ -0,0 +1,183 @@
...
+
+trap cleanup EXIT
+
+[[ -n "$TOP" ]] || die "must be run inside of the git repository"
+[[ $# -gt 1 ]] && usage "too many arguments specified"
+[[ $# -lt 1 ]] && usage "too few arguments specified"
+
+#
+# One of the "inputs" to this script is an Aptly repository stored in
+# the "~/.aptly/public". Thus, we do minimal verification of this input,
```
pzakha, line 49: typo: the "~/.aptly/public" -> drop "the" or append "directory"

scripts/upgrade-image-from-aptly-repo.sh
```diff
@@ -0,0 +1,183 @@
...
+# to checksum and store in "upgrade image" tarball itself. Rather than
+# the "upgrade image" tarball containing all of the files for the upgrade
+# scripts and APT repository, it'll only contain the "payload.tar.gz"
+# file, and a few other miscellaneous metadata files.
+#
+# We need to limit the files contained in the 'upgrade image" tarball
+# because the signing service that we use to sign the SHA256SUMS file
+# below, has a relatively small limit on the size of the file that we
+# can sign with it. Thus, we need the contents of the SHA256SUMS file to
+# be realtively small (and that file's size grows linearly with the
```
pzakha, line 73: typo: realtively

scripts/upgrade-image-from-aptly-repo.sh
```diff
@@ -0,0 +1,183 @@
...
+cp "$TOP/upgrade/version.info.template" version.info ||
+	die "failed to copy version.info.template"
+
+VERSION=$(dpkg -f "$(find ~/.aptly/public -type f -name 'delphix-entire-*' | head -n 1)" version)
+[[ -n "$VERSION" ]] || die "failed to determine delphix-entire version"
+sed -i "s/@@VERSION@@/$VERSION/" version.info ||
+	die "failed to set VERSION in version.info file"
+
+#
+# On 6.0 versions, the virtualization application expects to find the
```
pzakha, line 90: typo: 6.0 -> 6.0+, or on version 6.0 and above

scripts/upgrade-image-from-aptly-repo.sh
```diff
@@ -0,0 +1,183 @@
...
+VERIFICATION_VERSION=$(dpkg -f "$(find ~/.aptly/public -type f -name 'delphix-upgrade-verification*.deb' | head -n 1)" version)
+[[ -n "$VERIFICATION_VERSION" ]] || die "failed to determine verification version"
+sed -i "s/@@VERIFICATION_VERSION@@/$VERIFICATION_VERSION/" verification-version.info ||
+	die "failed to set VERIFICATION_VERSION in verification-version.info file"
+sed -i "s/@@VERIFICATION_VERSION@@/$VERIFICATION_VERSION/" version.info ||
+	die "failed to set VERIFICATION_VERSION in version.info file"
+
+#
+# If we're running the build manually, and these variables are not
+# already set in the environment, we use an arbitrarily low falue to
```
pzakha, line 107: typo: falue

scripts/upgrade-image-from-aptly-repo.sh
```diff
@@ -0,0 +1,183 @@
...
+# allow upgrading to the produce upgrade image from any version.
```
pzakha, line 108: typo: produce -> produced

scripts/upgrade-image-from-aptly-repo.sh
prakashsurya: I don't think this applies to all versions above 6.0.. IIRC, this applies strictly to 6.0.x versions, and not 6.1.x versions.. Thus, I'm more inclined to leave as is, rather than potentially imply to the reader that this comment applies to all versions above (and including) 6.0. What do you think?

pzakha: Sure, we can leave it as is.

prakashsurya: I verified this comment only applies to 6.0.x versions as I thought; here's the related JIRA issue: https://jira.delphix.com/browse/DLPX-69165

pzakha: Cool, thanks for posting the link, it all makes sense.

rasantel:
Can you remind me - what is the benefit of a sparse image as opposed to a full image? In the example, the sparse image is still over 5 GB in size, so presumably the largest packages have changed. Is it because we expect that, sometimes, sparse images will be small? (e.g. for some hotfixes).

rasantel:
Can you remind me - what is the benefit of a sparse image as opposed to a full image? In the example, the sparse image is still over 5 GB in size, so presumably the largest packages have changed. Is it because we expect that, sometimes, sparse images will be small? (e.g. for some hotfixes).

prakashsurya:
> what is the benefit of a sparse image as opposed to a full image?

The sparse image potentially contains fewer packages than the full image, potentially allowing it to be smaller in size. The drawback being, it can only be applied to DE of a specific version, rather than a full image being able to be applied to a wide range of versions.

> In the example, the sparse image is still over 5 GB in size, so presumably the largest packages have changed.

I'm not sure, but I assume you're correct. From the PR description, we can see the full image for master has 1042 packages, and the full image for 6.0/stage contains 1080 packages, while the sparse image contains only 199 packages.

The 5GB in size of the sparse image can be attributed to the size of the 199 packages it contains.

> Is it because we expect that, sometimes, sparse images will be small? (e.g. for some hotfixes).

I can't say if they _will_ or _will not_ be small, but a sparse image _could_ be small, whereas a full image generally _could not_ be small.

Currently, a full upgrade image's size is proportional to the size of the root filesystem contents for any given DE version. Thus, if we wanted a full upgrade image to be small, we'd need to limit what we store on the root filesystem for a given DE version (which is not practically feasible).

In contrast, a sparse upgrade image's size is proportional to the number of packages that differ between two DE versions. Thus, if we want a sparse upgrade image to be small, we'd need to limit the number of packages that have been modified between two DE versions.

And as you allude to, my assumption is that for most hotfixes we'll only modify a small number of packages, therefore a sparse upgrade image with the hotfix would likely be much smaller in size than a full upgrade image with the hotfix; obviously the actual size of the sparse image still depends on the size of the package(s) being modified (e.g. the virtualization package is relatively large, so any hotfix that modifies that package will still have a relatively large sparse upgrade image, but still smaller than the full upgrade image).

rasantel:
Thanks @prakashsurya for your detailed answer.

PR #482, opened 2020-07-31
DLPX-69447 Fixed hotfixes should be removed from /etc/hotfix during upgrade application

sanbhara:
remove the hotfix file post-upgrade.

ab-pre-push : http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3831/

Testing : tested deferred upgrade of an engine with a non-empty hotfix file using upgrade image from above build. The hotfix file was removed post-upgrade.

### DIFF
--- a/upgrade/upgrade-scripts/upgrade
+++ b/upgrade/upgrade-scripts/upgrade
@@ -446,6 +446,10 @@ function finalize() {
 	rm -rf "$IMAGE_PATH" || die "failed to remove unpacked upgrade image"
 
 	remove_upgrade_properties
+
+	if [[ -f "$HOTFIX_PATH" ]]; then
+		rm -f "$HOTFIX_PATH" || die "failed to remove hotfix file"
+	fi
 }
 
 [[ "$EUID" -ne 0 ]] && die "must be run as root"
--- a/upgrade/upgrade-scripts/common.sh
+++ b/upgrade/upgrade-scripts/common.sh
@@ -18,6 +18,7 @@
 # shellcheck disable=SC2034
 UPDATE_DIR="/var/dlpx-update"
 LOG_DIRECTORY="/var/tmp/delphix-upgrade"
+HOTFIX_PATH="/etc/hotfix"
 
 #
 # We embed information as dataset properties in our rootfs containers.


sanbhara:
@vimalraj-anbarasu @sumedhbala-delphix

upgrade/upgrade-scripts/upgrade
```diff
@@ -446,6 +448,10 @@ function finalize() {
 	rm -rf "$IMAGE_PATH" || die "failed to remove unpacked upgrade image"
 
 	remove_upgrade_properties
+
+	if [[ -f "$HOTFIX_PATH" ]]; then
+		rm -f "$HOTFIX_PATH" || die "failed to remove hotfix file"
```
prakashsurya, line 453: Can you comment on why you chose to put this logic here? e.g. why not within the `execute` script?

For not-in-place upgrades, where we create a new root filesystem, do we still want to remove this file from the old/current root filesystem? I ask, because for that type of upgrade, I don't think we would have modified the old/current root filesystem.. and we may still use it, in the case of a rollback.. at that point, would we want the file to still be there, or to have been removed?

Additionally, if we end up using these script for application of a hotfix, do we still want to remove this file after the "upgrade"? E.g. I'm imagining at some point, we may deliver hotfixes as an "upgrade image", and apply the hotfix similar to how we perform deferred upgrades.. in such an architecture, would this logic still apply?

sanbhara: Ya I put it here specifically since it should be beyond the point of no return as far as rollbacks are concerned. I do not want to remove this file if there is still a possibility of a rollback happening. This of course means that I think the original hotfix file is still relevant post-rollback. Let me know if you think otherwise.

If and when we start using these scripts to perform hotfix install, I would imagine the same location to still be the right place to add in new logic to append the new hotfix data into the file - rather than deleting it as in the upgrade case. This is assuming the finalize() function is relevant for the hotfix install workflow. If not, we can find a different location for the hotfix specific logic.

prakashsurya: > Let me know if you think otherwise.

Ah yes, I see now.. OK, yea, I think this location should work then. Thanks for confirming.

> I would imagine the same location to still be the right place to add in new logic to append the new hotfix data into the file

Sounds good to me; thanks for thinking that case through.

upgrade/upgrade-scripts/upgrade
```diff
@@ -23,6 +23,8 @@ IMAGE_PATH=$(get_image_path)
 IMAGE_VERSION=$(get_image_version)
 [[ -n "$IMAGE_VERSION" ]] || die "failed to determine image version"
 
+HOTFIX_PATH="/etc/hotfix"
```
prakashsurya, line 26: If you want to use a variable for this path, rather than inline it below where it's used, I'd suggest we move it into `common.sh`, in case it's ever needed in another script (at some later time).. that'll be more consistent with other similar variables/paths too.. e.g. `UPDATE_DIR`, `LOG_DIRECTORY`, etc.

sanbhara: Sure

sanbhara: Done.

PR #493, opened 2020-10-29
DLPX-71683 Generate new buildinfo package using metadata from combined packages

pzakha:
The old `delphix-build-info` package was removed by DLPX-71608 following the linux-pkg rework. This adds a new file to the `delphix-entire` package with similar content.

## Testing
ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/4248/


pzakha:
Contents of the build-info file located at `/usr/share/doc/delphix-entire-aws/build-info.gz`:
```
----------------------------------------
APPLIANCE-BUILD ENV
----------------------------------------
APPLIANCE_BUILD_GIT_URL: https://gitlab.delphix.com/misc/staging-appliance-build.git
APPLIANCE_BUILD_GIT_BRANCH: refs/heads/os-platform/appliance-build/___master___/no-check/pzakharov/appliance-build-orchestrator-pre-push/master/56be2bd648e816c6fa9daf6106b8b262bfc9e936
DELPHIX_PACKAGE_MIRROR_MAIN: http://linux-package-mirror.delphix.com/master/e4f58b8d/ubuntu
DELPHIX_PACKAGE_MIRROR_SECONDARY: http://linux-package-mirror.delphix.com/master/e4f58b8d/ppas
AWS_S3_OUTPUT: s3://dev-de-images/builds/jenkins-selfservice/devops-gate/master/appliance-build/master/pre-push/5922
DELPHIX_APPLIANCE_VERSION: 6.1.0.0-snapshot.20201029163241503+jenkins-selfservice-devops-gate-master-appliance-build-master-pre-push-5922

----------------------------------------
COMPONENTS
----------------------------------------
adoptopenjdk: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/adoptopenjdk/post-push/2
bcc: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/bcc/post-push/1
bpftrace: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/bpftrace/post-push/10
challenge-response: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/challenge-response/post-push/1
cloud-init: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/cloud-init/post-push/1
crash: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/crash/post-push/3
crash-python: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/crash-python/post-push/2
crypt-blowfish: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/crypt-blowfish/post-push/1
delphix-platform: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/delphix-platform/post-push/4
delphix-sso-app: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/delphix-sso-app/post-push/3
drgn: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/drgn/post-push/5
gdb-python: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/gdb-python/post-push/1
grub2: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/grub2/post-push/3
libkdumpfile: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/libkdumpfile/post-push/3
make-jpkg: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/make-jpkg/post-push/1
makedumpfile: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/makedumpfile/post-push/2
nfs-utils: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/nfs-utils/post-push/2
performance-diagnostics: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/performance-diagnostics/post-push/3
ptools: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/ptools/post-push/1
python-rtslib-fb: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/python-rtslib-fb/post-push/2
recovery-environment: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/recovery-environment/post-push/3
savedump: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/savedump/post-push/1
sdb: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/sdb/post-push/2
targetcli-fb: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/targetcli-fb/post-push/2
td-agent-prebuilt: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/td-agent-prebuilt/post-push/1
virtualization: s3://snapshot-de-images/builds/jenkins-ops/dlpx-app-gate/master/build-package/post-push/3902
masking: s3://snapshot-de-images/builds/jenkins-ops/dms-core-gate/master/build-package/post-push/1159
linux-kernel-generic: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/linux-kernel-generic/post-push/2
linux-kernel-aws: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/linux-kernel-aws/post-push/2
linux-kernel-gcp: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/linux-kernel-gcp/post-push/2
linux-kernel-azure: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/linux-kernel-azure/post-push/2
linux-kernel-oracle: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/linux-kernel-oracle/post-push/2
connstat: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/connstat/post-push/2
delphix-kernel: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/delphix-kernel/post-push/2
zfs: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/zfs/post-push/16

----------------------------------------
KERNEL VERSIONS
----------------------------------------
linux-kernel-generic: 5.4.0-48-generic
linux-kernel-aws: 5.4.0-1025-aws
linux-kernel-gcp: 5.4.0-1025-gcp
linux-kernel-azure: 5.4.0-1026-azure
linux-kernel-oracle: 5.4.0-1025-oracle

----------------------------------------
PACKAGES
----------------------------------------
Package adoptopenjdk

Tar file: OpenJDK8U-jdk_x64_linux_hotspot_8u262b10.tar.gz
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package bcc

Git hash: 8f4cda9c873c3afc30ec8a74588a681ea6e0fc98
Git repo: https://github.com/delphix/bcc.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package bpftrace

Git hash: aa49bca95bb12d0a019189d847142ef41306bfb9
Git repo: https://github.com/delphix/bpftrace.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/f30e87dc/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/f30e87dc/ppas

----------------------------------------
Package challenge-response

Git hash: 79cba9d4d1adad4336945ba80ba2d1cc01d46a1e
Git repo: https://gitlab.delphix.com/os-platform/linux-dlpx-pkgs.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package cloud-init

Git hash: 1b8f9934f61920f239b5b36c128a35f053bb72cd
Git repo: https://github.com/delphix/cloud-init.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package connstat

Git hash: ec14029659d35381d52174da5151dfef275ae3fe
Git repo: https://github.com/delphix/connstat.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package crash-python

Git hash: 433443759f22ae9473a9db6535cf50a9ba428019
Git repo: https://gitlab.delphix.com/os-platform/crash-python.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package crash

Git hash: 757fd980a84235f84e83849e78b779704e5c5ea7
Git repo: https://github.com/delphix/crash.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/2fb719e8/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/2fb719e8/ppas

----------------------------------------
Package crypt-blowfish

Git hash: 7309aa54d7725300448229a0eee038917b2ff0b2
Git repo: https://gitlab.delphix.com/os-platform/linux-3rd-party-pkgs.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package delphix-kernel

Git hash: e801e08457d00d12cf49f7497ee1990a31e6a591
Git repo: https://github.com/delphix/delphix-kernel.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package delphix-platform

Git hash: 75fc688fcc8dad04343ca5691158bdc0142b5a86
Git repo: https://github.com/delphix/delphix-platform.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/f30e87dc/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/f30e87dc/ppas

----------------------------------------
Package delphix-sso-app

Git hash: 8a166848fb7d27f0c59c83cdd919aa26e42f824d
Git repo: https://gitlab.delphix.com/app/saml-app.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package drgn

Git hash: 22cf497068c8617c699d0c8cf8b5066d16657ccd
Git repo: https://github.com/delphix/drgn.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/1e45ccd9/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/1e45ccd9/ppas

----------------------------------------
Package gdb-python

Git hash: c186d2f17a47b1d487f8c015c7582cd8af8f64c5
Git repo: https://gitlab.delphix.com/os-platform/gdb-python.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package grub2

Git hash: e0e682c77bf3d42d24cd33f909b948c06fe48d86
Git repo: https://github.com/delphix/grub2
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package libkdumpfile

Git hash: ba084c28fca10fae577404b562ff282ddb0a6100
Git repo: https://github.com/delphix/libkdumpfile.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/1e45ccd9/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/1e45ccd9/ppas

----------------------------------------
Package linux-kernel-aws

Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package linux-kernel-azure

Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package linux-kernel-gcp

Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package linux-kernel-generic

Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package linux-kernel-oracle

Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package make-jpkg

Git hash: 2bf1564c65b231e0e34177545d12244e0373f6ce
Git repo: https://github.com/delphix/make-jpkg.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package makedumpfile

Git hash: b88f556f43790617f22060c10ace9e7e39c3e941
Git repo: https://github.com/delphix/makedumpfile.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package masking


----------------------------------------
Package nfs-utils

Git hash: 2a9a5e763e988b50b969e5872b297459a31ebb64
Git repo: https://github.com/delphix/nfs-utils.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package performance-diagnostics

Git hash: 7451798b2534b70b65ea5af9dd8b9d142094d416
Git repo: https://github.com/delphix/performance-diagnostics.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/2fb719e8/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/2fb719e8/ppas

----------------------------------------
Package ptools

Git hash: 81c2e03707bcf5a4c06a146721225c95460c0f9c
Git repo: https://github.com/delphix/ptools.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package python-rtslib-fb

Git hash: e81c942ede519e1d3e36c4244bce5e26dbc444aa
Git repo: https://github.com/delphix/python-rtslib-fb.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas
----------------------------------------
Package recovery-environment

Git hash: 8e10508255dfb92d52ae766748be0b92e0b39e22
Git repo: https://github.com/delphix/recovery-environment.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package savedump

Git hash: 47c0984ec2b0a0b48d877ab8c6c714d4ead41293
Git repo: https://github.com/delphix/savedump.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package sdb

Git hash: f0aa360dfcb9d1571a1d2c6950b844702a2dc844
Git repo: https://github.com/delphix/sdb.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/1e45ccd9/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/1e45ccd9/ppas

----------------------------------------
Package targetcli-fb

Git hash: a78282a9623175ee199552f1f50a9d98a0c22752
Git repo: https://github.com/delphix/targetcli-fb.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package td-agent-prebuilt

Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package virtualization


----------------------------------------
Package zfs

Git hash: 6f73b647b9160d29de4aeff02791520bd8de5de1
Git repo: https://github.com/delphix/zfs.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/f30e87dc/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/f30e87dc/ppas

----------------------------------------
````

**Review by sebroy — APPROVED** [review]

File: scripts/build-ancillary-repository.sh [hunk 1]
```diff
@@ -61,6 +62,48 @@ function build_ancillary_repository() {
...
+	echo "APPLIANCE_BUILD_GIT_BRANCH: $APPLIANCE_BUILD_GIT_BRANCH"
+	echo "DELPHIX_PACKAGE_MIRROR_MAIN: $DELPHIX_PACKAGE_MIRROR_MAIN"
+	echo "DELPHIX_PACKAGE_MIRROR_SECONDARY: $DELPHIX_PACKAGE_MIRROR_SECONDARY"
+	echo "AWS_S3_OUTPUT: $AWS_S3_OUTPUT"
+	echo "DELPHIX_APPLIANCE_VERSION: $DELPHIX_APPLIANCE_VERSION"
+	echo
+	echo "----------------------------------------"
+	echo "COMPONENTS"
+	echo "----------------------------------------"
+	[[ -f "$metadir/COMPONENTS" ]] && cat "$metadir/COMPONENTS"
```
prakashsurya, line 81: I really dislike how we're so tightly coupling appliance-build to the linux-pkg repository; and this coupling has been only increasing over time.

The original design was for appliance-build to consume arbitrary debian packages and work properly and equally for all consumed debian packages.. but now we seem to be moving to a design where linux-pkg-based packages are handled differently, and perhaps to a design where appliance-build only works with linux-pkg-based packages, rather than arbitrary debian packages.

IMO, appliance-build shouldn't care about the build system of any of our debian packages (e.g. appliance-build shouldn't know that packages were built via linux-pkg or not), but I guess we no longer aspire to such a design...

scripts/build-ancillary-repository.sh
```diff
@@ -61,6 +62,48 @@ function build_ancillary_repository() {
 	EOF
 }
 
+function compile_build_info() {
```
prakashsurya, line 65: How much of this build information is related to the ancillary repository? It seems awkward to have information unnecessary to the ancillary repository being emitted by this script.

scripts/build-ancillary-repository.sh
```diff
@@ -119,23 +162,45 @@ AWS_S3_URI_COMBINED_PACKAGES=$(resolve_s3_uri \
...
+# combined-packages S3 location.
+#
+aws s3 sync --only-show-errors "$AWS_S3_URI_COMBINED_PACKAGES" "$WORK_DIRECTORY"
+
+#
+# Prepare a build-info file while we have access to the downloaded build
+# metadata files. Later in the build this file will be included in the
+# delphix-entire package.
+#
+compile_build_info "$WORK_DIRECTORY" >"$BUILD_INFO_FILE"
```
prakashsurya, line 187: Why are we generating this file at this point in the build?

At first glance, this doesn't really seem directly related to building the ancillary repository, so it just seems a little awkward that you're choosing to generate the file here.

Further, this file is only consumed by the delphix-entire package, so if we just look at this script in isolation, it seems super awkward to generate it and not use it for anything.

For example, if we want to run this script manually (or any other context other than how we currently call it) to build that ancillary repository, would we also need/want to generate this file? In such a case, this seems like it may be an undesirable side-effect.

I'm not sure how best to implement this, but I feel like a better solution would be to have a separate script, whose sole responsibility was to generate this build information. And then have an explicit place in the build, e.g. when we build delphix-entire, where the script was called to generate the file.

Along the same lines, I think it'd be nice to generate this file closer to the logic that would consume the file. As is, we generate the file, and then have to jump through some non-intuitive hoops to ensure this file is properly propagated through the build to the point where it's consumed (i.e. the change to `run-live-build.sh`).

prakashsurya (commented):
I can't approve this, but I wont block this either (if others wish to approve).

I'm not sure that I agree that appliance-build should be the mechanism that's used to capture this build information about individual packages, and inject it into delphix-entire. To me, this is leaking implementation details of linux-pkg into appliance-build, when appliance-build should not have such intimate knowledge.

For example, if we want to have an audit trail of the git repository information that was used to build a package, IMO, the package should take care of including that information within its own debian package. The solution being proposed here just feels overly complex and brittle.. i.e. relying on linux-pkg to build the package, then linux-pkg to combine packages, then storing this `PACKAGES` file, and then finally consuming this `PACKAGES` file when building the appliance.

Please let me know if other (perhaps more simple) solutions were considered and discounted for some reason, but as is, this feels like a Rube Goldberg machine type of a solution.

I believe a better solution would be to rely on each individual package to injecting information about itself into it's own package, then we simply pass around the debian package(s) and not have to worry about this additional metadata (e.g. the `PACKAGES` file). In such a solution, appliance-build would not have to concern itself about building and/or storing this package-specific build information.

prakashsurya:
Why don't we have build information about the masking and virtualization packages?
```
----------------------------------------
Package masking


----------------------------------------
```
```
----------------------------------------
Package virtualization


----------------------------------------
```

Should we have build information here about the upgrade-verification package? I don't see that package in the list. That package won't be installed on an engine, but it will be contained in an upgrade image; how/where would we store that information? We should include information about that package, right? I mean, it'll be used during upgrade, so I feel like having this same build information for that package is just as important as any other package.

pzakha:
Thanks for the thorough review Prakash.

When designing this change I went for the way of simplicity: how to get the package build info into the appliance in the simplest way.

I had considered an alternative way: have the `combine-packages` Jenkins job generate the build-info package, however it had a few caveats:
1. The combine-packages job is ran after any package is built, so it was meant to be a lightweight job that runs fairly quickly.
2. Generating a new metapackage package requires installing some packages, which we cannot do directly given that combine-packages doesn't clone a new VM for its execution and runs on a "misc" Jenkins agent .

I had not considered the other alternative of injecting the build-info into the packages being built, but thinking of it now it seems like it would be much more invasive and complex than the other alternatives. I feel like it's also convenient to have all the build-info data in the same place.

Regarding the coupling of appliance-build with linux-pkg, I could argue that it's quite fairly loosely coupled. I do agree that having too much coupling between multiple components of a system makes the system more brittle, as changing one component of the system can then unexpectedly break another component of the system. What's important is to have some clear interface points between the two components (similarly to how we have APIs). In this case the only links we have between appliance-build and linux-pkg is the structure of the S3 URL being provided and the assumption that some specific files will be present in there. Linux-pkg's logic to generate the build-info has been made generic and warnings have been added that the various pieces of the build-info are used in other parts of the build so they should not be changed carelessly.

Regarding the virtualization and masking packages metadata, I didn't add it because it didn't have the same files, but I could actually add the `metadata.json` files for now. I haven't looked much into the upgrade-verification package, but I could take a look at it.

scripts/build-ancillary-repository.sh
pzakha: The main reason why I'm generating this here is that it's the only time that we have access to downloaded packages.

I agree that it is a bit awkward to put this into build-ancillary-repository. I can look into creating a new function that would download the metadata and generate this info when generating the delphix-entire package (or even create a separate package, although I don't feel like that's needed).

prakashsurya:
> What's important is to have some clear interface points between the two components

I agree; we differ on what the interface should be.

> In this case the only links we have between appliance-build and linux-pkg is the structure of the S3 URL being provided and the assumption that some specific files will be present in there.

Right, and my disagreement is in the fact that appliance-build needs to know about linux-pkg specific details. For example, for packages that aren't built via linux-pkg, these details are not required. I view linux-pkg as a build implementation detail for certain packages that use it, whereas I think you view it as a critical part of our appliance build architecture as a whole (hence details of that build system leaking into other parts of product build architecture).

IMO, packages built by linux-pkg and packages not built by linux-pkg should be treated by appliance-build the same, which is not the case today. I'd argue even packages not built by Delphix (e.g. packages consumed from the Ubuntu mirrors) should be treated by appliance-build the same as packages built by Delphix.

My original intention was for the interface of appliance-build to consume debian packages directly (i.e. more precisely, APT repos and S3 links to packages, since we decided to use S3 rather than APT for our internal packages); but you're moving us to a world where we're adding a new interface is for appliance-build, which is the output of linux-pkg (i.e. linux-pkg specific `COMPONENTS` and `PACKAGES` files store in S3), adding a new linux-pkg specific abstraction into the mix.

It's this linux-pkg specific interface that I don't necessarily agree with; in contrast to the non-linux-pkg specific interface of debian packages.

> Regarding the virtualization and masking packages metadata, I didn't add it because it didn't have the same files, but I could actually add the metadata.json files for now. I haven't looked much into the upgrade-verification package, but I could take a look at it.

Part of my concern, is that we now have to have different logic for different packages, depending on the build system of each package. IMO, appliance-build should be treating any individual package differently.

For example, we shouldn't have to know that some packages were generating via linux-pkg, so we can get there build information the linux-pkg specific way. Other packages were built via Jenkins, and we have to get their build information via some `metadata.json` file. And then, further still, some other packages will not be contained in the VM images, so we have to get their build information a third way.

IMO, that's leaking too much implementation details about each individual package into the build system of the appliance; to me, by design, the appliance build system should not care at all about how each back is generated and/or where it was generated from. Hence my hesitation to move forward with this approach.

So far, I'm still hesitant to approve this design, but I'm not blocking this from moving forward if others approve of this approach.

pzakha:
Yeah, I understand your desire to have appliance-build know less about the way the packages are built and the specifics of the Delphix Appliance. In the beginning your design of appliance-build was meant to be generic, where it could be used to create any kind of appliance. Over time as we have added more and more logic to appliance-build that made it less generic and more Delphix-centric. From my perspective this is somewhat normal as the only consumer of appliance-build is the Delphix Appliance (+ DCoL, which to me is a just a variation of the Delphix Appliance) and creating something with a higher level of abstraction has its own cost.

> IMO, that's leaking too much implementation details about each individual package into the build system of the appliance; to me, by design, the appliance build system should not care at all about how each package is generated and/or where it was generated from. Hence my hesitation to move forward with this approach.

If we really wanted it, we could remove the need for appliance-build from care about the package implementation details all-together. For instance, we could have a separate project/repository/job that is aware of where our packages are stored, that takes care of downloading the packages, potentially generating the necessary metadata packages, and then creating the ancillary package repository that it would store somewhere. Then we'd pass a single link to this repository to appliance-build and it would be able to just consume it without any regards to what it contains. With all that said, this would potentially require a lot of engineering work, and I personally don't see this as a very high priority right now, although this may be something we could consider in the future if we see that the current approach is causing too much friction.

The point of the build-info package is to add convenient debug information to the Delphix Appliance and it is not a critical part of the product. If we see that it causes too much friction we could always remove/refactor it, the same way as I did it in DLPX-71608.

> For example, we shouldn't have to know that some packages were generating via linux-pkg, so we can get their build information the linux-pkg specific way. Other packages were built via Jenkins, and we have to get their build information via some metadata.json file. And then, further still, some other packages will not be contained in the VM images, so we have to get their build information a third way.

I'm planning to do this on a "best-effort" basis i.e. if a given file exists in the package's artifacts (e.g. BUILD_INFO or metadata.json) then include it in the package metadata, if not then skip it. Which means that if we happened to remove that information, it would not break the build. Given that the build-info is a convenience feature, any temporary regressions with it (e.g. not containing build-info for some packages) would be acceptable.

prakashsurya:
> Given that the build-info is a convenience feature, any temporary regressions with it (e.g. not containing build-info for some packages) would be acceptable.

IMO, if it's not required, then I'd prefer not to include it.

If it is required, then I'd prefer to try and ensure we don't introduce regressions to it; and would disagree with the notion that's its acceptable to silently not contain this information for some packages.

> For instance, we could have a separate project/repository/job that is aware of where our packages are stored, that takes care of downloading the packages, potentially generating the necessary metadata packages, and then creating the ancillary package repository that it would store somewhere.

While we _could_ do this, I'm definitely not advocating for such an approach.

Currently, this is handled via appliance-build and works just fine; I don't see a good reason to refactor that part out of appliance-build.

We've loosely (previously) agreed that the input interface to appliace-build is a S3 bucket full of `.deb` files, and I'm fine with that interface, since it's relatively trivial to generate such an output artifact and is generic enough that it should be applicable to lots of (all?) different package build systems.

My concern is that we've since added a new input interface to appliance-build, that's specifically catered to the linux-pkg build system, rather than having linux-pkg cater to the existing input interface of appliance-build.

I'm not sure we'll come to agreement here, but like I said, I'm not blocking this; I just can't give approval if I have some more fundamental issues with the overall design.

pzakha:
I've rebased this on top of our latest code. buildinfo is now generated for all Delphix-built packages, including `upgrade-verify`. Here's the additional info:
```
----------------------------------------
Package masking

{"dms-core-gate": {
    "git-hash": "d6caf38e342e61d20da66c2cb36586840d68c561",
    "date": "2020-12-17T20:32:40+0000",
    "version": "trunk"
}}
----------------------------------------
Package upgrade-verify

{"delphix-upgrade-verification": {
    "git-hash": "004a0cdfd3d813ac0b8acec7f357c86d429435b9",
    "date": "2020-12-09T03:57:12+0000",
    "version": "1.0.5"
}}

----------------------------------------
Package virtualization

{"dlpx-app-gate": {
    "git-hash": "7c6d5f120d2ff7f554f2ea1c588470e23e290969",
    "date": "2020-12-18T03:07:09+0000",
    "version": "trunk"
}}
```

prakashsurya (approved):
I still have reservations about the overall design, and how our various parts of the build process fit together, but I'll approve to unblock this.

pzakha:
Rebasing and retesting:
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/5231/

```$ zcat /usr/share/doc/delphix-entire-*/build-info.gz
----------------------------------------
APPLIANCE-BUILD ENV
----------------------------------------
APPLIANCE_BUILD_GIT_URL: https://gitlab.delphix.com/misc/staging-appliance-build.git
APPLIANCE_BUILD_GIT_BRANCH: refs/heads/os-platform/appliance-build/___master___/no-check/pzakharov/appliance-build-orchestrator-pre-push/dlpx-71683/35207ccc5b62f3d860f057d7a8f3bb076db52060
appliance-build git hash: 35207ccc5b62f3d860f057d7a8f3bb076db52060
DELPHIX_PACKAGE_MIRROR_MAIN: http://linux-package-mirror.delphix.com/master/474e89c7/ubuntu
DELPHIX_PACKAGE_MIRROR_SECONDARY: http://linux-package-mirror.delphix.com/master/474e89c7/ppas
AWS_S3_OUTPUT: s3://dev-de-images/builds/jenkins-selfservice/devops-gate/master/appliance-build/master/pre-push/6460
DELPHIX_APPLIANCE_VERSION: 6.1.0.0-snapshot.20210507142844190+jenkins-selfservice-devops-gate-master-appliance-build-master-pre-push-6460

----------------------------------------
COMPONENTS
----------------------------------------
linux-kernel-generic: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/linux-kernel-generic/post-push/16
linux-kernel-aws: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/linux-kernel-aws/post-push/16
linux-kernel-gcp: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/linux-kernel-gcp/post-push/15
linux-kernel-azure: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/linux-kernel-azure/post-push/15
linux-kernel-oracle: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/linux-kernel-oracle/post-push/15
connstat: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/connstat/post-push/19
delphix-kernel: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/delphix-kernel/post-push/19
zfs: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/zfs/post-push/188
adoptopenjdk: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/adoptopenjdk/post-push/2
bcc: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/bcc/post-push/3
bpftrace: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/bpftrace/post-push/70
challenge-response: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/challenge-response/post-push/3
cloud-init: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/cloud-init/post-push/10
crash-python: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/crash-python/post-push/2
crypt-blowfish: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/crypt-blowfish/post-push/1
delphix-platform: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/delphix-platform/post-push/21
delphix-sso-app: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/delphix-sso-app/post-push/19
drgn: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/drgn/post-push/46
docker-python-image: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/docker-python-image/post-push/2
gdb-python: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/gdb-python/post-push/1
grub2: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/grub2/post-push/10
libkdumpfile: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/libkdumpfile/post-push/8
make-jpkg: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/make-jpkg/post-push/1
makedumpfile: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/makedumpfile/post-push/2
masking: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/masking/post-push/147
misc-debs: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/misc-debs/post-push/1
nfs-utils: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/nfs-utils/post-push/2
performance-diagnostics: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/performance-diagnostics/post-push/7
ptools: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/ptools/post-push/4
python-rtslib-fb: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/python-rtslib-fb/post-push/2
recovery-environment: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/recovery-environment/post-push/8
savedump: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/savedump/post-push/1
sdb: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/sdb/post-push/13
targetcli-fb: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/targetcli-fb/post-push/2
virtualization: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/build-package/virtualization/post-push/343
upgrade-verify: s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/upgrade-verify/master/post-push/30

----------------------------------------
KERNEL VERSIONS
----------------------------------------
linux-kernel-generic: 5.4.0-72-dx2021042100-57347bd21-generic
linux-kernel-aws: 5.4.0-1047-dx2021050705-b7b22b94c-aws
linux-kernel-gcp: 5.4.0-1042-dx2021042100-981ec6815-gcp
linux-kernel-azure: 5.4.0-1046-dx2021042100-b65b06645-azure
linux-kernel-oracle: 5.4.0-1043-dx2021042100-a69b530f0-oracle

----------------------------------------
PACKAGES
----------------------------------------
Package adoptopenjdk

Tar file: OpenJDK8U-jdk_x64_linux_hotspot_8u262b10.tar.gz
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package bcc

Git hash: d37492af897d788e5439cb640f70722c03070bde
Git repo: https://github.com/delphix/bcc.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/5f7c1cb9/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/5f7c1cb9/ppas

----------------------------------------
Package bpftrace

Git hash: 5c01e35c47e5a3c37e37f6fbeac91c6119c53bd9
Git repo: https://github.com/delphix/bpftrace.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/490756d0/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/490756d0/ppas

----------------------------------------
Package challenge-response

Git hash: 4ae0e2d62c22ef4e19ee4edc24a6a80085556624
Git repo: https://gitlab.delphix.com/os-platform/linux-dlpx-pkgs.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/7cd157c6/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/7cd157c6/ppas

----------------------------------------
Package cloud-init

Git hash: 1c8ffe7eb4098b5d8b9c8fefe37a22358eafa406
Git repo: https://github.com/delphix/cloud-init.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/ba99df38/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/ba99df38/ppas

----------------------------------------
Package connstat

Git hash: 550b2d70c35c8704a8b688909a6d97d0a5d3c4ed
Git repo: https://github.com/delphix/connstat.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/f0b1a8af/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/f0b1a8af/ppas

----------------------------------------
Package crash-python

Git hash: 433443759f22ae9473a9db6535cf50a9ba428019
Git repo: https://gitlab.delphix.com/os-platform/crash-python.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package crypt-blowfish

Git hash: 7309aa54d7725300448229a0eee038917b2ff0b2
Git repo: https://gitlab.delphix.com/os-platform/linux-3rd-party-pkgs.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package delphix-kernel

Git hash: a20482975ece0125b6d6a7d46d744952a5e7857e
Git repo: https://github.com/delphix/delphix-kernel.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/f0b1a8af/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/f0b1a8af/ppas

----------------------------------------
Package delphix-platform

Git hash: b52e8f35eb27fe7d2ef693a805020b37dd5ce4c2
Git repo: https://github.com/delphix/delphix-platform.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/08d66092/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/08d66092/ppas

----------------------------------------
Package delphix-sso-app

Git hash: 6f25985dba4c1bab6a68e7c9415d85c42be5b6e2
Git repo: https://gitlab.delphix.com/app/saml-app.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/ca31e19c/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/ca31e19c/ppas

----------------------------------------
Package docker-python-image

Git hash: 3fbec4795c5577f6676feb354edead563feadbb0
Git repo: https://github.com/delphix/docker-python-image.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/5f7c1cb9/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/5f7c1cb9/ppas

----------------------------------------
Package drgn

Git hash: 63d30c17167730f0ba48565e68e30ab5685ed059
Git repo: https://github.com/delphix/drgn.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/ba99df38/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/ba99df38/ppas

----------------------------------------
Package gdb-python

Git hash: c186d2f17a47b1d487f8c015c7582cd8af8f64c5
Git repo: https://gitlab.delphix.com/os-platform/gdb-python.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package grub2

Git hash: 5e5e1e26a9968e4c2cd7d9d6cdc0091188473ab8
Git repo: https://github.com/delphix/grub2
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/a7685b97/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/a7685b97/ppas

----------------------------------------
Package libkdumpfile

Git hash: 2ecf7136a9dcd0a1e8d807ed627b6e0d40de8a01
Git repo: https://github.com/delphix/libkdumpfile.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/2336f99f/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/2336f99f/ppas

----------------------------------------
Package linux-kernel-aws

Git hash: b7b22b94c161e81d2536a011299f5b314e3d6bc3
Git repo: https://github.com/delphix/linux-kernel-aws.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/f0b1a8af/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/f0b1a8af/ppas

----------------------------------------
Package linux-kernel-azure

Git hash: b65b06645e3acae868c1bff4b3ca471181b558e3
Git repo: https://github.com/delphix/linux-kernel-azure.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/ca31e19c/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/ca31e19c/ppas

----------------------------------------
Package linux-kernel-gcp

Git hash: 981ec68153d0b4c9e1a19cd463ec5475336c31f8
Git repo: https://github.com/delphix/linux-kernel-gcp.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/ca31e19c/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/ca31e19c/ppas

----------------------------------------
Package linux-kernel-generic

Git hash: 57347bd2129084045061a4851d0a3647cd223bb8
Git repo: https://github.com/delphix/linux-kernel-generic.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/ca31e19c/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/ca31e19c/ppas

----------------------------------------
Package linux-kernel-oracle

Git hash: a69b530f0a67f3f40f07d6fb9cf3710a8b595848
Git repo: https://github.com/delphix/linux-kernel-oracle.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/ca31e19c/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/ca31e19c/ppas

----------------------------------------
Package make-jpkg

Git hash: 2bf1564c65b231e0e34177545d12244e0373f6ce
Git repo: https://github.com/delphix/make-jpkg.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package makedumpfile

Git hash: b88f556f43790617f22060c10ace9e7e39c3e941
Git repo: https://github.com/delphix/makedumpfile.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package masking

{
  "dms-core-gate": {
    "git-hash": "a66790c6061f0e45d79e5561b1cee3d6bac7031e",
    "date": "2021-05-05T20:59:29+00:00"
  }
}

Git hash: a66790c6061f0e45d79e5561b1cee3d6bac7031e
Git repo: https://gitlab.delphix.com/masking/dms-core-gate.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/f0b1a8af/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/f0b1a8af/ppas

----------------------------------------
Package misc-debs

Fetched debs:
td-agent_3.5.0-delphix-2019.09.18.20_amd64.deb 84dfa2436039ff2a6312484bd7295ebaf570b5f59f100380b57e68b4800855c4
unzip_6.0-21ubuntu1_amd64.deb d46069c369ce88c8dd91c52abb8de8d6053606748ef18b3b9bc290fdd8ad2953
Main miror: http://linux-package-mirror.delphix.com/master/aaaaf778/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/aaaaf778/ppas

----------------------------------------
Package nfs-utils

Git hash: 2a9a5e763e988b50b969e5872b297459a31ebb64
Git repo: https://github.com/delphix/nfs-utils.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package performance-diagnostics

Git hash: 7c1fcc3f7f2e4bb5d2ea96a66e39c39dc8607d9c
Git repo: https://github.com/delphix/performance-diagnostics.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/490756d0/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/490756d0/ppas

----------------------------------------
Package ptools

Git hash: 9b66ad99b10c52f16494523b18ce993df9f3a283
Git repo: https://github.com/delphix/ptools.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/2336f99f/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/2336f99f/ppas

----------------------------------------
Package python-rtslib-fb

Git hash: e81c942ede519e1d3e36c4244bce5e26dbc444aa
Git repo: https://github.com/delphix/python-rtslib-fb.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package recovery-environment

Git hash: 6786d99e507b9bf8f141c0aa30e16e2b99da3ad2
Git repo: https://github.com/delphix/recovery-environment.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/0c5adf2c/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/0c5adf2c/ppas

----------------------------------------
Package savedump

Git hash: 47c0984ec2b0a0b48d877ab8c6c714d4ead41293
Git repo: https://github.com/delphix/savedump.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package sdb

Git hash: 3e6c69ab6ab498204aa4a66ea985a61ea1d2f12a
Git repo: https://github.com/delphix/sdb.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/ba99df38/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/ba99df38/ppas

----------------------------------------
Package targetcli-fb

Git hash: a78282a9623175ee199552f1f50a9d98a0c22752
Git repo: https://github.com/delphix/targetcli-fb.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/02dbac36/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/02dbac36/ppas

----------------------------------------
Package upgrade-verify

{"delphix-upgrade-verification": {
    "git-hash": "e34d1a4a0035308987cd068cde975246b4015ddb",
    "date": "2021-05-04T22:38:30+0000",
    "version": "1.0.11"
}}

----------------------------------------
Package virtualization

{
  "dlpx-app-gate": {
    "git-hash": "653e1b32636947989e74a1af9c12e569de0e0fbd",
    "date": "2021-05-07T13:15:00+00:00"
  }
}

Git hash: 653e1b32636947989e74a1af9c12e569de0e0fbd
Git repo: https://gitlab.delphix.com/app/dlpx-app-gate.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/474e89c7/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/474e89c7/ppas

----------------------------------------
Package zfs

Git hash: 82bfe23117efb50c04074383b3ed11875de60fd5
Git repo: https://github.com/delphix/zfs.git
Git branch: master
Main miror: http://linux-package-mirror.delphix.com/master/f0b1a8af/ubuntu
Secondary miror: http://linux-package-mirror.delphix.com/master/f0b1a8af/ppas

----------------------------------------```

**Comment by matt-skinner-delphix** [comment]
Have you thought of using a standardized output format for the build-info file rather than what looks like a unique format  to make any machine/program processing easier?

**Comment by pzakha** [comment]
> Have you thought of using a standardized output format for the build-info file rather than what looks like a unique format to make any machine/program processing easier?

This was originally meant to be consumed by humans only, so I didn't design this to be machine parseable. Given that the build info generated by each package can be arbitrary, if we wanted to make it machine-parseable more easily we'd probably need to refactor this and create multiple files, at which point we should probably generate a separate package for this build-info.

I'm not opposed to the latter as it could be useful in other workflows.

**Comment by matt-skinner-delphix** [comment]
> > Have you thought of using a standardized output format for the build-info file rather than what looks like a unique format to make any machine/program processing easier?
> 
> This was originally meant to be consumed by humans only, so I didn't design this to be machine parseable. Given that the build info generated by each package can be arbitrary, if we wanted to make it machine-parseable more easily we'd probably need to refactor this and create multiple files, at which point we should probably generate a separate package for this build-info.
> 
> I'm not opposed to the latter as it could be useful in other workflows.

👍

**Comment by pzakha** [comment]
I've refactored the previous logic to generate a `delphix-build-info` package instead of a file included in `delphix-entire`. I've also changed the logic so that some of the metadata is now required to be populated otherwise the build will fail. The new logic is cleaner than the previous one and the generated files can be parsed more easily by automation. Aside from giving developers/support useful debug info, here are some of the potential uses for this metadata:
1. The ARTIFACTS_S3_LOCATION data can be useful when creating a sparse hotfix image intended to be applied on this specific appliance.
2. As a pre-release step, the GIT_HASH files for each package could be read and compared against the tip of the release branches for each package, to make sure that the appliance does indeed contain the expected versions of all packages.

Here's the file structure for this new package:
```
/var/delphix-buildinfo$ tree
.
├── appliance-build
│   ├── ARTIFACTS_S3_LOCATION
│   ├── DELPHIX_APPLIANCE_VERSION
│   ├── DELPHIX_PACKAGE_MIRROR_MAIN
│   ├── DELPHIX_PACKAGE_MIRROR_SECONDARY
│   ├── GIT_BRANCH
│   ├── GIT_HASH
│   └── GIT_URL
└── packages
    ├── COMPONENTS
    ├── KERNEL_VERSIONS
    ├── adoptopenjdk
    │   ├── BUILD_INFO
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── bcc
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── bpftrace
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── challenge-response
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── cloud-init
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── connstat
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── crash-python
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── crypt-blowfish
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── delphix-kernel
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── delphix-platform
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── delphix-sso-app
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── docker-python-image
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── drgn
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── gdb-python
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── grub2
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── host-jdks
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── libkdumpfile
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── linux-kernel-aws
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── linux-kernel-azure
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── linux-kernel-gcp
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── linux-kernel-generic
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── linux-kernel-oracle
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── make-jpkg
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── makedumpfile
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── masking
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   ├── PACKAGE_MIRROR_URL_SECONDARY
    │   └── metadata.json
    ├── misc-debs
    │   ├── BUILD_INFO
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── nfs-utils
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── performance-diagnostics
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── ptools
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── python-rtslib-fb
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── recovery-environment
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── savedump
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── sdb
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── targetcli-fb
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   └── PACKAGE_MIRROR_URL_SECONDARY
    ├── upgrade-verify
    │   └── metadata.json
    ├── virtualization
    │   ├── BUILD_INFO
    │   ├── GIT_HASH
    │   ├── PACKAGE_MIRROR_URL_MAIN
    │   ├── PACKAGE_MIRROR_URL_SECONDARY
    │   └── metadata.json
    └── zfs
        ├── BUILD_INFO
        ├── GIT_HASH
        ├── PACKAGE_MIRROR_URL_MAIN
        └── PACKAGE_MIRROR_URL_SECONDARY
```
Here are some of the files:
```
/var/delphix-buildinfo/appliance-build$ for file in *; do echo "$file: $(cat $file)"; done
ARTIFACTS_S3_LOCATION: s3://dev-de-images/builds/jenkins-selfservice/devops-gate/master/appliance-build/master/pre-push/6623
DELPHIX_APPLIANCE_VERSION: 6.1.0.0-snapshot.20210623173812022+jenkins-selfservice-devops-gate-master-appliance-build-master-pre-push-6623
DELPHIX_PACKAGE_MIRROR_MAIN: http://linux-package-mirror.delphix.com/master/de4e9474/ubuntu
DELPHIX_PACKAGE_MIRROR_SECONDARY: http://linux-package-mirror.delphix.com/master/de4e9474/ppas
GIT_BRANCH: dlpx-71683-test
GIT_HASH: 36b912017e7632edc272a27c5172ad1aa3a8203e
GIT_URL: https://github.com/pzakha/appliance-build
```

**Comment by matt-skinner-delphix** [comment]
Thanks for taking the effort to make it machine readable. 👍

File: scripts/create-build-info-package.sh [hunk 4]
```diff
@@ -0,0 +1,113 @@
...
+
+#
+# Build the package
+#
+cd "$TOP/build-info"
+#
+# We include some random characters in the version string to avoid collisions
+# with other build-info packages built for different platforms but included in
+# the same upgrade image. The timestamp is insufficient here since the other
+# platforms are built in parallel.
```
prakashsurya, line 104: Can you help me understand this a little better.. as-is, we'll build multiple identical packages for each platform, where the only difference is the package name? why can't we use the same package for all platforms? or, why can't we build a unique/different package for all platforms, but have the name and version be the same for all platforms? or, why can't we use the platform name here, rather than some random bits?

scripts/create-build-info-package.sh
```diff
@@ -0,0 +1,113 @@
...
+	# all environment to be set, so skip the env check.
+	#
+	[[ -n "$JENKINS_URL" ]] || return 0
+
+	local val="${!1}"
+	[[ -n "$val" ]] || die "check_env: $1 must be non-empty"
+	return 0
+}
+
+check_env APPLIANCE_BUILD_GIT_URL
```
prakashsurya, line 83: where is this `check_env` defined?

prakashsurya: nvm.. I see it just above.

prakashsurya (commented):
Can you comment on why it's preferred to build this package here, in appliance-build, rather than in the combine-packages jenkins job (such that appliance-build could consume the package built in combine-packages, the same way as any other package)?

prakashsurya:
> Can you comment on why it's preferred to build this package here, in appliance-build, rather than in the combine-packages jenkins job (such that appliance-build could consume the package built in combine-packages, the same way as any other package)?

And while I know I've asked this before.. I ask again, because I feel like we're getting closer to a solution that would make even more sense to have combine-packages build this new package.. for example, given the two points raised previously:

> 1. The combine-packages job is ran after any package is built, so it was meant to be a lightweight job that runs fairly quickly.

I'd consider building a simple package like this to be relatively "lightweight", and on the order of the same amount of work we already do in the combine-packages job.

> 2. Generating a new metapackage package requires installing some packages, which we cannot do directly given that combine-packages doesn't clone a new VM for its execution and runs on a "misc" Jenkins agent.

I'd expect we can "solve" this issue by doing the build in a docker image via Jenkins, with the correct dependencies installed, rather than relying on the "misc" agent to have the necessary dependencies installed.

I'm not necessarily against running this in appliance-build, just want to understand the motivations for choosing this approach given the recent refactorings, and see if the motivations have changed at all.

prakashsurya:
Further, we only run this script when it's executing within a Jenkins job, which only adds to my suspicion that it likely makes the most sense to run via combine-packages (which only runs via Jenkins), than appliance-build (which can be run outside of Jenkins).

scripts/create-build-info-package.sh
```diff
@@ -0,0 +1,113 @@
...
+# platforms are built in parallel.
+#
+rnd="$(uuidgen | tr -d '-' | fold -w 8 | head -n 1)"
+version="1.0.0-delphix-$(date '+%Y.%m.%d.%H.%M.%S')-$rnd"
+dch --create --package delphix-build-info -v "$version" \
+	"Automatically generated changelog entry."
+
+dpkg-buildpackage -uc -us -b
+
+mv ../delphix-build-info*deb "$PKG_DIR/packages/"
```
prakashsurya, line 113: How do we pull this package into the appliance? Does another package depend on this new package?

prakashsurya (commented):
overall, this looks reasonable to me.. I'd like to understand why we need to create multiple versions of this package (with identical contents, if I'm understanding correctly), and how this package gets installed into the appliance, before re-approving.

scripts/create-build-info-package.sh
pzakha: > Can you help me understand this a little better.. as-is, we'll build multiple identical packages for each platform, where the only difference is the package name? why can't we use the same package for all platforms?

Having the same package built for all platforms would require us to have it built outside of the script called by appliance-build-stage1, which would require quite some refactoring. Technically we could do something like that for both this package and the ancillary itself repository but I don't think we have a good enough motivation for undertaking this extra work.

>  or, why can't we build a unique/different package for all platforms, but have the name and version be the same for all platforms? or, why can't we use the platform name here, rather than some random bits?

I think this would be quite reasonable actually. I didn't do it because there was no real per-platform differences, but it would be better than adding a random string.

scripts/create-build-info-package.sh
pzakha: Yeah, I have a separate change that adds this package to delphix-platform dependencies

scripts/create-build-info-package.sh
prakashsurya: > which would require quite some refactoring.

IMO, another reason it'd make more sense to build this package via combine-packages.

prakashsurya: Since we'll have multiple packages, each with different versions, this means the upgrade image will essentially have this package duplicated for each platform, right?

prakashsurya: Presumably, a hotfix image will contain the duplicate packages as well? even though this package is irrelevant to any issue being fixed in the hotfix image?

I think even if we refactored to only have a single package (not per-platform), we'd still have to include the package in the hotfix image, even though the package is irrelevant to the hotfix.

scripts/create-build-info-package.sh
prakashsurya: Do you have that change posted anywhere? Curious how that'll work..

Does this mean, the VM image built for a platform of a given build, might result in a different version of this package installed than a system upgraded to the upgrade image matching that same build? E.g. the upgrade image will have multiple versions of this package installed, such that on upgrade, the system may pull in the version from a build of a different platform (i.e. whatever verison is the latest in the upgrade image)? whereas the VM image will always pull in the version of the build-info package built for that specific platform?

prakashsurya: Actually.. maybe both upgrade and the initial VM image will still match.. since it looks like the `delphix-entire` package will pin the correct package version for the platform..? i.e. since we `apt-get install` the package matching the version in the platform specific `delphix-entire` package, I think my concern above is not the case...

pzakha:
Regarding whether we want to build this in combine-packages or not, I see pros ad cons to both approaches:
PROs for combine-packages:
- no duplicate packages created with the same content
- the combine-packages bundle will contain all the Delphix-generated packages (except delphix-entire, and possibly the future hotfix-metadata package)

CONs for combine-packages:
- We'd probably want to avoid running this logic on the misc agent as this could very likely be a performance bottleneck, so we'd have to clone a separate VM.
- If we download all packages in combine-packages where we run this logic we'd incur costs for additional storag now that every combine-packages bundle will need to contain all the debs.

PROs for appliance-build:
- can include appliance-build metadata (which is what prompted me to do this refactoring in the first place)

scripts/create-build-info-package.sh
pzakha: This package could be useful for the hotfix image as well, since it can be used to validate that the hotfix image indeed contains the packages we want it to contain.

prakashsurya:
> If we download all packages in combine-packages where we run this logic we'd incur costs for additional storag now that every combine-packages bundle will need to contain all the debs.

We need the actual `.deb` files to run this script? I thought it was just operating on the same information that combine-packages uses (which isn't the actual `.deb` files).. but maybe I'm misunderstanding..

> can include appliance-build metadata (which is what prompted me to do this refactoring in the first place)

Yea, agreed.. we'd have to do this some other way.. I'd say this is a "con" for doing this in combine-packages (as well as a "pro" for appliance-build).

> We'd probably want to avoid running this logic on the misc agent as this could very likely be a performance bottleneck, so we'd have to clone a separate VM.

would it be any more than a bottleneck than it currently is? it's not clear to me why this would be a new bottle neck..

scripts/create-build-info-package.sh
prakashsurya: Yea, I guess that's true. Can you comment on the size of the package within the upgrade image? Presumably it's on the order of KBs, since it's just a collection of text files? I just want to confirm that it's a trivial size increase to sparse hotfix images.

prakashsurya (approved):
Looks reasonable to me.

scripts/create-build-info-package.sh
pzakha: Yeah, the package is tiny, it's 8.8KB.

pzakha:
> We need the actual .deb files to run this script? I thought it was just operating on the same information that combine-packages uses (which isn't the actual .deb files).. but maybe I'm misunderstanding..

I think you are right we probably don't need the debs.

> would it be any more than a bottleneck than it currently is? it's not clear to me why this would be a new bottle neck..

I've been a bit concerned it could be already a bottleneck, but we fortunately don't run that much logic currently. Ideally this would be solved in ops, where we'd have a cluster of VMs for this type of jobs but I'm not sure what type of changes that would require.

scripts/create-build-info-package.sh
pzakha: > I think this would be quite reasonable actually. I didn't do it because there was no real per-platform differences, but it would be better than adding a random string.

Actually looking at it more, it would require some additional changes that I'm not sure we want to make. Currently the `ancillaryRepository` gradle task is platform-agnostic, so we'd have to make it per-platform which doesn't seem to make much sense conceptually.

scripts/create-build-info-package.sh
pzakha: delphix-platform change is trivial for the reasons you mentioned: https://github.com/pzakha/delphix-platform/commit/2d90030b28d0d05de2a2ec4023668d24e5eb7ed1

scripts/create-build-info-package.sh
prakashsurya: > Having the same package built for all platforms would require us to have it built outside of the script called by appliance-build-stage1

Can you clarify why this refactoring would be necessary? just so I understand..

e.g. why can each per-platform build, generate the same package at the same version? if each unique build did this, what would fail?

pzakha: I've had some issues with apt in the past when there were 2 slightly different packages with the same version number, but I don't recall the details. That said, I was curious so I've decided to test it out here just to see what would happen.

I've tried building and image where the timestamp & random digits are removed, and it does seem like the build and the upgrade image generation succeeds (see http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-stage0/job/pre-push/2422/console). We can see that the aptly repo used for the upgrade image only has a single instance of the delphix-build-info package added.

In practice we'd still want to have a timestamp included in the revision number, so unless we generate a timestamp outside of this script and pass it to all the individual build jobs, we'd likely have multiple build-info packages with slightly different timestamps. I've also compared the produced packages and they are almost the same (the only difference is in the changelog files as they are auto-generated by dch and include a timestamp).

So in short, it seems like we have some other options here, we just need to figure out what we want. I personally don't have much of a preference either way, but if we want to remove the random characters we'd probably want to pass a pre-established timestamp or some other string that is unique to this build but common among all sub-build-jobs.

prakashsurya: > In practice we'd still want to have a timestamp included in the revision number

Well, I was thinking that maybe we could/should use the `DELPHIX_APPLIANCE_VERSION`.. similar to what we do for `delphix-entire`.. that will automatically include build specific information for builds that are not within the code complete window.. or just the version, e.g. `6.0.9.0`, for builds within code complete.

If we did it this way, we might also need to slightly modify the `execute` script for upgrade, to ensure we always `--reinstall` this package.. similar to what we do for `delphix-entire`..

With that said, I'm not convinced using `DELPHIX_APPLIANCE_VERSION` is better.. just something that I thought about.

Do you have any thoughts about such an approach? I'm fine with what you have, as-is.. just thinking out loud, here..

pzakha: Personally I'd rather just leave the logic as is given how small the extra packages are and given that it should avoid any of those issues.

prakashsurya: I mean sure, I'm fine with that.. but just to be clear, this is a less than optimal solution (i.e. building and including multiple packages, rather than a single package, in the resultant upgrade and hotfix images).

My approval still stands, so feel free to land.

PR #497, opened 2020-11-30
DLPX-72869 Patch VMDK generation tool to mark that the vmdk contains a vmtools install

pzakha:
I've downloaded the original livecd-rootfs package, manually applied the patch (attached in JIRA) and repackaged the package.

An alternative approach would be to patch the installed package after its installation (see https://github.com/pzakha/appliance-build/commit/99c88ca4121e3c6f3d44e56a6edba6dffb1cf034). However the patch could fail to apply in the future if Ubuntu was to update the package with some changes in that file.

We will need to create a bug in JIRA to remember to revert back to the default Ubuntu package once that patch is integrated upstream and pulled into bionic. This bug is being tracked upstream by https://bugs.launchpad.net/ubuntu/+source/open-vm-tools/+bug/1893898.

## Motivation
Without this fix, Guest OS Customizations cannot be applied when cloning this OVA as ESX thinks that open-vm-tools is not installed on the Guest image and so GOSC would fail (See https://communities.vmware.com/t5/VMware-vCenter-Discussions/VM-customization-does-not-work-until-booted-once/m-p/522372). Without this fix, the VM must be booted at least once, at which point ESX will detect that open-vm-tools is running, and store the version of the open-vm-tools package inside the vmdk file; however we want to be able to apply GOSC on a clone of an OVA that has never been booted, as the Delphix Appliance does a bunch of customizations on first boot.

## Open Questions
Do we want to find another place for the package in artifactory? I've put it under linux-pkg because that's the bucket I had write access to, but it might not be the best place for it.

## Testing
- ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/4361/
- Checked appliance-build output and made sure custom package was installed.
- Manually cloned produced OVA and verified that GOSC works.

### DIFF
--- a/bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
+++ b/bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
@@ -44,6 +44,20 @@
       - zfsutils-linux
     state: present
 
+#
+# See DLPX-72860 for more info on the custom package.
+#
+- name: Custom livecd-rootfs package | Download
+  get_url:
+    url: 'https://artifactory.delphix.com:443/artifactory/linux-pkg/livecd-rootfs/6.0.6.0/livecd-rootfs_2.525.47-delphix1_amd64.deb'
+    dest: '/var/tmp/livecd-rootfs_2.525.47-delphix1_amd64.deb'
+    checksum: 'sha256:9f090adf288d115b2eb10d2dced2a76113339eb95dc5db91fac4b89b2bef07a0'
+
+- name: Custom livecd-rootfs package | Install
+  apt:
+    deb: '/var/tmp/livecd-rootfs_2.525.47-delphix1_amd64.deb'
+    state: present
+
 - modprobe:
     name: zfs
     state: present


prakashsurya:
Out of curiosity, if we need to make changes to an upstream package, why aren't we pulling that modification into the linux-pkg build framework? or have I missed where we're using linux-pkg with this upstream package? It just seems a little odd that we'd rely on some manual process to patch and build this new build dependency, when we have another process for doing that, which we use for modifications to other packages.

pzakha:
@prakashsurya I was considering that, I just did it this way because it's faster. Even if we were to build the package with linux-pkg we would likely still want to manually upload it into artifactory as we don't currently have a way for appliance-build to pull custom-built packages from S3 in the bootstrap phase.

Also Linux-pkg would have made more sense if we expected to keep this change for a while and then needed to leverage auto-update logic, however in this case the change has already been fixed upstream, so we are mostly just waiting for Ubuntu to pull it into bionic.

pzakha:
With all that said, we may need to review how we pull dependencies into appliance-build during the bootstrap phase as pulling the latest changes every time prevents us from guaranteeing reproducible builds. If we were to go that route, it might make sense to have all the packages in linux-pkg so that they can be properly tagged with a release. However, given that in this case this package is part of the build rather than the product, I'm not sure if it would need to be treated differently from the others.

prakashsurya:
> so we are mostly just waiting for Ubuntu to pull it into bionic.

Do we have an ETA on when it'll be in bionic?

In general, I'm hesitant to adopt short term solutions. For example, once this change gets integrated into a tagged release (I assume it'll be included in the next release), how will we phase it out, such that we no longer require this manually built package from artifactory when building that tag? I think we'll need to maintain this ad-hoc package indefinitely; i.e. as long as we support building any tagged release this change is contained in (a timespan that's ill-defined, due to hotfixes/etc).

> With all that said, we may need to review how we pull dependencies into appliance-build during the bootstrap phase as pulling the latest changes every time prevents us from guaranteeing reproducible builds.

Yea, I agree. I think we should probably use our pinned mirrors for that case. Perhaps, in addition to making linux-pkg built packages available to be pulled in; in which case, we could leverage linux-pkg more seamlessly here.

pzakha:
I don't have the exact ETA for when this will be in bionic, but it has already been backported to
`livecd-rootfs:ubuntu/bionic` and assigned a release tag (see https://code.launchpad.net/~ubuntu-core-dev/livecd-rootfs/+git/livecd-rootfs/+ref/ubuntu/bionic).

> I think we'll need to maintain this ad-hoc package indefinitely; i.e. as long as we support building any tagged release this change is contained in (a timespan that's ill-defined, due to hotfixes/etc).

Note that we don't need to "maintain" that package, we just need to make sure we don't delete it from artifactory for as long as we want to be able to build 6.0.6.0. This is not very different from other debs that we keep in artifactory (e.g. adoptopenjdk, td-agent, linux kernel modules with iscsi fixes).

prakashsurya:
> This is not very different from other debs that we keep in artifactory (e.g. adoptopenjdk, td-agent, linux kernel modules with iscsi fixes).

Yea, agreed; although, I have similar reservations about the solution taken in those cases as well.

> we just need to make sure we don't delete it from artifactory for as long as we want to be able to build 6.0.6.0.

I understand; we've been using that approach for a while (e.g. our illumos build servers), but I'm not sure I view that as an ideal solution. If/when it gets removed (perhaps by accident, corruption, etc.), it'd be good to (at least) have a written plan in place to replace it; and perhaps documentation to make this special case abundantly clear to anybody working/relying on our build system (more documentation would be good for lots of things, so nothing new here).

Can I at least request we add a document (e.g. perhaps add a `docs` top-level directory) to outline the process taken to build the original package, and perhaps some instructions on how it could be replaced if that's ever needed? I understand the precedence is already set, to not requirement such documentation, but I'd like to change that (e.g. I'm not sure how we'd recover from a situation where we lost all of these pre-built packages, and couldn't rely on you to rebuild them anymore).

pzakha:
Sure, I can document the steps in DLPX-72869 in JIRA.

prakashsurya:
> We should port that patch to our version of livecd-rootfs so that vmware is aware that on first boot, the vmtools daemon will be running

Can I also request we add some information, either to the bug or this PR, to describe _why_ we need to do this.. as-is, the motivation for making these changes isn't clear (to me, at least, based solely on the information in the bug and PR). E.g. why do we require that vmware is aware of vmtools on first boot?

prakashsurya:
> Sure, I can document the steps in DLPX-72869 in JIRA.

This is a little tangential, but IMO it'd be good to start consolidating our documentation to some common place.. I'd expect it'll be difficult to find the comment with these instructions if it's needed later (e.g. it was difficult to find the hotfix instructions recently, without being pointed to the JIRA comments by others).

But, since we haven't settled on a solution for storing docs like this as a team, I'm OK with using a JIRA comment for this.

prakashsurya (approved):

pzakha:
I've updated the description with the motivation for this change.

pcd1193182 (approved):

PR #498, opened 2020-12-01
DLPX-72429 ssh to admin CLI shows error message after force-not-in-place upgrade, permissions issue on /export dir

sanbhara:
Bitesized fix to add umask to the upgrade-container script. Added it at top of the script so it is uniform since this is the entry-point to scripted upgrade operations initiated by the mgmt stack.

git ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/4391/


pzakha (approved):

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -15,6 +15,13 @@
 # limitations under the License.
 #
 
+#
+# set umask to 0022 so that all new directories will be created with correct
+# permissions. This is especially required for not-in-place upgrades. This will
+# override the umask set by mgmt stack.
+#
+umask 0022
```
prakashsurya, line 23: I believe this is already the default umask, so is part of the problem the fact that the app-gate changes this?

I think I might agree with Pavel, that we could/should apply this more generally by placing it in `common.sh`.. or alternatively, we could have a more targeted fix, e.g. by making a change in `mount_upgrade_container_dataset` where I believe the problematic `mkdir` is called.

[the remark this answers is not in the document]
sanbhara: Ok I can move it to common.sh which is called by most of the scripts.

sanbhara: Yes the issue is due to the umask being set when calling these scripts. I will move it to common.sh.

sebroy: Please be careful with such an approach. Countless bugs have been introduced by code that has set umask globally under the assumption that it was the "correct" umask for everyone, only to find out that some script implicitly depended on some umask that it had inherited by accident. This is not without risk.

prakashsurya: Yea. I think, in general, I'd rather we make the logic independent of w/e the umask is.. so if we need something specific, it may be better to target that specific `mkdir` call to work regardless of what the umask is.

But, I'm good with whatever works.. if going back to the default umask, works, I'm OK with that.. but comes with it's own set of risk(s).

sanbhara: @sebroy yes that is a risk, but in this particular case, I intend to set umask to 0022 which is the bash default. I would hope that any new script/statements that require a different umask would explicitly set it themselves. Is that a reasonable assumption? This particular common.sh is used by the upgrade scripts only.

pzakha: I think for most platform-related scripts it is assumed that umask would be `0022`, whereas the app-gate is a special more restrictive case.

prakashsurya:
I'll work on fixing the ansible check; that's unrelated, but needs to be fixed so we can land this, so I'll start on that.

upgrade/upgrade-scripts/common.sh
```diff
@@ -15,6 +15,13 @@
 # limitations under the License.
 #
 
+#
+# set umask to 0022 so that all new directories will be created with correct
+# permissions. This is especially required for not-in-place upgrades. This will
+# override the umask set by mgmt stack.
```
prakashsurya, line 21: I'd recommend we slightly modify this comment, now that it's contained in a more generic location.. e.g. something like:
```
#
# The virtualization service uses a different umask than the default. Thus to account for the
# fact that these scripts may be running with a non-default umask, we explicitly change it back
# to the default value here. This helps ensure any files and directories generated by these scripts,
# will be created with the correct permissions, regardless of the umask of the calling process.
#
```

prakashsurya (approved):

upgrade/upgrade-scripts/common.sh
```diff
@@ -15,6 +15,13 @@
...
+#
+umask 0022
+
 # shellcheck disable=SC2034
```
pzakha, line 25: You should put this after the shellcheck instruction, otherwise the shellcheck instruction doesn't apply to the whole file (and you get a shellcheck failure)

upgrade/upgrade-scripts/common.sh
sanbhara: That looks good.

upgrade/upgrade-scripts/common.sh
sanbhara: Got it, thanks

PR #506, opened 2020-12-04
Enable hot-add

pcd1193182:
This PR instructs VMWare that the Delphix Engine can handle memory and CPU hot-add. It has been tested in conjunction with the associated zfs and app-stack changes, and behaved correctly.

### DIFF
--- a/live-build/config/hooks/vm-artifacts/template.ovf
+++ b/live-build/config/hooks/vm-artifacts/template.ovf
@@ -130,12 +130,12 @@
         <rasd:ResourceType>10</rasd:ResourceType>
         <vmw:Config ovf:required="false" vmw:key="wakeOnLanEnabled" vmw:value="true"/>
       </Item>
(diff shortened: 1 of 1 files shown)


prakashsurya (approved):

ahrens:
I assume you would have to reboot for this to take effect (and therefore you'd get the new ZFS bits as well).

Do these changes only apply to newly-deployed VM's?  For existing VM's, is there a way to tell VMware that the VM now supports hot-add?

prakashsurya:
This will only affect new VMs; I was assuming that was the intention with my review. As far as I know, we don't have any automated why to modify the VMX file after the OVA is deployed; although, I'm sure the VMX could be manually modified by the customer/support, if they have access to the file.

pcd1193182:
There is an option in the UI to enable hot-adding of CPU and memory, under the VM settings, but it can only be modified while the VM is off. As far as I know there is no way to enable it while the VM is running.

ahrens (approved):

PR #510, opened 2020-12-14
DLPX-73299 appliance-build: Fetch upgrade-verify package from combined-packages bundle

### LINKED ISSUE #509: Running appliance-build manually without Jenkins fails when fetching linux-pkg packages
This was caused by the linux-pkg-rework.
Error is:
```
+ aws s3 sync --only-show-errors s3://snapshot-de-images/builds/jenkins-ops/devops-gate/master/linux-pkg/master/combine-packages/post-push/350/packages .
+ for subdir in */
+ pushd '*/'
FAILURE: Build failed with an exception.
```
The reason for the failure is that the `packages` directory is created by Jenkins when resolving and fetching all the component packages in the combined-packages bundle.

pzakha:
This is a follow-up to http://reviews.delphix.com/r/64085/, which adds the upgrade-verify package to the combine-packages bundle.

Fixes #509

## Dependencies
Must be pushed after http://reviews.delphix.com/r/64085/

## Testing
- See comments in https://jira.delphix.com/browse/TOOL-10766 for the testing that was done
- Manually tested that this fixes issue #509

### DIFF
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -15,6 +15,7 @@
 # limitations under the License.
 #
 
+# shellcheck disable=SC2034
 TOP=$(git rev-parse --show-toplevel 2>/dev/null)
 
 function die() {
@@ -24,8 +25,7 @@ function die() {
 
 function resolve_s3_uri() {
 	local pkg_uri="$1"
-	local pkg_prefix="$2"
-	local latest_subprefix="$3"
+	local latest_subprefix="$2"
 
 	local bucket="snapshot-de-images"
 	local jenkinsid="jenkins-ops"
@@ -33,10 +33,6 @@ function resolve_s3_uri() {
 
 	if [[ -n "$pkg_uri" ]]; then
 		resolved_uri="$pkg_uri"
-	elif [[ "$pkg_prefix" == s3* ]]; then
-		resolved_uri="$pkg_prefix"
-	elif [[ -n "$pkg_prefix" ]]; then
-		resolved_uri="s3://$bucket/$pkg_prefix"
 	elif [[ -n "$latest_subprefix" ]]; then
 		aws s3 cp --quiet \
 			"s3://$bucket/builds/$jenkinsid/$latest_subprefix" .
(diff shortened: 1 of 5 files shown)


scripts/common.sh
```diff
@@ -55,40 +51,87 @@ function resolve_s3_uri() {
...
+# When appliance-build is ran via Jenkins, the Jenkins job copies the original
+# combined-packages artifacts to a new S3 location, then dereferences the
+# COMPONENTS file and copies all individual package artifacts into a
+# "packages" directory created under that new S3 location. Jenkins then passes
+# that combined-packages URI to live-build.
+#
+# Thus if a "packages" directory is found under the combined-packages S3 URI,
+# we assume that the dereferencing has already been done and so we just sync
+# the whole directory. Otherwise, we must dereference the COMPONENTS file here
+# and fetch the artifacts for each package.
```
prakashsurya, line 72: Why do we have two different situations, one where the "packages" directory will already exist, and another where it wont yet exist? Can we consolidate to a single mechanism?

pzakha: The situation where the "packages" directory doesn't exist is really just for being able to run `appliance-build` manually without Jenkins. Right now the `combine-packages` job itself doesn't generate the packages directory in order to avoid copying all package artifacts every time that job is called (that would also potentially require cloning a DCoA VM each time we run combine-packages). If we decide to change that in the future, we can remove that check.

prakashsurya: In what case will the "packages" directory exist? and why does it create the directory in that case?

pzakha: We create the "packages" directory in `appliance-build-stage0` when we copy the package artifacts from their original locations into the new "input-artifacts" S3 location. We do that before calling `appliance-build-stage1`

prakashsurya: ah, ok.. I see.

prakashsurya (approved):

matt-skinner-delphix:
👍 Looks good to me.

matt-skinner-delphix (approved):

sebroy (approved):

PR #520, opened 2021-02-03
DLPX-74045 adapt upgrade "execute" script to better handle new hotfix workflow

prakashsurya:
Currently when applying a hotfix image generated by the new hotfix
workflow being developed in CP-4190, nothing will happen when the
upgrade is attempted. This is due to the "delphix-entire" package in the
hotfix being installed, matching the version of the package that's
already installed. Thus, the package manager chooses not to install the
new packagecontained in the upgrade/hotfix image, since it thinks the
same package is already installed, due to the two having the same
version.

We need to adapt the "execute" script that's used during upgrade, such
that we always install the new "delphix-entire" package contained in the
upgrade/hotfix image, even if the package versions are identical.

### DIFF
--- a/upgrade/upgrade-scripts/execute
+++ b/upgrade/upgrade-scripts/execute
@@ -303,8 +303,22 @@ dpkg-query -Wf '${Package}\n' "delphix-kernel-*" | xargs apt-mark manual ||
 #
 
 # shellcheck disable=SC2153
-apt_get install -y --allow-downgrades "delphix-entire-$platform=$VERSION" ||
+apt_get install \
+	-y --allow-downgrades --reinstall \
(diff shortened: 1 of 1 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/4697/)

prakashsurya:
I tested this works properly with a hotfix image generated by my devops-gate changes [here](http://reviews.delphix.com/r/65472/).

pzakha (approved):

sanbhara (approved):

PR #533, opened 2021-04-01
DLPX-75089 Upgrade verification should not enable nfs services

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/4986/)

sebroy:
Can you describe how the fix was tested (beyond ab-pre-push) to verify that the fix works?

prakashsurya:
> Can you describe how the fix was tested (beyond ab-pre-push) to verify that the fix works?

Sure; using a trunk-based VM, I applied this change to an unpacked trunk-based upgrade image, created an upgrade container, and verified these services were not running within that upgrade container. I also verified that without this change, these services do run within an upgrade container.

prakashsurya:
Additionally, @don-brady tested the upgrade image produced by the `git-ab-pre-push` run with his reproducer (using VDBs on Solaris targets), and verified that with this change, the VDBs no longer go down when upgrade verification is run via the product.

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -262,6 +262,27 @@ function create_upgrade_container() {
...
 	EOF
 
+	#
+	# DLPX-75089 - Since older versions of Delphix did not properly
+	# disable the NFS services within the upgrade container, we have
+	# to dynamically disable them here. This way, when the container
+	# is started, those NFS services will not run (no matter the
+	# version of Delphix we're upgrading from).
+	#
+	for svc in "nfs-mountd" "nfs-server" "rpc-statd"; do
```
don-brady, line 272: to be complete, we should also include `rpc-statd-notify` service.  It only runs on a reboot but I'm not sure if starting the container acts like a reboot.

sebroy (approved):

upgrade/upgrade-scripts/upgrade-container
prakashsurya: Thanks, done.

don-brady (approved):

PR #536, opened 2021-04-07
Don't remove "delphix-entire" when generating image diffs

prakashsurya:
Currently, if we generate an upgrade image diff from two images that
have the same delphix-entire package version, the resultant Aptly
repository will not contain any delphix-entire package in it. This is
problematic when that Aptly repository is later used with the script
used to generate an upgrade image from that repository, as the script
requires the repository to contain a "delphix-entire" package in it.
Thus, to avoid this issue, this change modifies the image diff logic to
ensure we never remove the "delphix-entire" package.

### DIFF
--- a/scripts/aptly-repo-from-image-diff.sh
+++ b/scripts/aptly-repo-from-image-diff.sh
@@ -102,9 +102,28 @@ aptly repo search image-a | xargs aptly repo copy image-a upgrade-repository ||
 # Here we're performing step 3 from the comment above, but since the
 # "delphix-upgrade-verification" and "delphix-entire" packages are a bit
 # different than most other packages on a Delphix appliance, we need to
-# handle these packages uniquely here. Specifically, we want to ensure
-# these packages are always contained in the resultant aptly repository
-# (even if the package is the same within both image A and image B).
+# handle these packages uniquely here. Specifically, we want to ensure these
+# packages are always contained in the resultant Aptly repository (even if
+# the package is the same within both image A and image B), as each package
+# is essential to the proper functioning of any Delphix appliance upgrade.
+#
+# The "delphix-upgrade-verification" package is used to perform upgrade
+# specific logic (and verification) during the upgrade process, such that we
+# can avoid common pitfalls that would result in the upgrade process
+# failing. Without this package being contained in the repository (and thus,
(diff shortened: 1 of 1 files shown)


scripts/aptly-repo-from-image-diff.sh
```diff
@@ -100,14 +100,15 @@ aptly repo search image-a | xargs aptly repo copy image-a upgrade-repository ||
...
-# "delphix-upgrade-verification" package is a bit different than most
-# other packages on a Delphix appliance, we need to handle that package
-# uniquely here. Specifically, we want to ensure this package is always
-# contained in the resultant aptly repository (even if the package is
-# the same within both image A and image B).
+# "delphix-upgrade-verification" and "delphix-entire" packages are a bit
+# different than most other packages on a Delphix appliance, we need to
+# handle these packages uniquely here. Specifically, we want to ensure
+# these packages are always contained in the resultant aptly repository
+# (even if the package is the same within both image A and image B).
```
matt-skinner-delphix, line 107: I remember reading elsewhere why we have delphix-entire package in the first place. Could we replicate that reasoning here again (or reference where one might learn more) so that if a developer is introduced to delphix-entire here they are not left wondering why they are different or worse just trusting this comment and moving on?

prakashsurya: I don't think here is the right place to explain what the delphix-entire package is, and/or why it's different.. similarly, I don't think it's the right place to explain what the delphix-upgrade-verification package is, and/or why it's different.

I do think it'd be reasonable to drop a reference to some other location where we explain that.. I'm not sure where to refer the read to, though. Do you have an idea of where we'd refer them to? I don't think we have either of those packages documented in a place that'd be easy to refer to...

prakashsurya: I think it might be reasonable to touch on why we always want these packages to remain in the sparse image, though.. I could expand a bit on that, specifically.

prakashsurya: I just added some more commentary, let me know if you think that's sufficient (or not).

sebroy (approved):

matt-skinner-delphix (approved):

PR #537, opened 2021-04-07
Don't remove "delphix-entire" when generating image diffs

prakashsurya:
Currently, if we generate an upgrade image diff from two images that
have the same delphix-entire package version, the resultant Aptly
repository will not contain any delphix-entire package in it. This is
problematic when that Aptly repository is later used with the script
used to generate an upgrade image from that repository, as the script
requires the repository to contain a "delphix-entire" package in it.
Thus, to avoid this issue, this change modifies the image diff logic to
ensure we never remove the "delphix-entire" package.

### DIFF
--- a/scripts/aptly-repo-from-image-diff.sh
+++ b/scripts/aptly-repo-from-image-diff.sh
@@ -100,14 +100,34 @@ aptly repo search image-a | xargs aptly repo copy image-a upgrade-repository ||
 
 #
 # Here we're performing step 3 from the comment above, but since the
-# "delphix-upgrade-verification" package is a bit different than most
-# other packages on a Delphix appliance, we need to handle that package
-# uniquely here. Specifically, we want to ensure this package is always
-# contained in the resultant aptly repository (even if the package is
-# the same within both image A and image B).
+# "delphix-upgrade-verification" and "delphix-entire" packages are a bit
+# different than most other packages on a Delphix appliance, we need to
+# handle these packages uniquely here. Specifically, we want to ensure these
+# packages are always contained in the resultant Aptly repository (even if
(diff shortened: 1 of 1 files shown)


scripts/aptly-repo-from-image-diff.sh
```diff
@@ -100,14 +100,15 @@ aptly repo search image-a | xargs aptly repo copy image-a upgrade-repository ||
...
-# "delphix-upgrade-verification" package is a bit different than most
-# other packages on a Delphix appliance, we need to handle that package
-# uniquely here. Specifically, we want to ensure this package is always
-# contained in the resultant aptly repository (even if the package is
-# the same within both image A and image B).
+# "delphix-upgrade-verification" and "delphix-entire" packages are a bit
+# different than most other packages on a Delphix appliance, we need to
+# handle these packages uniquely here. Specifically, we want to ensure
+# these packages are always contained in the resultant aptly repository
+# (even if the package is the same within both image A and image B).
```
sebroy, line 107: Code looks good, but it doesn't look like you've incorporated the changes you made in the `master` PR to this one.

sebroy (changes requested):

scripts/aptly-repo-from-image-diff.sh
prakashsurya: Correct, I've updated this with the commit from master which I just landed.

prakashsurya:
@sebroy @matt-skinner-delphix I think this is ready to go, whenever you have time to give this a look.

sebroy (approved):

matt-skinner-delphix (approved):

PR #540, opened 2021-04-16
DLPX-75347 old rootfs versions are not removed on upgrade in 6.0.8.0

prakashsurya:

### DIFF
--- a/upgrade/upgrade-scripts/rootfs-cleanup
+++ b/upgrade/upgrade-scripts/rootfs-cleanup
@@ -1,0 +1,170 @@
+#!/usr/bin/env python3
+
+import os
+import re
+import subprocess
+from typing import Any, List, Type
+
+IMAGE_PATH = os.path.dirname(os.path.realpath(__file__))
+
+
+#
+# Helper to compare two version strings using "dpkg --compare-versions".
+#
+def dpkgcmp(vers1: str, op: str, vers2: str) -> bool:
+    return subprocess.run(["dpkg", "--compare-versions", vers1, op,
+                           vers2]).returncode == 0
+
+
+#
+# Helper to obtain the "current-version" of a rootfs filesytem/snapshot.
+#
+def version(rootfs: str) -> str:
+    return str(
+        subprocess.check_output([
+            "zfs", "get", "-Hpo", "value", "com.delphix:current-version", rootfs
+        ]).decode("UTF-8").strip())
+
+
+#
+# Helper class used when sorting rootfs filesystems and snapshots. This allows
+# us to use the "sorted" builtin, to build a list of all rootfs filesystems and
+# snapshots on the system, sorted baseed on the "current-version" property.
+#
+class rootfscmp(object):
+
+    def __init__(self, rootfs: str, *args: List[Any]):
+        self.rootfs = rootfs
+
+    def __lt__(self, other: Type["rootfscmp"]) -> bool:
+        return dpkgcmp(version(self.rootfs), "lt", version(other.rootfs))
+
+    def __gt__(self, other: Type["rootfscmp"]) -> bool:
+        return dpkgcmp(version(self.rootfs), "gt", version(other.rootfs))
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, rootfscmp):
+            return NotImplemented
+        return dpkgcmp(version(self.rootfs), "eq", version(other.rootfs))
+
+    def __le__(self, other: Type["rootfscmp"]) -> bool:
+        return dpkgcmp(version(self.rootfs), "le", version(other.rootfs))
+
+    def __ge__(self, other: Type["rootfscmp"]) -> bool:
+        return dpkgcmp(version(self.rootfs), "ge", version(other.rootfs))
+
+    def __ne__(self, other: object) -> bool:
+        if not isinstance(other, rootfscmp):
+            return NotImplemented
+        return dpkgcmp(version(self.rootfs), "ne", version(other.rootfs))
+
+
+def main() -> None:
+    snapshots = subprocess.check_output([
+        "zfs", "list", "-r", "rpool/ROOT", "-d", "2", "-Ho", "name", "-t",
+        "snapshot"
+    ]).decode("UTF-8").splitlines()
+
+    for snap in snapshots:
+        name = snap.split("@")[1]
+
+        #
+        # We filter out any container-delphix snapshots, as those removed by
+        # the calls to "rootfs-container delete", on the relevant filesystems.
+        #
+        if re.search(r"^container-delphix.\w{7}$", name):
+            snapshots.remove(snap)
+            continue
+
+        #
+        # Out of an abundance of caution, verify the snapshot name matches
+        # our expectations.
+        #
+        if not re.search(r"^execute-upgrade.\w{7}$", name):
+            raise SystemExit("unexpected snapshot name: '{}'".format(name))
+
+    filesystems = subprocess.check_output([
+        "zfs", "list", "-r", "rpool/ROOT", "-d", "1", "-Ho", "name", "-t",
+        "filesystem"
+    ]).decode("UTF-8").splitlines()
+
+    #
+    # The "rpool/ROOT" dataset is not versioned and also not eligible
+    # for deletion, so we must remove it from the list before we attempt
+    # to sort and remove rootfs snapshots and filesystems.
+    #
+    filesystems.remove('rpool/ROOT')
+
+    current = subprocess.check_output(["get-appliance-version"
+                                      ]).decode("UTF-8").strip()
+
+    #
+    # Our policy is to keep two delphix versions at any given time; the
+    # currently running version, and the version used prior to the
+    # current one.
+    #
+    # By sorting the list of all filesystems and snapshots on the sytem,
+    # we can easily determine the highest two versions on the system,
+    # and skip those versions (to maintain the policy described above);
+    # this is based on the assumption that the versions of an appliance
+    # will always increase (or remain constant, in the case of rollback)
+    # after an upgrade (i.e. we never upgrade to a lower version).
+    #
+    for rootfs in sorted(filesystems + snapshots, key=rootfscmp)[:-2]:
+        #
+        # In the event of a rollback, we want to be careful to not
+        # delete the currently running version, as well as any versions
+        # that may have failed to upgrade resulting in the rollback
+        # event to begin with.
+        #
+        # Additionally, we want to keep all rootfs snapshots that map to
+        # hotfixes for the currently running version. This way, we don't
+        # inadvertently remove a rootfs snapshot that we may need for
+        # root-causing a given problem; at least, not until we upgraded
+        # off of the problematic delphix version.
+        #
+        # Thus, to account for these cases, we skip any version that's
+        # equal to or greater than what's currently running.
+        #
+        #
+        if dpkgcmp(version(rootfs), "ge", current):
+            break
+
+        #
+        # If a rollback event has occured, it's possible that we'll be
+        # unable to delete a snapshot or filesystem; e.g. if it's still
+        # being used as the "origin" of another filesystem. Rather than
+        # adding more complicated logic to handle this situation, we
+        # simply allow the commands to throw an error.
+        #
+        # If we fail to remove a snapshot or filesystem, this shouldn't
+        # negatively impact the appliance, apart from potentially
+        # consuming some more storage than strictly necessary.
+        #
+        # Further, as the appliance continues to be upgraded to new
+        # versions over time, these error conditions will subside; it
+        # just might take a couple upgrades to allow the previously
+        # "pinned" version to be removed and cleaned up.
+        #
+        # Lastly, it's possible for us to delete an entire rootfs
+        # container early in the iteration, and then attempt to delete
+        # some of that container's snapshots later in the iteration.
+        # Since deleting the container will either delete (or promote)
+        # all of that container's snapshots, the attempt (later) to
+        # delete the snapshots individually will silently fail.
+        #
+        if rootfs in filesystems:
+            subprocess.run([
+                "{}/rootfs-container".format(IMAGE_PATH), "delete",
+                rootfs.split('/')[2]
+            ])
+        elif rootfs in snapshots:
+            subprocess.run(["zfs", "destroy", "-r", rootfs])
+        else:
+            raise SystemExit(
+                "rootfs not found in filesystems or snapshots: '{}'".format(
+                    rootfs))
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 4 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/5076/)

upgrade/upgrade-scripts/rootfs-cleanup
```diff
@@ -0,0 +1,154 @@
...
+    # current one.
+    #
+    # By sorting the list of all filesystems and snapshots on the sytem,
+    # we can easily determine the highest two versions on the system,
+    # and skip those versions (to maintain the policy described above);
+    # this is based on the assumption that the versions of an appliance
+    # will always increase (or remain constant, in the case of rollback)
+    # after an upgrade (i.e. we never upgrade to a lower version).
+    #
+    for rootfs in sorted(filesystems + snapshots, key=rootfscmp)[:-2]:
```
sanbhara, line 112: wondering if it is not sufficient to sort the above zfs list commands by the "creation" field with "-s creation". Are there scenarios where the two orders (one by creation time and the other by associated delphix version) can be different?

upgrade/upgrade-scripts/rootfs-cleanup
```diff
@@ -0,0 +1,154 @@
...
+        # consuming some more storage than strictly necessary.
+        #
+        # Further, as the appliance continues to be upgraded to new
+        # versions over time, these error conditions will subside; it
+        # just might take a couple upgrades to allow the previously
+        # "pinned" version to be removed and cleaned up.
+        #
+        if rootfs in filesystems:
+            subprocess.run([
+                "{}/rootfs-container".format(IMAGE_PATH), "delete",
```
sanbhara, line 142: How does this code ensure that snapshots are deleted before their filesystems? It isn't clear to me how the sorted function works on the union of filesystems + snapshots lists.

upgrade/upgrade-scripts/rootfs-cleanup
prakashsurya: I thought about that, after I had implemented it this way. I think that would probably work too. You think I should try that way? I'm not sure if sorting by creation date would result it a cleaner implementation, but it might.. I can try it and see how I like it.

upgrade/upgrade-scripts/rootfs-cleanup
prakashsurya: The filesystem may be removed prior to the snapshots, sure, but is that a problem? When the filesystem is removed, all the snapshots under it will be removed as well.

upgrade/upgrade-scripts/rootfs-cleanup
prakashsurya: I think sorting on creation date would work.. do you prefer that approach? vs. sorting on delphix version?

If we do that, should we still maintain the logic to skip versions greater than the current? i.e. this logic:
```
113         #                                                                       
114         # In the event of a rollback, we want to be careful to not              
115         # delete the currently running version, as well as any versions         
116         # that may have failed to upgrade resulting in the rollback             
117         # event to begin with. Thus, to account for the exceptional case        
118         # of rollback, we simply skip any version that's equal to or            
119         # greater than what's currently running.                                
120         #                                                                       
121         if dpkgcmp(version(rootfs), "ge", current):                             
122             break   
```
This assumes versions always increase, and I feel like the point of using the creation date is to remove that assumption.. so we should also remove the assumption from this condition (i.e. remove it entirely), but then we need to ensure the cleanup (based on creation date) works properly for rollback..

prakashsurya: @sanbhara Looking at this a little closer, I don't think the creation date would work for the filesystems.. Since, for those, the creation date will be the date it was originally created, which may have been at a much older version/time than what it's currently at..

So, if we wanted to use the creation date, I think we could only do that for snapshots, and would have to revert back to versions for the filesystems.. Since we can't use it for both, I'm not sure it's worth the effort (and/or inconsistency and complexity) to use it for one..

What do you think?

sanbhara: Ok I see. Yes if we have to use the version property for one set of objects, we might as well do it for both sets of objects.

prakashsurya: OK, I'll go ahead and mark this comment as resolved.

upgrade/upgrade-scripts/rootfs-cleanup
sanbhara: Ok, I thought the delete of the filesystem will fail if it had dependencies but I understand now that it recursively destroys any snapshots on the filesystem. How about the "subprocess.run" command? When it attempts to delete the non-existent snapshot later on, will the error be suppressed allowing the loop to continue or will it exit with error?

prakashsurya: Yea, so if we did later in the iteration, try to delete a snapshot that was already removed, we'd likely throw an error on the call to "zfs destroy".. but since I opted to not check for errors (I'm using "run" instead of "check_call"), it shouldn't cause any issues.

We could try to ensure we delete the snapshot first, or perhaps check if it exists before we do the destroy, but IMO it's not worth the complexity. Let me know if you disagree.

sanbhara: I am fine with the present code. Can we document that the subprocess.run is being used in order to ignore errors, specifically this case where the order of deletion means that snapshots could already have gotten deleted?

prakashsurya: Yea, I can add some more comments about that case specifically.

upgrade/upgrade-scripts/upgrade
```diff
@@ -361,6 +361,41 @@ function rollback() {
...
+	# upgrade attempt; at which point the old hotfix file will no
+	# longer be representative of the new rootfs contents.
+	#
+	# If a rollback occured, then we do not want to remove this
+	# file, since the currently running rootfs will be the "old"
+	# version, and still should match the hotfix file. We detect if
+	# a rollback occurred by the presence of this field (which would
+	# be set in the upgrade properites file).
+	#
+	if [[ -z "$ROLLBACK_BASE_VERSION" ]]; then
```
sanbhara, line 391: I agree with this location for removing the hotfix file. It is critical to remove it similar to removing the upgrade.properties file once we are beyond the point of no return as far as upgrade is concerned. So lets keep it here in finalize().

upgrade/upgrade-scripts/rootfs-cleanup
```diff
@@ -0,0 +1,161 @@
...
+    filesystems = subprocess.check_output([
+        "zfs", "list", "-r", "rpool/ROOT", "-d", "1", "-Ho", "name", "-t",
+        "filesystem"
+    ]).decode("UTF-8").splitlines()
+
+    #
+    # We remove the first entry in the list, as that'll refer to the "rpool/ROOT"
+    # dataset, which is not versioned and also not eligible for deletion.
+    #
+    filesystems = filesystems[1:]
```
pzakha, line 95: nit: instead of relying on a specific ordering you can also do something like:
```
filesystems.remove('rpool/ROOT')
```

```
>>> filesystems = subprocess.check_output([
...     "zfs", "list", "-r", "rpool/ROOT", "-d", "1", "-Ho", "name", "-t",
...     "filesystem"
... ]).decode("UTF-8").splitlines()
>>> filesystems
['rpool/ROOT', 'rpool/ROOT/delphix.SBr1UZg']
>>> filesystems.remove('rpool/ROOT')
>>> filesystems
['rpool/ROOT/delphix.SBr1UZg']
```

upgrade/upgrade-scripts/rootfs-cleanup
```diff
@@ -0,0 +1,161 @@
...
+        # Lastly, it's possible for us to delete an entire rootfs
+        # container early in the iteration, and then attempt to delete
+        # some of that container's snapshots later in the iteration.
+        # Since deleting the container will either delete (or promote)
+        # all of that container's snapshots, the attempt (later) to
+        # delete the snapshots individually will silently fail.
+        #
+        if rootfs in filesystems:
+            subprocess.run([
+                "{}/rootfs-container".format(IMAGE_PATH), "delete",
```
pzakha, line 149: Would this command be logged somewhere easily accessible?
It might be useful to add a quick print that mentions that a container is being deleted.

pzakha (approved):

upgrade/upgrade-scripts/rootfs-cleanup
```diff
@@ -0,0 +1,161 @@
...
+
+
+#
+# Helper to obtain the "current-version" of a rootfs filesytem/snapshot.
+#
+def version(rootfs: str) -> str:
+    return str(
+        subprocess.check_output([
+            "zfs", "get", "-Hpo", "value", "com.delphix:current-version", rootfs
+        ]).decode("UTF-8").strip())
```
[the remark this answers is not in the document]
prakashsurya: What older version don't have that property?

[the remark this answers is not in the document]
pzakha: Would we have that property on something like 6.0.2.0?

upgrade/upgrade-scripts/rootfs-cleanup
prakashsurya: ah, right.. good point, I'll do that.

upgrade/upgrade-scripts/rootfs-cleanup
```diff
@@ -0,0 +1,161 @@
...
+
+
+#
+# Helper class used when sorting rootfs filesystems and snapshots. This allows
+# us to use the "sorted" builtin, to build a list of all rootfs filesystems and
+# snapshots on the system, sorted baseed on the "current-version" property.
+#
+class rootfscmp(object):
```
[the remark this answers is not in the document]
prakashsurya: I don't mind the deviation.. we don't yet have any requirements to maintain PEP8.

upgrade/upgrade-scripts/rootfs-cleanup
prakashsurya: it'll be captured in syslog.

upgrade/upgrade-scripts/rootfs-cleanup
```diff
@@ -0,0 +1,161 @@
...
+    #
+    for rootfs in sorted(filesystems + snapshots, key=rootfscmp)[:-2]:
+        #
+        # In the event of a rollback, we want to be careful to not
+        # delete the currently running version, as well as any versions
+        # that may have failed to upgrade resulting in the rollback
+        # event to begin with. Thus, to account for the exceptional case
+        # of rollback, we simply skip any version that's equal to or
+        # greater than what's currently running.
+        #
```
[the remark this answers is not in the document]
prakashsurya: Can you elaborate on what you mean?

You're saying, because we will have multiple rootfs containers/snapshots that will report the same version, and based on how we sort those same-versioned containers/snapshots, without this check we may attempt to remove the one we're currently using?

If so, I think that's valid, and could deserve a comment.

Additionally, I don't think we've discussed how we should handle the cleanup of old hotfix versions.. e.g. in the case that we apply multiple hotfix versions, all based on the same delphix version.. we could potentially accumulate many rootfs snapshots. Based on this logic, they should eventually get cleaned up, when all of them are "old enough", but I'm not sure if that's what we want.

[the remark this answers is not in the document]
pzakha: > You're saying, because we will have multiple rootfs containers/snapshots that will report the same version, and based on how we sort those same-versioned containers/snapshots, without this check we may attempt to remove the one we're currently using?

Yeah that's right.

upgrade/upgrade-scripts/rootfs-cleanup
```diff
@@ -0,0 +1,162 @@
...
+        if re.search("^container-delphix.\w{7}$", name):
+            snapshots.remove(snap)
+            continue
+
+        #
+        # Out of an abundance of caution, verify the snapshot name matches
+        # our expectations.
+        #
+        if not re.search("^execute-upgrade.\w{7}$", name):
```
pzakha, line 83: both flake8 and pylint seem to be complaining about `\w`:
flake8:
```
upgrade/upgrade-scripts/rootfs-cleanup:83:44: W605 invalid escape sequence '\w'
```
pylint:
```
upgrade/upgrade-scripts/rootfs-cleanup:83:17: W1401: Anomalous backslash in string: '\w'. String constant might be missing an r prefix. (anomalous-backslash-in-string)
```

Not sure but I think you may need to use `\\w` instead

upgrade/upgrade-scripts/rootfs-cleanup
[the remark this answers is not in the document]
prakashsurya: IIRC, all 6.0-based version will have it. Please correct me if I'm wrong.

upgrade/upgrade-scripts/rootfs-cleanup
prakashsurya: I've tested this to work as-is.

upgrade/upgrade-scripts/rootfs-cleanup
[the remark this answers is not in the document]
pzakha: Yeah looks like you are right, I was somehow under impression this wasn't introduced in 6.0.0.0

upgrade/upgrade-scripts/rootfs-cleanup
pzakha: Yeah looks like this syntax still works but is deprecated and will generate a syntax error in the future: see https://www.flake8rules.com/rules/W605.html

prakashsurya: Additionally, when I run `flake8` locally, I don't get that warning...

prakashsurya: OK, I've added `r` to those two strings.

pzakha: Yeah it seems to be introduced in python 3.6, perhaps your `flake8` is linked against python 2?

prakashsurya: not sure:
```
± python3 --version
Python 3.6.9

± python3 -m flake8 --version
3.9.1 (mccabe: 0.6.1, pycodestyle: 2.7.0, pyflakes: 2.3.1) CPython 3.6.9 on
Linux
```
I don't think it's important, though.. I went ahead and added `r`, which I think is correct, and shouldn't really change much w.r.t. testing or anything.

sanbhara (approved):
LGTM

PR #543, opened 2021-04-23
DLPX-75091 DLPX-75122 Make upgrade image prepare script hotfix aware

matt-skinner-delphix:
DLPX-75122 Use delphix version and hotfix in version.info in verify-jar script
DLPX-75091 Make upgrade image prepare script hotfix aware

The hotfix id needs to be passed into delphix-upgrade-verification/SUV so that SUV may operate on the correct version. Only works with corresponding [SUV changes](http://reviews.delphix.com/r/68590), otherwise verification will fail when an older SUV is given the hotfix value and does not know what to do with it.

Also refactored how version is determined. Instead of relying on a value in the filesytem path, the version is entirely sourced from version.info (and literally sourced via bash, pun intended). 

Tested with upgrade images and hotfix images with the corresponding SUV changes.

### DIFF
--- a/upgrade/upgrade-scripts/common.sh
+++ b/upgrade/upgrade-scripts/common.sh
@@ -142,7 +142,14 @@ function get_image_path() {
 }
 
 function get_image_version() {
-	basename "$(get_image_path)"
+#	basename "$(get_image_path)"
+  source_version_information
+  echo "$VERSION"
+}
+
+function get_image_hotfix() {
+  source_version_information
+  echo "$HOTFIX"
 }
 
 function get_mounted_rootfs_container_dataset() {
@@ -256,9 +263,6 @@ function source_version_information() {
 	local IMAGE_PATH="${IMAGE_PATH:-$(get_image_path)}"
 	[[ -n "$IMAGE_PATH" ]] || die "failed to determine image path"
 
-	local IMAGE_VERSION="${IMAGE_VERSION:-$(get_image_version)}"
-	[[ -n "$IMAGE_VERSION" ]] || die "failed to determine image version"
-
 	[[ -f "$IMAGE_PATH/version.info" ]] ||
 		die "image for version '$IMAGE_VERSION' missing version.info"
 	. "$IMAGE_PATH/version.info" ||
@@ -268,6 +272,7 @@ function source_version_information() {
 	[[ -n "$MINIMUM_VERSION" ]] || die "MINIMUM_VERSION is empty"
 	[[ -n "$MINIMUM_REBOOT_OPTIONAL_VERSION" ]] ||
 		die "MINIMUM_REBOOT_OPTIONAL_VERSION is empty"
+	IMAGE_VERSION="$VERSION"
 }
 
 function verify_upgrade_is_allowed() {
--- a/upgrade/upgrade-scripts/common.sh
+++ b/upgrade/upgrade-scripts/common.sh
@@ -141,17 +141,6 @@ function get_image_path() {
 	readlink -f "${BASH_SOURCE%/*}"
 }
 
-function get_image_version() {
-#	basename "$(get_image_path)"
-  source_version_information
-  echo "$VERSION"
-}
-
-function get_image_hotfix() {
-  source_version_information
-  echo "$HOTFIX"
-}
-
 function get_mounted_rootfs_container_dataset() {
 	dirname "$(zfs list -Hpo name /)"
 }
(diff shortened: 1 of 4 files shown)


prakashsurya (commented):
> Also refactored how version is determined. Instead of relying on a value in the filesytem path, the version is entirely sourced from version.info (and literally sourced via bash, pun intended).

What's the motivation for doing this refactor?

upgrade/upgrade-scripts/common.sh
```diff
@@ -142,7 +142,14 @@ function get_image_path() {
 }
 
 function get_image_version() {
-	basename "$(get_image_path)"
+#	basename "$(get_image_path)"
+  source_version_information
+  echo "$VERSION"
```
prakashsurya, line 147: Rather than modify this function, I think we might be able to remove it entirely.. e.g. by using `source_version_information` within the `verify-jar` script.. I think that's the only place where this function is currently needed, and we can replace that with `source_version_information`, and then use the `VERSION` and `HOTFIX` variables to pass into the upgrade-verify JAR.. What do you think?

upgrade/upgrade-scripts/common.sh
```diff
@@ -142,7 +142,14 @@ function get_image_path() {
...
+}
+
+function get_image_hotfix() {
```
prakashsurya, line 150: Where is this new function used?

upgrade/upgrade-scripts/common.sh
```diff
@@ -268,6 +272,7 @@ function source_version_information() {
 	[[ -n "$MINIMUM_VERSION" ]] || die "MINIMUM_VERSION is empty"
 	[[ -n "$MINIMUM_REBOOT_OPTIONAL_VERSION" ]] ||
 		die "MINIMUM_REBOOT_OPTIONAL_VERSION is empty"
+	IMAGE_VERSION="$VERSION"
```
prakashsurya, line 275: If we remove the `get_image_version` function, I think we can remove this as well.. I don't think we need the `IMAGE_VERSION` variable, do we?

upgrade/upgrade-scripts/common.sh
matt-skinner-delphix: `get_image_version` is used in the `upgrade` script as well, but it is used identically to `verify-jar` script. Sure, it is a simplification that can improve readability.

upgrade/upgrade-scripts/common.sh
matt-skinner-delphix: Nowhere, was left over from trying mirror existing patterns. Removed.

upgrade/upgrade-scripts/common.sh
matt-skinner-delphix: I was attempting to minimize changes to avoid introducing simple bugs while working on bigger issues. However now that these bigger issues are not as much a concern, I certainly agree that cleanup work makes some sense.

upgrade/upgrade-scripts/common.sh
prakashsurya: it's used in `upgrade`, but the ways it's used is dead code..
```
 23 IMAGE_VERSION=$(get_image_version)                                              
 24 [[ -n "$IMAGE_VERSION" ]] || die "failed to determine image version" 
 ```
 while it's used here, `IMAGE_VERSION` is never used.. so I'm pretty sure we can simply remove these two lines and everything would still work.

**Review by sanbhara — APPROVED** [review]

File: upgrade/upgrade-scripts/verify-jar [hunk 4]
```diff
@@ -18,11 +18,16 @@
...
-IMAGE_VERSION=$(get_image_version)
-[[ -n "$IMAGE_VERSION" ]] || die "failed to determine image version"
+# We source the version information to bring in values from
+# the version.info file, such as
+#   - VERSION : the appliance version
+#   - HOTFIX : the hotfix version if any
+
+source_version_information
+
+[[ -n "$VERSION" ]] || die "failed to determine image version"
```
prakashsurya, line 28: This is unnecessary, right? If so, how do you feel about removing it? IIRC, this same check is already done by the `source_version_information` function.

prakashsurya (approved):
LGTM.

upgrade/upgrade-scripts/upgrade
```diff
@@ -20,8 +20,6 @@
 IMAGE_PATH=$(get_image_path)
 [[ -n "$IMAGE_PATH" ]] || die "failed to determine image path"
 
-IMAGE_VERSION=$(get_image_version)
-[[ -n "$IMAGE_VERSION" ]] || die "failed to determine image version"
```
prakashsurya, line 23: it looks like shfmt only wants a single empty line here; see [also](https://github.com/delphix/appliance-build/pull/543/checks?check_run_id=2496161744).

prakashsurya (approved):

prakashsurya (approved):

PR #551, opened 2021-06-08
Add "awscli" to dcenter variant

prakashsurya:
In order to make it easier to download upgrade images, this change adds
the awscli package to our dcenter images.

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
@@ -19,6 +19,7 @@
 - apt:
     name:
       - adoptopenjdk-java8-jdk
+      - awscli
       - bind9
       - dnsutils
       - git


basil (commented):
I think we'd rather prefer to use the latest AWS CLI in a virtual environment rather than relying on the out-of-date Ubuntu package. For the same reason, we don't install the `awscli` package on Jenkins agents.

prakashsurya:
This doesn't prevent the use of a virtualenv provided `aws` command within our automation. Currently, without this package pre-installed, it means we don't have a pre-installed method of downloading upgrade images. When trying to download images outside of automation, is it still preferred to have the user create a virtual environment?

basil:
I think so, because Python packages installed through `apt(1)` are generally out-of-date, and changes to the list of packages requires heavyweight `appliance-build` changes and backporting. The fact that we rely on `apt(1)` to install (old versions of) Tenacity and Marshmallow is bad enough (and requires workarounds in `dcenter-gate`), so I'd rather keep the problem from getting worse. The goal I have is to install zero Python packages through `apt(1)` other than `python3-venv` itself, from which (recent) versions of all other packages can be installed. Then maintaining and updating those versions can be handled more easily using `requirements.txt`, lock files, Dependabot, etc.

PR #548, opened 2021-05-11
DLPX-64201 determine how to remove old linux kernels

prakashsurya:

PR #554, opened 2021-06-09
Revert "DLPX-64201 determine how to remove old linux kernels (#548)"

prakashsurya:
This reverts commit be1ecc9701ca18b05dec0ec6b6016bb39bc79b6c.

### DIFF
--- a/upgrade/upgrade-scripts/execute
+++ b/upgrade/upgrade-scripts/execute
@@ -242,32 +242,19 @@ dpkg-query -Wf '${Package}\n' | xargs apt-mark auto ||
 
 #
 # While we marked all packages to "auto" above, so they'd be removed
-# when they're no longer required, we treat kernel packages slightly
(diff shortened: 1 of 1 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/5481/)

pzakha (approved):

prakashsurya:
After discussing with @pzakha, I think the original fix for DLPX-64201 might not work correctly in the case of a customer running a kernel version from a 6.0.5 or earlier release. Thus, rather than attempting to fix the issue late in the release cycle, I propose we remove this logic from 6.0.9, and look to re-integrate for 6.0.10 with a better implementation.

sebroy (approved):

PR #557, opened 2021-06-15
CP-5334 Add ufw package to appliance-build

jeffngo:
Context
The Uncomplicated Firewall (ufw) utility is a frontend for iptables which is well-suited for host-based firewalls. `ufw` aims to provide an easy to use interface for those unfamiliar with firewall concepts, while at the same time simplifying iptables commands to help an admin who knows what he or she is doing.

Problem
The AppData team wants to add testing around the docker plugin runtime without internet access. The build should target internal qa and dev builds, but should be excluded from customer engines.

Solution
The ufw utility should allow us to configure the engine to allow outgoing connections to SSH, DNS and specific source and target host environments while also restricting internet access elsewhere.

Testing
git ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/5514/
I cloned from the snapshot of the aws image and found that the ufw pkg was installed on the delphix engine. The default state of ufw is inactive, which is expected.
```
delphix@ip-10-110-194-195:~$ sudo systemctl status ufw
● ufw.service - Uncomplicated firewall
   Loaded: loaded (/lib/systemd/system/ufw.service; enabled; vendor preset: enabled)
   Active: active (exited) since Tue 2021-06-15 16:30:10 UTC; 53min ago
     Docs: man:ufw(8)
 Main PID: 486 (code=exited, status=0/SUCCESS)
    Tasks: 0 (limit: 4915)
   CGroup: /system.slice/ufw.service

delphix@ip-10-110-194-195:~$ sudo ufw status
Status: inactive
```

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.qa-internal/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.qa-internal/tasks/main.yml
@@ -1,5 +1,5 @@
 #
-# Copyright 2018 Delphix
+# Copyright 2018, 2021 Delphix
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -22,6 +22,7 @@
     name:
       - nftables
       - snmptrapd
+      - ufw
     state: present
   register: result
   until: result is not failed


pzakha (approved):

prakashsurya:
I'm not necessarily against the addition of `ufw`, but from the explanation, it sounds like the needs are similar to why we're already adding the `nftables` package; is that true? Just curious if there's overlap between the `nftables` and `ufw` packages, and/or if we could get away with installing only one or the other (e.g. do we really need both)?

jeffngo:
> I'm not necessarily against the addition of `ufw`, but from the explanation, it sounds like the needs are similar to why we're already adding the `nftables` package; is that true? Just curious if there's overlap between the `nftables` and `ufw` packages, and/or if we could get away with installing only one or the other (e.g. do we really need both)?

Hmm yeah, you're right. We originally went with `ufw` because it was simple and effective for our use case, but since `nftables` is already installed, we could technically convert our commands from `ufw` to `nft` to avoid installing a new package.

prakashsurya:
OK, so I'm not necessarily against adding the `ufw` package (as I mentioned), but I do think the default stance should be to avoid creating more differences between the QA image, and the image we ship to customers (that's the whole reason the QA image exists).

So in this case, I'll tentatively approve to unblock this, but I do think it's strictly a worse solution than either:

a) replacing `nftables` with `ufw` (i.e. converting our current use of `nft` to `ufw`), or
b) replacing `ufw` with `nftables` (i.e. closing this, and converting to `nft` as you mention)

since, as is, this will add some more drift between our QA image and our external images; not a _huge_ deal, but I do think we should be critical each time we add more drift between the two.

prakashsurya (approved):

PR #563, opened 2021-06-24
DLPX-76328 upgrade script obtains hotfix version incorrectly

prakashsurya:
In commit e7dd85e2 we introduced the `get_hotfix_version` function for
obtaining the hotfix version of the currently running root filesystem.
The problem is that in c4357a4e we added logic that intended to use this
function, but instead tries to call the `get_current_hotfix` function,
which doesn't exist. This leads us to never retreiving the hotfix
version correctly via the `rollback` subcommand of the `upgrade` script.

### DIFF
--- a/upgrade/upgrade-scripts/upgrade
+++ b/upgrade/upgrade-scripts/upgrade
@@ -317,7 +317,7 @@ function rollback() {
 	ROLLBACK_BASE_VERSION="$(get_current_version)"
 	[[ -n "$ROLLBACK_BASE_VERSION" ]] ||
 		die "unable to determine current appliance version"
-	ROLLBACK_BASE_HOTFIX="$(get_current_hotfix)"
+	ROLLBACK_BASE_HOTFIX="$(get_hotfix_version)"
 
 	set_upgrade_property "ROLLBACK_BASE_CONTAINER" "$ROLLBACK_BASE_CONTAINER" ||
 		die "failed setting 'ROLLBACK_BASE_CONTAINER' property"


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/5570/)

sanbhara (commented):
Is it possible to add a manual test of this hotfix rollback codepath? Btw, are there new upgrade blackbox tests planned/implemented as part of the hotfix project?

sanbhara (approved):

prakashsurya:
> Is it possible to add a manual test of this hotfix rollback codepath?

I think we can use the rollback automation, but using a hotfix image.. e.g. 6.0.9 -> 6.0.9 + HF -> back to 6.0.9.. right? I think it'd be good to do that.

> Btw, are there new upgrade blackbox tests planned/implemented as part of the hotfix project?

I don't think we have any new tests planned for the hotfix project, but I think we can use all of the same tests that we use for normal upgrade, but using a hotfix image instead of a regular upgrade image.

sanbhara:
> I think we can use the rollback automation, but using a hotfix image.. e.g. 6.0.9 -> 6.0.9 + HF -> back to 6.0.9.. right? I think it'd be good to do that.

Yup that should be sufficient. In this scenario would the 6.0.9+HF be a sparse image or a full image? Do both support rollback? If so, should we test both?

prakashsurya:
> In this scenario would the 6.0.9+HF be a sparse image or a full image? Do both support rollback? If so, should we test both?

combined vs. sparse image doesn't really matter for this; both will exercise the same exact code paths. We can do a "full" upgrade using both "combined" and "sparse" hotfix images, so yea, we support rollback for both types.

IMO, since they both exercise the same code paths, I think it's sufficient to only test the "combined" hotfix, but I'm not opposed to testing both (I think it'll duplicate the tests we run for minimal benefit, but that's fine with me).

sumedhbala-delphix (approved):

PR #565, opened 2021-06-28
DLPX-76376 rootfs property "com.delphix:hotfix-version" incorrect after rollback of hotfix

prakashsurya:


prakashsurya:
- `git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/5631/)
- `git-hotfix-build` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/hotfix-build/136/)

upgrade/upgrade-scripts/common.sh
```diff
@@ -233,6 +233,13 @@ function copy_optional_dataset_property() {
 	#
 	[[ -n "$PROP_VALUE" && "$PROP_VALUE" != "-" ]] || return
 
+	#
+	# Additionally, we only want to copy the dataset property when
+	# it's a local value, rather than a potentially inherited value.
+	#
+	PROP_SOURCE=$(zfs get -Hpo source "$PROP_NAME" "$SRC_DATASET")
```
sanbhara, line 240: I saw a suggestion by Pavel that "-s local" can be used on line 226. Have you taken a call which to use?

prakashsurya: Yea, I thought about that.. I wouldn't be opposed to using `-s local`.. I left what I have here for two main reasons:

1. I had already written it this way, before I learned about `-s local`.. and..
2. I thought this might make it more clear what were doing.. i.e. more verbose, and thus less likely for a reader to miss..

With that said, using `-s local` would be a bit more terse, and perhaps more elegant..

So, I'm good with either approach.. If you'd prefer using `-s local` I can make that change.. what do you think?

sanbhara: I also think adding a comment to point out that we are limiting the scope while using the -s switch might be cleaner.

prakashsurya: OK, sounds good to me. I'll make that change.

sanbhara (approved):

sumedhbala-delphix (approved):

PR #569, opened 2021-07-01
TOOL-11824 Harden against mistakes when running via Jenkins

prakashsurya:
This change adds logic to cause a failure when certain environment
variables are missing, and the build is being run via Jenkins.

When environment variables are missing, we currently will provide
default values, such that the build doesn't fail. These default values
generally are not sufficient for anything besides allowing the build to
complete (e.g. the resultant upgrade images may fail upgrade). Further,
the main motivation for generating these default values, is to make it
easier to run the build manually, and aid developer iteration.

Thus, when it's clear that the build is being run via Jenkins, where
these missing environment variables more likely point to an error or
broken automation, than an intentional omission of these variables, it's
better to report the error than set and use a default value. This way,
we can more quickly detect and address the problem, than silently
produce (potentially) incorrect or broken artifacts.

### DIFF
--- a/scripts/upgrade-image-from-aptly-repo.sh
+++ b/scripts/upgrade-image-from-aptly-repo.sh
@@ -122,6 +122,8 @@ if [[ -n "$DELPHIX_UPGRADE_MINIMUM_VERSION" ]]; then
 		"s/@@MINIMUM_VERSION@@/$DELPHIX_UPGRADE_MINIMUM_VERSION/" \
 		version.info ||
 		die "failed to set MINIMUM_VERSION in version.info"
+elif [[ -n "$JENKINS_URL" ]]; then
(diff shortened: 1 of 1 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/5680/)

pzakha (approved):
Should we create a JIRA for this so we don't forget to backport it?

pzakha:
Btw, thanks for doing this work!

prakashsurya:
> Should we create a JIRA for this so we don't forget to backport it?

Sure, done; https://jira.delphix.com/browse/TOOL-11824

sanbhara (approved):

PR #574, opened 2021-07-12
DLPX-76293 Copy Hotfix Metadata Generation to appliance-build

matt-skinner-delphix:
As part of the ongoing effort to move build elements out of the app-stack this work copies the generation script into appliance-build. Additional changes were made to the script to remove dead code and accept parameters instead of hard coded values. One notable change is that the hotfixmetadat will be generated on trunk/master branches. 

The hotfix metadata generation is made a dependency of the main appliance build and so will fail the build if the metadata is unable to be generated. Friendly error messages were also added.

# Testing
[Automated](http://collins.d.delphix.com:36640/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/
)

# Manual
See the [corresponding code review](http://reviews.delphix.com/r/70951/) for the devops piece for examples.


scripts/generate-hotfix-metadata/generate-hotfix-metadata.py
```diff
@@ -0,0 +1,332 @@
+#!/usr/bin/python
```
pzakha, line 1: nit: this should show python3

scripts/generate-hotfix-metadata/generate-hotfix-metadata.py
```diff
@@ -0,0 +1,332 @@
...
+  JIRA_PASSWORD = os.getenv('JIRA_PASSWORD', None)
+
+  # Gradle will pass in the string "null" if the values are not set
+  errors = []
+  if not JIRA_URL or JIRA_URL == 'null':
+    errors.append("Environment variable JIRA_URL not set.")
+  if not JIRA_USER or JIRA_USER == 'null':
+    errors.append("Environment variable JIRA_USER not set.")
+  if not JIRA_PASSWORD or JIRA_PASSWORD == 'null':
+    errors.append("Environment variable JIRA_PASSWORD not set.")
```
pzakha, line 242: Is appliance-build going to fail when run manually if those environment variables are not set?

We are trying to make it easier for users to run appliance-build manually, so they can do so without having to set a bunch of environment variables; the build skips the steps that requires those variables to be set. At the same time we want to make sure that all of those variables are passed when run by automation to avoid silent regressions (see for instance https://github.com/delphix/appliance-build/pull/570). Perhaps you could add a similar check to the calling `run.sh` script (it could generate an empty hotfix file so that generation of delphix-entire doesn't fail)?

scripts/generate-hotfix-metadata/run.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
pzakha, line 1: running with `-e` will ensure that we do not silently fail in this script

pzakha (commented):
Thanks for moving this to appliance-build Matt!

Can you provide a summary of the changes you did to the `genhotfixmetadata` script? Note that I've already applied your python2 -> python3 changes in http://reviews.delphix.com/r/70536 with some minor modifications.

live-build/build.gradle
```diff
@@ -52,6 +54,51 @@ def runTypesAndSuffixes = [[upgradeArtifactsRunType, "UpgradeArtifacts"],
                            [vmArtifactsRunType, "VmArtifacts"],
                            [allRunType, ""]]
 
+
+
+task generateHotfixMetaData(type: Exec) {
```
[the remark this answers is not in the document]
matt-skinner-delphix: A separate gradle task provides:

1. Checking parameters before running other build steps, failing the build early rather than spending time to only then fail later.
2. Allows shorter and easier dev-test cycles via gradle.
3. `generateHotfixMetadata` can take a while to run and so is better to use gradle, which support not re-running tasks that don't have changed inputs.

The intent is to make the hotfix metadata an input of the build rather than something the build generates and uses.

scripts/generate-hotfix-metadata/generate-hotfix-metadata.py
matt-skinner-delphix: A build without hotfixmetadata is not a valid build as I would assume other cases where steps are skipped due to missing inputs. Do we provide a means to tell the user what they are missing by not providing these addtional environment variables? For example, if we do not provide the environment variables for querying Jira to get hotfix metadata, then the resulting image should not be used for anything related to hotfixes.

What about providing flags to bypass features that require these environment variables. When these bypasses are not provided and the environment variables are not present we must then require the error message explain the flags needed to bypass reading the environment variable. This approach makes it is clear to users of appliance-build that they are intentionally running appliance-build in such a way that it may generate an image that may not work as intended, such as having missing features in the product. 

I am concerned that if the default behavior is to build an image that has missing components without knowing what is missing, there are going to be issues with using the image because incorrect assumptions are going to be made. I also don't think reading through the build log is sufficient information to convey to the user what is skipped unless we also have an annoying summary message at the end saying what was skipped.

matt-skinner-delphix:
The python script was modified:

- include additional error checking
- moved logic from gradle into python code
- parse the version passed in so that extended version strings are supported (e.g. `6.0.9.0-something-else`)
- removed hard-coded password, username, jira url

scripts/generate-hotfix-metadata/generate-hotfix-metadata.py
pzakha: Yeah we have the same issue with a bunch of other environment variables, for instance the signature, the delphix appliance version, etc.

I think it would be reasonable to check for all the environment variables in the beginning of the build and issue clear warnings if they are not set. However I believe it would be a regression to fail the build if some of those variables are not set when running the build manually. Note that if we use similar checks to https://github.com/delphix/appliance-build/pull/570 we would not be able to accidentally produce a broken production image.

prakashsurya: FWIW, if appliance-build is currently run outside of Jenkins, and the user does not properly construct their environment to contain the all of the "special" envirionment variables, then the image will not work when used with the product's upgrade UI; this is already the case today.

So while I understand the sentiment, that we should try to produce an image that will work properly, that goal conflicts with the other goal of being able to run the build outside of Jenkins, and not requiring the user to need intimate knowledge about how to set up the environment properly (i.e. the build should work with the default environment).

IMO, if the JIRA environment variables are not present, and we're not building via Jenkins, than I think it's reasonable to produce an image that might not be entirely "correct", such that it doesn't have the proper hotfix metadata. When run outside of Jenkins, the priority should be to allow the build to succeed, while performing as much of the normal build logic as we can, without hindering the user-interface of the build for that manual invocation (i.e. without forcing the user to specify lots of environment variables properly). Further, it's less of a priority to ensure the manually generated artifacts match the Jenkins produced artifacts.

I think of a build run outside of Jenkins more of a best-effort type of thing, rather than necessarily having to product artifacts that will work properly when used with the product.

> there are going to be issues with using the image because incorrect assumptions are going to be made.

The artifacts that are generated outside of Jenkins, shouldn't really be used for anything beyond manual testing and/or developer iteration, and the developer that built the artifacts will need to know what is an appropriate test with the manually generated artifacts, and what is not appropriate. So, I think the concern is valid, but we've already adopted the notion that it's up to the developer to ensure they don't use the artifacts incorrectly, so I don't think we should be changing that approach in the context of this PR.

For example, if we did want to change that approach, we should do it holistically, and look at other areas of the build that will not (currently) work "correctly" when missing environment variables, which I (also) don't think is appropriate for this PR.

live-build/build.gradle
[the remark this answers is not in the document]
prakashsurya: > The intent is to make the hotfix metadata an input of the build rather than something the build generates and uses.

Hm, this has me thinking.. would it be better, from a build architecture stand-point, to move the script that builds the metadata file into the devops-gate.. such that we could call it via a new Jenkins job, and have it store it's output in S3.. such that appliance-build would just consume the file from S3? rather than having appliance-build run the script, and generate it for each build?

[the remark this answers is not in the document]
prakashsurya: Further, we likely don't really want this script to be versioned alongside the product.. and since appliance-build is currently versioned based on the product version, that already makes me think that having this in the appliance-build repository is not the ideal location..

scripts/generate-hotfix-metadata/generate-hotfix-metadata.py
matt-skinner-delphix: I am suggesting is to either force the user to supply the correct parameters to skip parts of the build  (e.g. "without flags") or to provide a summary of what was not included in the build and what environment variables would be required to enable the feature, similar to how many open source software reports what features are missing at the end of a compile or configure (E.g. `./configure` or `make`). 

While we do not want to require users to have detailed knowledge of which environment variables to set to run a build, we might want to make it a little easier for a developers to become more knowledgable without having to hunt through the code or the devops code just to run a build a certain way. 

Of course all of this is out of scope for this review.

live-build/build.gradle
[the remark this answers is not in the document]
matt-skinner-delphix: We also seem to run appliance build for multiple variants as separate tasks, which means the hotfix metadata file is generated multiple times. Although the variants are done in parallel, it adds a few minutes to each build.

scripts/generate-hotfix-metadata/generate-hotfix-metadata.py
prakashsurya: Yea, I'm not against making things more clear as you suggest. I'd expect that to be a decent amount of effort, so doing this in another PR makes sense (if it's going to be done, at all).

live-build/build.gradle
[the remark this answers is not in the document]
prakashsurya: oh, yea.. there's that, too.

how do you feel about potentially moving this script into a separate job, like I was thinking?

scripts/generate-hotfix-metadata/generate-hotfix-metadata.py
pzakha: > I am suggesting is to either force the user to supply the correct parameters to skip parts of the build (e.g. "without flags") or to provide a summary of what was not included in the build and what environment variables would be required to enable the feature, similar to how many open source software reports what features are missing at the end of a compile or configure

That's reasonable and would probably be helpful. I agree with Prakash that it would require a bit of extra work though, so I'm OK deferring this to another PR.

johngallagher-dlpx (approved):

live-build/config/hooks/configuration/81-upgrade-repository.binary
```diff
@@ -80,6 +80,20 @@ EOF
 #
 rename 's/\%3a/:/g' binary/packages/*.deb
 
+#
+# Copy over the generated hotfix metadata.
+#
+if [[ -z "$HOTFIX_METDATA" ]]; then
```
prakashsurya, line 86: Would it make sense to name this consistently with other environment variables that point to S3 artifacts.. e.g. something like `AWS_S3_URI_HOTFIX_METADATA`? we seem to use the `AWS_S3_URI_` prefix for other variables..

live-build/config/hooks/configuration/template.ctl
```diff
@@ -6,7 +6,7 @@ Standards-Version: 3.9.2
 Package: delphix-entire-@@PLATFORM@@
 Provides: delphix-entire
 Version: @@VERSION@@
-Extra-Files: packages.list, variant
+Extra-Files: packages.list, variant, hotfix_metadata
```
prakashsurya, line 9: Can we come up with a filename that better describes the contents of the file.. e.g. perhaps `fixed-hotfixes.list` or perhaps just `hotfixes.list`? since the contents is a list of all the hotfixes that have been fixed in this upgrade image that is being generated, and the `.list` suffix would be consistent with the `packages.list` file.

live-build/config/hooks/configuration/81-upgrade-repository.binary
```diff
@@ -80,6 +80,20 @@ EOF
...
+	touch hotfix_metadata
+else
+	aws s3 cp --only-show-errors "$HOTFIX_METDATA" hotfix_metadata
+fi
+
+if [[ ! -f "hotfix_metadata" ]]; then
```
prakashsurya, line 92: is this check necessary? e.g. I'm wondering that if the file doesn't exist, would the package fail to build? i.e. since the package requires that file, which wouldn't exist.. I'm not necessarily against an explicit check, it just seems a bit redundant, and inconsistent with other files required by the package (e.g. we don't do this for `packages.list`)..

PR #580, opened 2021-07-29
DLPX-76769 [Backport of DLPX-76694 to 6.0.10.0] Failure to build DelphixConnector on Delphix Engine, causing gui_sanity testing to fail

pzakha:
Clean cherry-pick of #578 

## Testing
ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/5808/

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.virtualization-development/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.virtualization-development/tasks/main.yml
@@ -15,6 +15,19 @@
 #
 
 ---
+#
+# pkg-config is necessary for building the DelphixConnector, which is built
+# when running "ant all".
+#
+- apt:
+    name:
+      - pkg-config
+    state: present
+  register: result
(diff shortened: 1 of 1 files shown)


prakashsurya (commented):
I believe we have all the app-gate build dependencies in the QA variant ansible role, should this be included in that list, rather than here?

pzakha:
> I believe we have all the app-gate build dependencies in the QA variant ansible role, should this be included in that list, rather than here?

I wasn't aware of that, but yeah reading the comment in appliance-build.qa-internal that would probably make sense. I could push a follow-up to the parent bug on master and that backport that instead.

prakashsurya (approved):
ah, right.. this is a backport, so I think we should keep it consistent with the original master push.

pzakha:
I'll push this backport, then open another review to move this to internal-qa

rasantel (approved):

PR #576, opened 2021-07-27
TOOL-11707 Install Java 11 on DCoL hosts

blim747:
DCoL hosts are Jenkins agents and therefore has a Java 8 JDK installed (cf. [`appliance-build.dcenter`](https://github.com/delphix/appliance-build/blob/be1ecc9701ca18b05dec0ec6b6016bb39bc79b6c/live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml#L21)). In the coming months, the Jenkins project will start to recommend users move off Java 8 and to Java 11. With the move to Java 11, we need to also add a Java 11 JDK.

Note: `adoptopenjdk-java11-jdk` is technically before `adoptopenjdk-java8-jdk` alphabetically

PR #585, opened 2021-08-06
Want variant-specific package lists for `live-build(7)`

### LINKED ISSUE #582: Want variant-specific package lists for Live Build
There currently exist two ways to install packages:

- Add a package to a Live Build package list in `live-build/config/package-lists`, where it will be processed by Live Build in an environment where `/proc` is mounted. Such package lists apply to all variants.
- Add a package to a role referenced by an Ansible playbook in `live-build/variants/${VARIANT}/ansible`, where it will be processed by Ansible in an environment where `/proc` is not mounted. Such Ansible playbooks apply to specific variants.

As discovered in #576, some packages (notably `ca-certificates-java`) require `/proc` to be mounted during installation, and those packages must be installed in some variants (notably `internal-dcenter`, to provide Java 11 support for Jenkins agents) and not others (notably `internal-dev`, `internal-qa`, and `external-standard`, which are designed to support products that use Java 8 and are incompatible with Java 11).

No infrastructure currently exists to support this use case. The solution is to enhance `appliance-build`'s build to support variant-specific package lists (just as it already supports variant-specific Ansible playbooks) and plumb those through to Live Build alongside the existing package lists that apply to all variants.

A possible design would be to introduce variant-specific package lists in e.g. `live-build/variants/${VARIANT}/package-lists`, updating `scripts/run-live-build.sh` to merge them with the existing package lists that apply to all variants and plumb through the result to Live Build.

basil:
Supersedes #576. Closes #582. Introduces support for variant-specific package lists in `live-build/variants/${APPLIANCE_VARIANT}/package-lists`, updating `scripts/run-live-build.sh` to merge them with the existing package lists that apply to all variants before plumbing through the result to `live-build(7)`. This allows us to move the list of packages for the `internal-dcenter` variant from the `appliance-build.dcenter` Ansible role to a new `dcenter.list.chroot` package list, which contains all of the original packages as well as `openjdk-11-jdk-headless`. These packages are installed in a context where `/proc` is mounted, which is necessary for `ca-certificates-java` (a dependency of `openjdk-11-jdk-headless`). We continue to use Java 8 as the default Java runtime to allow us to gradually migrate to Java 11. When we have fully migrated to Java 11, we can change the default Java runtime from Java 8 to Java 11.

### Implementation

`live-build(7)` uses a [flat](https://gitlab.com/debian-live/live-build/-/blob/489a09ba92328cef846598a15329e9ff91519e3c/scripts/build/binary_package-lists#L87-L88) [directory structure](https://gitlab.com/debian-live/live-build/-/blob/489a09ba92328cef846598a15329e9ff91519e3c/scripts/build/chroot_package-lists#L80-L82) for package lists, but there could still be problems if a variant has a package list of the same name as one that applies to all variants. To avoid such problems, we require all package lists to have unique names and fail fast if a variant has a package list of the same name as one that applies to all variants.

### Future work

Several of the existing Ansible roles install variant-specific packages and could be refactored to use this new subsystem if desired. For example, [this code in the `appliance-build.unittest-internal` role](https://github.com/delphix/appliance-build/blob/7cc8789445b73594b2265b70b86886b7ec262c5e/live-build/misc/ansible-roles/appliance-build.unittest-internal/tasks/main.yml#L18-L29) could easily be turned into a new package list for the `internal-unittest` variant. Others have suggested that the infrastructure introduced in this PR may overlap with the fix for [DLPX-76872](https://jira.delphix.com/browse/DLPX-76872).

#### Functional testing

Logged into an `internal-dcenter` VM with these changes and verified that Java 11 was installed:

```
$ /usr/lib/jvm/java-11-openjdk-amd64/bin/java -version
openjdk version "11.0.11" 2021-04-20
OpenJDK Runtime Environment (build 11.0.11+9-Ubuntu-0ubuntu2.18.04)
OpenJDK 64-Bit Server VM (build 11.0.11+9-Ubuntu-0ubuntu2.18.04, mixed mode, sharing)
```

Verified that Java 8 remained the default:

```
$ readlink -f /usr/bin/java
/usr/lib/jvm/adoptopenjdk-java8-jdk-amd64/bin/java
$ java -version
openjdk version "1.8.0_282"
OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_282-b08)
OpenJDK 64-Bit Server VM (AdoptOpenJDK)(build 25.282-b08, mixed mode)
```

Finally, set up the VM as a Jenkins agent using the Java 11 runtime and verified that Jenkins could execute builds on that agent.

#### Negative testing

Intentionally introduced a naming conflict and verified that [the build failed as expected](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-stage1/job/pre-push/10436/consoleFull):

```
+ echo 'Duplicate package list: delphix-platform.list.chroot'
Duplicate package list: delphix-platform.list.chroot
+ exit 1

FAILURE: Build failed with an exception.
```

#### Automated testing

- `git ab-pre-push -v internal-dcenter -p esx`: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/5883/
- `git ab-pre-push -v internal-qa -p aws`: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/5884/

### DIFF
--- a/scripts/run-live-build.sh
+++ b/scripts/run-live-build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright 2018, 2020 Delphix
+# Copyright 2018, 2021 Delphix
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -93,13 +93,24 @@ if [[ "$RUN_TYPE" == "$ALL_RUN_TYPE" || "$RUN_TYPE" == "$VM_RUN_TYPE" ]]; then
 	cp -r "$TOP/live-build/config/hooks/$VM_RUN_TYPE/." "$build_dir/config/hooks"
 fi
 
+sed "s/@@PLATFORM@@/$APPLIANCE_PLATFORM/" \
+	<"$build_dir/config/package-lists/delphix-platform.list.chroot.in" \
+	>"$build_dir/config/package-lists/delphix-platform.list.chroot"
+
+if [[ -d "$TOP/live-build/variants/$APPLIANCE_VARIANT/package-lists" ]]; then
+	for list in "$TOP/live-build/variants/$APPLIANCE_VARIANT/package-lists/"*; do
+		[[ -f $list ]] || continue
+		if [[ -f "$build_dir/config/package-lists/$(basename "$list")" ]]; then
+			echo "Duplicate package list: $(basename "$list")" >&2
+			exit 1
+		fi
+		cp "$list" "$build_dir/config/package-lists"
+	done
+fi
+
 cp -r "$TOP/live-build/variants/$APPLIANCE_VARIANT/ansible" "$build_dir"
 
 cd "$build_dir"
-
-sed "s/@@PLATFORM@@/$APPLIANCE_PLATFORM/" \
-	<config/package-lists/delphix-platform.list.chroot.in \
-	>config/package-lists/delphix-platform.list.chroot
 
 #
 # The ancillary repository contains all of the first-party Delphix
--- a/live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
@@ -15,45 +15,16 @@
 #
 
 ---
-
-- apt:
-    name:
-      - adoptopenjdk-java8-jdk
-      - bind9
-      - dnsutils
-      - git
-      - isc-dhcp-server
-      - libldap2-dev
-      - libsasl2-dev
-      - nfs-common-dbgsym
-      - nfs-kernel-server
-      - nfs-kernel-server-dbgsym
-      - python3
-      - python3-dbg
-      - python3-dev
-      - python3-ldap
-      - python3-marshmallow
-      - python3-marshmallow-doc
-      - python3-pip
-      - python3-pyvmomi
-      - python3-six
-      - python3-tenacity
-      - python3-toml
-      - python3-venv
-      - targetcli-fb
-      - telnet
-    state: present
-  register: result
-  until: result is not failed
-  retries: 3
-  delay: 60
-
 - git:
     repo: 'https://gitlab.delphix.com/devops/dcenter-gate.git'
     version: master
     dest: /opt/dcenter/lib/dcenter-gate
     accept_hostkey: yes
     update: no
+
+- alternatives:
+    name: java
+    path: /usr/lib/jvm/adoptopenjdk-java8-jdk-amd64/bin/java
 
 #
 # By default, ubuntu restricts directories where dhcpd and named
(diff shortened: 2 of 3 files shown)


scripts/run-live-build.sh
```diff
@@ -93,14 +93,25 @@ if [[ "$RUN_TYPE" == "$ALL_RUN_TYPE" || "$RUN_TYPE" == "$VM_RUN_TYPE" ]]; then
 	cp -r "$TOP/live-build/config/hooks/$VM_RUN_TYPE/." "$build_dir/config/hooks"
 fi
 
+sed "s/@@PLATFORM@@/$APPLIANCE_PLATFORM/" \
+	<"$build_dir/config/package-lists/delphix-platform.list.chroot.in" \
+	>"$build_dir/config/package-lists/delphix-platform.list.chroot"
+
+if [[ -d "$TOP/live-build/variants/$APPLIANCE_VARIANT/package-lists" ]]; then
+	for list in "$TOP/live-build/variants/$APPLIANCE_VARIANT/package-lists/"*; do
```
pzakha, line 101: nit: you can remove the `-d` check, and set `shopt -s nullglob`, which will produce an empty list if directory doesn't exist or is empty.

scripts/run-live-build.sh
```diff
@@ -93,14 +93,25 @@ if [[ "$RUN_TYPE" == "$ALL_RUN_TYPE" || "$RUN_TYPE" == "$VM_RUN_TYPE" ]]; then
...
+		[[ -f $list ]] || continue
```
pzakha, line 102: Should we fail instead, and perhaps check that the file name ends with `.chroot`?

pzakha (approved):
LGTM, thanks for implementing this.

scripts/run-live-build.sh
basil: I don't see why we should.

pzakha: I believe the package would not be consumed unless it ends with `.chroot` or `.binary`, as the suffix tells live-build at which stage to consume that package-list. We currently do not have good reasons to install it during the `.binary` stage, which is why restricting it to `.chroot` makes the most sense.

Regarding the other check, there are no valid reasons to have sub-directories in here since they won't be used, so why just silently skip them instead of issuing a failure?

basil: The existing code that deals with package lists that apply to all variants does not enforce that package lists apply to any particular stage or that they end with `.chroot` or `.binary`. In fact, an existing `delphix-platform.list.chroot.in` file is currently passed to `live-build(7)`, which ignores it. This preëxisting issue could be fixed as you describe, but such a fix is orthogonal to this PR. This PR doesn't make the preëxisting issue any worse.

Regarding the other check, it is necessary to ensure the glob match was successful.

pzakha: You are right that we don't do the check for the existing list, although we probably should. If you want to improve the checking here while you are already modifying this logic, that would be welcome, but if not feel free to drop the issue.

PR #584, opened 2021-08-06
DLPX-76802 Starting Verification container enables IP forwarding on host

sanbhara:
It has been observed that simply starting the verification container enables ip forwarding on the host (and in turn disables LRO). This seems to persist post a DEFERRED upgrade, leading to potential performance regression in networking. Having attempted to prevent the container from setting ip forwarding to true at all (see PR https://github.com/delphix/delphix-platform/pull/314) and hit an issue when the upgrade needs to install a new version of systemd (see jira for more info on this), this PR attempts to leave the container properties as is, while attempting to reset the networking state post-verification. This is a relatively short window in which LRO gets disabled and should be tolerable.

The relevant properties can be viewed on the engine using below commands :

sudo sysctl net.ipv4.ip_forward
net.ipv4.ip_forward = 0

ethtool -k <nic> | grep large
large-receive-offload: on

Manual Testing :
Kevin Greene ran network performance benchmark tests pre and post verification using an upgrade image with the changes in this PR and found that there was no longer any performance regression. The LRO property is successfully reset to enabled post verification, and the change is immediately effective without needed a reboot.

git ab-pre-push: 
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/5888/

http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/5898/ - building esx upgrade image.

PR #611, opened 2021-09-20
DLPX-77019 [Backport of DLPX-76998 to 6.0.11.0] CLONE - Starting Verification container enables IP forwarding on host

prakashsurya:


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/6138/)
`git-ab-pre-push --test-upgrade-from 6.0.7.0` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/6139/)

prakashsurya:
@sanbhara can you clarify why we did not go with this approach in #584 ?

prakashsurya:
FWIW, it looks like we added the virtual NIC in #277, and from the comments, it doesn't look like we had a good reason for doing so.

I've verified that with this change, `net.ipv4.ip_forward` stays at a value of `0`, even after running upgrade verification (I've only verified starting, upgrading, and stopping the container so far).

sanbhara:
As discussed, please keep the existing code to reset the port forwarding and lro values for deferred upgrades on engines where a prior deferred upgrade would have flipped these options. Please also confirm that lro does not get disables on starting the verification container (it should not, now that ip port forwarding is not going to get enabled, but good to verify).

sanbhara:
> @sanbhara can you clarify why we did not go with this approach in #584 ?

Not sure why this was not found as a viable solution. I remember testing with different combinations of the IPMasquerade, private, and VirtualEthernet settings but might not have gotten the right combination in my manual testing. If the container is getting upgraded without issue even when a new systemd package needs to be installed, then I think we can go ahead with this change.

prakashsurya:
> As discussed, please keep the existing code to reset the port forwarding and lro values for deferred upgrades on engines where a prior deferred upgrade would have flipped these options.

How do you feel about removing this logic from the upgrade scripts, and instead, handling this case like so: http://reviews.delphix.com/r/73426/ ?

sanbhara:
> > As discussed, please keep the existing code to reset the port forwarding and lro values for deferred upgrades on engines where a prior deferred upgrade would have flipped these options.
> 
> How do you feel about removing this logic from the upgrade scripts, and instead, handling this case like so: http://reviews.delphix.com/r/73426/ ?

That should work as well. Two points 1) do we want to do this on every stack startup and what happens if that piece of code fails...do we ever see those messages? 2) Leaving it in the upgrade code path makes sense to me since its lifecycle is somewhat tied to upgrade. We can remove it once direct upgrades from a version without the fix is no longer feasible. 

But I would be fine with either if you feel strongly about moving it to stack startup.

prakashsurya:
My motivating factor is the notion that running upgrade verification (e.g. the `upgrade-container` script, or the verification JAR, etc..) should not impact the host, and this logic in the `upgrade-container` script explicitly violates that goal. Previously, I thought it was OK to have that violation, since it was starting the container (via the same script) that implicitly enabled the setting, so it was OK to have the script disable that setting as well. Now, with this fix, the script will no longer enable the setting, so it's harder for me to justify this script violate the goal of having no impact on the host system, by explicitly disabling the setting.

I don't feel strongly, as practically I don't think it really matters.. but just from a layering perspective, I feel like it's better to have the virtualization service manipulate the host system, rather than the `upgrade-container` script.

Lets see if folks are OK landing my virtualization change; if that hits any road blocks during review, then I'll drop that, and keep changes from #584.

> 1) do we want to do this on every stack startup and what happens if that piece of code fails...do we ever see those messages?

That's a valid point, about if we want to run it on each stack restart, or not.. let's continue this discussion on that review..

prakashsurya:
@sanbhara I've decided to go with what you suggested, and just keep the logic from #584 in place.

sebroy (approved):

sanbhara (approved):

PR #615, opened 2021-10-14
DLPX-77901 Prevent services from being re-enabled on upgrade

pzakha:
The main idea behind this change is to make it easier to control which services get to run and which services should not run. In short, the solution is to mask services instead of just disabling them when we want them to not run, and then unmask them when we want them to run again.

This work is split up into 3 reviews which must all be pushed together. The 2 other PRs are:
- dlpx-app-gate: http://reviews.delphix.com/r/74397/
- delphix-platform: https://github.com/delphix/delphix-platform/pull/333

This addresses the following issues:
1. Some services that we want to never run end up getting re-enabled when some packages are upgraded / re-installed.
2. Other services that we want to be disabled / enabled conditionally end up in the wrong state after / during an upgrade.

One of the reasons for those issues is that the previous approach of relying on preinst and postinst scripts relies on a specific order for those scripts to be executed with respect to different packages, which did not hold anymore once we switched to Ubuntu 20.04. The other reason is that sometimes packages are re-installed or upgraded and end up re-enabling themselves unexpectedly.

The new approach does not rely on preinst/postinst scripts anymore, and instead performs operations that modify the state of services at certain specific times:
1. On initial install, after all packages have been installed
2. During the normal operation of the appliance, outside of upgrade
3. On upgrade, before packages are upgraded, and after all packages have been upgraded (must consider: in-place upgrade, not-in-place upgrade, inside container, outside container)

This review takes care of masking services that need to be masked on initial install (1) and on upgrade (3), while the app-gate review takes care of masking/unmasking services during the normal operation of the appliance (2).

The old preinst/postinst logic that is now made obsolete by this review is removed both by the delphix-platform and the dlpx-app-gate reviews.

## Additional Notes
 - Remove the DELPHIX_FLUENTD_IS_ENABLED fix since the minimum upgrade version is now way past 6.0.0.0, so that fix is not necessary anymore
 - snmpd.service should have been disabled on initial install, but it wasn't so for systems where the SNMP functionality is not in use, we disable (mask) that service during the upgrade. This fixes DLPX-77197 and DLPX-76796.
 - docker.service is a special case. It appears that we need to mask it for the duration of the upgrade (see comment in the code). This was previously "accidentally" done in the preinst/postinst hooks of delphix-virtualization since docker.service comes disabled out of the box, and remains disabled forever. Note that the fact that it is disabled doesn't prevent it from being started by the delphix-mgmt service as it is listed as `Requires=docker.service`. I keep the same behaviour here (keep docker disabled), however we may look into enabling it in the future instead.
- The reason why `fix_and_migrate_services()` both fixes service states and also migrates them in the case of not-in-place upgrade is to keep all the logic related to services in the same place, so it is easier to reason about it in its ensemble. 

## Testing
- Checked that the snmpd service is masked on upgrade when it was supposed to be disabled.
- ab-pre-push on master: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/6368
- ab-pre-push on focal: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/6371/
- ab-pre-push on focal (forced not-in-place upgrade): http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/6372/

Note that some of the upgrade runs have hit known failures: QA-30927, QA-30764, DLPX-72032

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
@@ -39,17 +39,93 @@
     state: link
 
 #
+# The section below deals with disabling and masking services that need to be
+# disabled on initial install. Some of those services will remain disabled
+# forever, while others could be enabled by the various applications running
+# on the Delphix Appliance.
+#
+# Our approach is to mask services that need to be disabled. Masking a service
+# instead of just disabling it has 2 advantages:
+#  1. A service that is just disabled will not be started automatically by
+#     systemd on boot, however it can be started by another service if it has
+#     a certain set of dependencies on it (such as "Requires", "PartOf"). A
+#     service that is masked cannot be started in any case (although a
+#     service that is already running and has just been masked will not be
+#     stopped automatically).
+#  2. A service that is disabled can get re-enabled by dpkg maintenance scripts
+#     when the package that provides that service is upgraded. A service that
+#     is masked will not be unmasked.
+#
+# While masking a service is sufficient, we also disable most of the services
+# that we mask. This is not strictly necessary, but is done to remain
+# consistent with how we did things in the past. Our current logic that deals
+# with re-enabling those services will both unmask and enable # them.
+#
+# We have divided the disabling and masking of the services in multiple
+# code blocks to group together services that are handled in the same way by
+# the appliance.
+#
+# Note that if you want to modify this list make sure to also update
+# the logic in fix_and_migrate_services() that is invoked during upgrade.
+# You may also want to look at the logic that handles enabling disabling
+# services in dlpx-app-gate.
+#
+
+#
+# The services in this section should always remain disabled & masked.
+#
+- name: Disable services that should never run
+  command: "systemctl disable {{ item }}"
+  with_items:
+    - nginx.service
+    - postgresql.service
+    - systemd-timesyncd.service
+
+- name: Mask services that should never run
+  command: "systemctl mask {{ item }}"
+  with_items:
+    - nginx.service
+    - postgresql.service
+    - systemd-timesyncd.service
+
+#
+# The services in this section should be disabled & masked initially, but
+# can be later dynamically enabled by the appliance.
+#
+- name: Disable services that should not be running by default
+  command: "systemctl disable {{ item }}"
+  with_items:
+    - delphix-fluentd.service
+    - delphix-masking.service
+    - ntp.service
+    - snmpd.service
+
+- name: Mask services that should not be running by default
+  command: "systemctl mask {{ item }}"
+  with_items:
+    - delphix-fluentd.service
+    - delphix-masking.service
+    - ntp.service
+    - snmpd.service
+
+#
 # Because we want an NFSv4-only configuration out of the box, we need to mask
 # NFSv3 services so that they don't get automatically started at boot via
 # dependencies. The virtualization software is responsible for unmasking and
 # starting these services if NFSv3 needed at runtime.
 #
-- name: Mask NFSv3 services
-  file:
-    src: "/dev/null"
-    dest: "/etc/systemd/system/{{ item }}"
-    state: link
+- name: Mask NFSv3 services that should not be started automatically
+  command: "systemctl mask {{ item }}"
   with_items:
-      - rpc-statd.service
-      - rpcbind.service
-      - rpcbind.socket
+    - rpcbind.service
+    - rpcbind.socket
+    - snmpd.service
+
+#
+# We disable docker. Instead of being started automatically by systemd, it
+# gets started via a dependency of delphix-virtualization.
+#
+- name: Disable docker
+  command: "systemctl disable {{ item }}"
+  with_items:
+    - docker.service
--- a/live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
@@ -74,15 +74,10 @@
 #
 # The services in this section should always remain disabled & masked.
 #
-- name: Disable services that should never run
-  command: "systemctl disable {{ item }}"
-  with_items:
-    - nginx.service
-    - postgresql.service
-    - systemd-timesyncd.service
-
-- name: Mask services that should never run
-  command: "systemctl mask {{ item }}"
+- name: Disable and mask services that should never run
+  shell: |
+    systemctl disable {{ item }}
+    systemctl mask {{ item }}
   with_items:
     - nginx.service
     - postgresql.service
@@ -92,16 +87,10 @@
 # The services in this section should be disabled & masked initially, but
 # can be later dynamically enabled by the appliance.
 #
-- name: Disable services that should not be running by default
-  command: "systemctl disable {{ item }}"
-  with_items:
-    - delphix-fluentd.service
-    - delphix-masking.service
-    - ntp.service
-    - snmpd.service
-
-- name: Mask services that should not be running by default
-  command: "systemctl mask {{ item }}"
+- name: Disable and mask services that should not be running by default
+  shell: |
+    systemctl disable {{ item }}
+    systemctl mask {{ item }}
   with_items:
     - delphix-fluentd.service
     - delphix-masking.service
@@ -119,7 +108,7 @@
   with_items:
     - rpcbind.service
     - rpcbind.socket
-    - snmpd.service
+    - rpc-statd.service
 
 #
 # We disable docker. Instead of being started automatically by systemd, it
--- a/upgrade/upgrade-scripts/common.sh
+++ b/upgrade/upgrade-scripts/common.sh
@@ -393,7 +393,7 @@ function is_svc_masked_or_disabled() {
 # This function has 2 tasks:
 #  1. Fix/update the state of some services to be in line with what is expected
 #     in this version of the appliance.
-#  2. If we are doing a not-in-plaace upgrade, then migrate the state of the
+#  2. If we are doing a not-in-place upgrade, then migrate the state of the
 #     services into the upgrade container.
 #
 # It can be called from 2 different contexts:
@@ -406,26 +406,16 @@ function fix_and_migrate_services() {
 	local container="$1"
 
 	#
-	# This function can be called from the execute script, however in that
-	# case we only want it to run when called on the running system and not
-	# in the upgrade container.
+	# This function must be called from outside an upgrade container as it
+	# uses the state of the services on the running system as the source of
+	# truth. Since we want the logic in this script to apply both to an
+	# upgrade container and to the running system (in case of an in-place
+	# upgrade), we call it from two places: create_upgrade_container() and
+	# the execute script. The former will apply this logic on a container
+	# while the latter will apply this logic to the running system.
 	#
 	if systemd-detect-virt --container --quiet; then
 		echo "fix_and_migrate_services: should not run inside container"
-		return
-	fi
-
-	#
-	# We only want to run this logic on the Delphix Appliance variants.
-	# We especially want to avoid running it on the internal-dcenter
-	# variant, which has a different set of requirements for which
-	# services should be enabled / disabled. For instance, the
-	# dcenter-internal variant relies on either systemd-timesyncd.service
-	# or ntp.service to be enabled.
-	#
-	if ! dpkg -l delphix-virtualization &>/dev/null; then
-		echo "fix_and_migrate_services: should only run on the" \
-			"Delphix Appliance"
 		return
 	fi
 
@@ -439,17 +429,6 @@ function fix_and_migrate_services() {
 			mask_service snmpd "$container"
 		fi
 	fi
-
-	#
-	# Some services should be always masked.
-	#
-	while read -r svc; do
-		mask_service "$svc" "$container"
-	done <<-EOF
-		nginx.service
-		postgresql.service
-		systemd-timesyncd.service
-	EOF
 
 	#
 	# docker is a special case. It needs to be masked for the duration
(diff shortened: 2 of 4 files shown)


live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
```diff
@@ -38,18 +38,94 @@
...
+    - nginx.service
+    - postgresql.service
+    - systemd-timesyncd.service
+
+- name: Mask services that should never run
+  command: "systemctl mask {{ item }}"
+  with_items:
+    - nginx.service
+    - postgresql.service
+    - systemd-timesyncd.service
```
sebroy, line 89: would be good to combine the disable + mask commands into one ansible task so that the list of services isn't duplicated (and is easier to read). E.g.:
```
- name: Disable and mask services that should never run
  shell: |
    systemctl disable {{ item }}
    systemctl mask {{ item }}
  with_items:
    - nginx.service
    - postgresql.service
    - systemd-timesyncd.service
```

upgrade/upgrade-scripts/common.sh
```diff
@@ -362,3 +362,145 @@ function verify_upgrade_not_in_progress() {
...
+	# case we only want it to run when called on the running system and not
+	# in the upgrade container.
+	#
+	if systemd-detect-virt --container --quiet; then
+		echo "fix_and_migrate_services: should not run inside container"
+		return
+	fi
+
+	#
+	# We only want to run this logic on the Delphix Appliance variants.
```
sebroy, line 419: It's too bad that we're leaking the requirements of specific variants into the generic upgrade code. It makes me think that there is an upgrade design element missing for handling different variants.

upgrade/upgrade-scripts/common.sh
```diff
@@ -362,3 +362,145 @@ function verify_upgrade_not_in_progress() {
...
+
+	#
+	# Some services should be always masked.
+	#
+	while read -r svc; do
+		mask_service "$svc" "$container"
+	done <<-EOF
+		nginx.service
+		postgresql.service
+		systemd-timesyncd.service
```
sebroy, line 451: It feels like this list could live in one place (e.g. it could be a list in a file that is packaged/installed with the relevant variant). That way, the list isn't duplicated, and you also don't need the special code that skips this function for some variants?

upgrade/upgrade-scripts/common.sh
```diff
@@ -362,3 +362,145 @@ function verify_upgrade_not_in_progress() {
...
+			mask_service "$svc" "$container"
+	done <<-EOF
+		delphix-fluentd.service
+		delphix-masking.service
+		nfs-mountd.service
+		ntp.service
+		rpc-statd.service
+		rpcbind.service
+		rpcbind.socket
+		snmpd.service
```
sebroy, line 504: Similarly as above, it feels like we could declare these in one place instead of two.

sebroy (approved):

live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
```diff
@@ -38,18 +38,94 @@
...
+#
+# The services in this section should be disabled & masked initially, but
+# can be later dynamically enabled by the appliance.
+#
+- name: Disable services that should not be running by default
+  command: "systemctl disable {{ item }}"
+  with_items:
+    - delphix-fluentd.service
+    - delphix-masking.service
+    - ntp.service
```
prakashsurya, line 100: Previously, we disabled this for all variants (via the delphix-platform package), but now it seems this will only apply to variants that include the virtualization package.. is that OK?

live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
```diff
@@ -38,18 +38,94 @@
...
+    - snmpd.service
+
+- name: Mask services that should not be running by default
+  command: "systemctl mask {{ item }}"
```
prakashsurya, line 104: Why do we need multiple "disable + mask" sections, when both appear to be doing the same thing, but with different sets of services? why cant all these services use the same tasks to disable + mask them?

live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
prakashsurya: Better yet, we could probably use the systemd module: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/systemd_module.html

upgrade/upgrade-scripts/common.sh
prakashsurya: Yea, I'd like to think a bit more about this.. I also don't like this..

upgrade/upgrade-scripts/common.sh
```diff
@@ -362,3 +362,145 @@ function verify_upgrade_not_in_progress() {
...
+function fix_and_migrate_services() {
+	local container="$1"
+
+	#
+	# This function can be called from the execute script, however in that
+	# case we only want it to run when called on the running system and not
+	# in the upgrade container.
+	#
+	if systemd-detect-virt --container --quiet; then
+		echo "fix_and_migrate_services: should not run inside container"
```
prakashsurya, line 414: why not?

upgrade/upgrade-scripts/common.sh
```diff
@@ -362,3 +362,145 @@ function verify_upgrade_not_in_progress() {
...
+	# upgrade, which would also force a restart of the delphix-mgmt
+	# service (since the latter has a dependency on docker.service), and
+	# thus interrupt the upgrade.
+	#
+	# Once the upgrade is done we restart delphix.target, which will
+	# attempt to restart both delphix-mgmt and docker, so docker
+	# needs to be unmasked before that point. As such, docker is
+	# unmasked at the end of the execute script.
+	#
+	mask_service docker.service "$container"
```
prakashsurya, line 466: Ugh, I don't like how the upgrade code now need to know about service dependencies like this.. I'm not sure I agree that putting this logic here, is the correct place to do this, from a layering/abstraction point of view.. the dependency comes from the app-gate, so I feel like it's the app-gate's (or maybe the docker package itself) responsibility to ensure this works..

I really don't like how we continue to need complex integrations between what's supposed to be package-agnostic upgrade logic, and the packages that are being upgrade.

upgrade/upgrade-scripts/common.sh
pzakha: I've added a comment explaining this.

live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
pzakha: Here are the changes using the ansible module. I personally can't say I really like it more than just using the commands, but here are the changes nevertheless: https://github.com/pzakha/appliance-build/commit/92456c12e0bc55ff5cdcf8c9d1f14d886202fd35

This is the output it produces:
```
16:48:15  TASK [appliance-build.virtualization-common : Disable and mask services that should never run] ***
16:48:15  task path: /var/tmp/jenkins/workspace/devops-gate/master/appliance-build-stage1/pre-push/appliance-build/live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml:77
16:48:15  changed: [binary] => (item=nginx.service) => {"ansible_loop_var": "item", "changed": true, "enabled": false, "item": "nginx.service", "name": "nginx.service", "status": {}}

16:48:17  changed: [binary] => (item=postgresql.service) => {"ansible_loop_var": "item", "changed": true, "enabled": false, "item": "postgresql.service", "name": "postgresql.service", "status": {}}

16:48:20  changed: [binary] => (item=systemd-timesyncd.service) => {"ansible_loop_var": "item", "changed": true, "enabled": false, "item": "systemd-timesyncd.service", "name": "systemd-timesyncd.service", "status": {}}
16:48:20  
16:48:20  TASK [appliance-build.virtualization-common : Disable and mask services that should not be running by default] ***
16:48:20  task path: /var/tmp/jenkins/workspace/devops-gate/master/appliance-build-stage1/pre-push/appliance-build/live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml:91

16:48:23  changed: [binary] => (item=delphix-fluentd.service) => {"ansible_loop_var": "item", "changed": true, "enabled": false, "item": "delphix-fluentd.service", "name": "delphix-fluentd.service", "status": {}}

16:48:25  changed: [binary] => (item=delphix-masking.service) => {"ansible_loop_var": "item", "changed": true, "enabled": false, "item": "delphix-masking.service", "name": "delphix-masking.service", "status": {}}

16:48:28  changed: [binary] => (item=ntp.service) => {"ansible_loop_var": "item", "changed": true, "enabled": false, "item": "ntp.service", "name": "ntp.service", "status": {}}

16:48:31  changed: [binary] => (item=snmpd.service) => {"ansible_loop_var": "item", "changed": true, "enabled": false, "item": "snmpd.service", "name": "snmpd.service", "status": {}}
16:48:31  
16:48:31  TASK [appliance-build.virtualization-common : Mask NFSv3 services that should not be started automatically] ***
16:48:31  task path: /var/tmp/jenkins/workspace/devops-gate/master/appliance-build-stage1/pre-push/appliance-build/live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml:108

16:48:34  changed: [binary] => (item=rpcbind.service) => {"ansible_loop_var": "item", "changed": true, "item": "rpcbind.service", "name": "rpcbind.service", "state": "stopped", "status": {}, "warnings": ["Target is a chroot or systemd is offline. This can lead to false positives or prevent the init system tools from working."]}

16:48:36  changed: [binary] => (item=rpcbind.socket) => {"ansible_loop_var": "item", "changed": true, "item": "rpcbind.socket", "name": "rpcbind.socket", "state": "stopped", "status": {}, "warnings": ["Target is a chroot or systemd is offline. This can lead to false positives or prevent the init system tools from working."]}

16:48:38  ok: [binary] => (item=snmpd.service) => {"ansible_loop_var": "item", "changed": false, "item": "snmpd.service", "name": "snmpd.service", "state": "stopped", "status": {}, "warnings": ["Target is a chroot or systemd is offline. This can lead to false positives or prevent the init system tools from working."]}
16:48:38  [WARNING]: Target is a chroot or systemd is offline. This can lead to false
16:48:38  positives or prevent the init system tools from working.
16:48:38  
16:48:38  TASK [appliance-build.virtualization-common : Disable docker] ******************
16:48:38  task path: /var/tmp/jenkins/workspace/devops-gate/master/appliance-build-stage1/pre-push/appliance-build/live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml:122

16:48:41  changed: [binary] => (item=docker.service) => {"ansible_loop_var": "item", "changed": true, "enabled": false, "item": "docker.service", "name": "docker.service", "status": {}}
```
Note that I caught a bug where we were masking snmpd twice, which is fixed now. 

The thing I don't like so much is that leaving a service enabled but masked fails with:
```
15:58:43  failed: [binary] (item=rpcbind.service) => {"ansible_loop_var": "item", "changed": false, "item": "rpcbind.service", "msg": "Unable to enable service rpcbind.service: Synchronizing state of rpcbind.service with SysV service script with /lib/systemd/systemd-sysv-install.\nExecuting: /lib/systemd/systemd-sysv-install enable rpcbind\nFailed to enable unit, unit /etc/systemd/system/rpcbind.service is masked.\n"}
```
Instead have to use "state: stopped" with "masked: yes", which makes it different from other calls, and also gives warnings.

With all that said, if both of you prefer the systemd module, I can switch to that.

live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
```diff
@@ -38,18 +38,83 @@
...
-    state: link
+- name: Mask NFSv3 services that should not be started automatically
+  command: "systemctl mask {{ item }}"
+  with_items:
+    - rpcbind.service
+    - rpcbind.socket
+    - rpc-statd.service
+
+#
+# We disable docker. Instead of being started automatically by systemd, it
```
prakashsurya, line 114: Why are you choosing to do it this way? I'm not aware of any other service that we disable, but intend for it to get started anyways, so I think it'd be good to explain why docker must be different in this regard.

upgrade/upgrade-scripts/common.sh
prakashsurya: Yes, please open a bug; this is bizarre, and not something the upgrade logic should be "fixing".

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -395,6 +395,8 @@ function create_upgrade_container() {
 	EOF
 		die "failed to create container service override file"
 
+	fix_and_migrate_services "$CONTAINER"
```
prakashsurya, line 398: since this function will mask the docker service, where do we unmask that service?

live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
```diff
@@ -38,18 +38,83 @@
...
 # dependencies. The virtualization software is responsible for unmasking and
 # starting these services if NFSv3 needed at runtime.
 #
-- name: Mask NFSv3 services
-  file:
-    src: "/dev/null"
-    dest: "/etc/systemd/system/{{ item }}"
-    state: link
+- name: Mask NFSv3 services that should not be started automatically
+  command: "systemctl mask {{ item }}"
```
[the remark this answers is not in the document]
sebroy: These need to be masked, otherwise they get started automatically (even when disabled) via dependency.

upgrade/upgrade-scripts/execute
```diff
@@ -375,6 +354,12 @@ dpkg-query -Wf '${Conffiles}\n' | awk '$3 == "obsolete" {print $1}' |
 			die "failed to reinstall package '$package'"
 	done || die "failed to remove obsolete package configuration files"
 
+#
+# Unmask docker, which was masked at the beginning of the upgrade in
+# fix_and_migrate_services().
+#
+systemctl unmask docker.service
```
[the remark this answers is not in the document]
pzakha: Yeah I think I could do that.

upgrade/upgrade-scripts/upgrade-container
pzakha: Yeah, this way docker is masked straight away when we first run the container, that said I'm not convinced that is necessary. I'll try the change you proposed and move it into execute instead.

upgrade/upgrade-scripts/common.sh
pzakha: I've opened https://jira.delphix.com/browse/DLPX-77949

pzakha:
Re-testing with latest feedback:
6.0.11.0 -> master: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/6388/
6.0.11.0 -> focal: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/6389/

prakashsurya (approved):

PR #620, opened 2021-10-19
DLPX-77974 Add minimum version property to root dataset

vimalraj-anbarasu:
Adding minimum version property to root dataset

The minimum version property is used for
1. Upgrade: to determine minimum version needed for this image
2. Replication, FCR: minimum FCR version compatible

Test:
http://collins.d.delphix.com:20855/job/devops-gate/job/master/job/appliance-build/job/master/job/pre-push/29/console
Verified the upgrade image for the property.

New Job: http://collins.d.delphix.com:20855/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1/console

### DIFF
--- a/upgrade/upgrade-scripts/upgrade-container
+++ b/upgrade/upgrade-scripts/upgrade-container
@@ -119,6 +119,13 @@ function create_upgrade_container() {
 			"$ROOTFS_DATASET" "$ROOTFS_DATASET@$SNAPSHOT_NAME"
 		copy_optional_dataset_property "$PROP_HOTFIX_VERSION" \
 			"$ROOTFS_DATASET" "$ROOTFS_DATASET@$SNAPSHOT_NAME"
+		#
+		# The minimum version property was introduced in 6.0.12.0.
+		#
+		if compare_versions "$(get_current_version)" ge "6.0.12.0"; then
+			copy_required_dataset_property "$PROP_MINIMUM_VERSION" \
+				"$ROOTFS_DATASET" "$ROOTFS_DATASET@$SNAPSHOT_NAME"
+		fi
 		;;
 	rollback)
 		#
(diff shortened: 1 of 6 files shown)


pzakha (commented):
The logic LGTM.
- Can you verify that on rollback the property is reverted to the previous value?
- Could you check with QA to see if we can add tests that verify that this value is set appropriately?

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -149,6 +151,8 @@ function create_upgrade_container() {
 			"$ROOTFS_DATASET@$SNAPSHOT_NAME" "rpool/ROOT/$CONTAINER"
 		copy_optional_dataset_property "$PROP_HOTFIX_VERSION" \
 			"$ROOTFS_DATASET@$SNAPSHOT_NAME" "rpool/ROOT/$CONTAINER"
+		copy_optional_dataset_property "$PROP_MINIMUM_VERSION" \
```
prakashsurya, line 154: Why are we using `copy_optional` here, rather than `copy_required`?

Presumably this is to handle the case of moving from a version that didn't have the field, to a version that does have the field; is that correct?

If this is the only consideration, then I'd rather not use the `optional` function for all delphix versions, even for versions that we expect the field to be there.. I'd suggest we tweak this such that we only attempt the copy when coming from versions where we expect the field to exist, and then simply use the `required` function.. this way, if we're on a version that should have the field, but it doesn't, we'll properly report the error..

I think we can easily do this.. e.g. something like this:
```
if compare_versions "$(get_current_version)" ge "6.0.11.0"; then
    copy_required_dataset_property ...
fi
```

What do you think?

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -149,6 +151,8 @@ function create_upgrade_container() {
...
+			"$ROOTFS_DATASET" "$ROOTFS_DATASET@$SNAPSHOT_NAME"
```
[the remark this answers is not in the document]
vimalraj-anbarasu: https://jira.delphix.com/browse/QA-31226 --> QA task to add test

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -119,6 +119,13 @@ function create_upgrade_container() {
 			"$ROOTFS_DATASET" "$ROOTFS_DATASET@$SNAPSHOT_NAME"
 		copy_optional_dataset_property "$PROP_HOTFIX_VERSION" \
 			"$ROOTFS_DATASET" "$ROOTFS_DATASET@$SNAPSHOT_NAME"
+		#
+		# The minimum version property was introduced in 6.0.11.0.
+		#
+		if compare_versions "$(get_current_version)" ge "6.0.11.0"; then
```
pzakha, line 125: I'm assuming this will be backported to 6.0/stage and not 6.0/release, so this should be 6.0.12.0 instead

pzakha (approved):
Approved, pending version check fix and rollback tests

sumedhbala-delphix (approved):

PR #624, opened 2021-11-04
DLPX-78255 ui-precommit fails due to missing Chrome dependency libnss3.so

pzakha:
See comment in the code for a bit more context on the description.

## Alternative approach
I've considered an alternative which is to use a 3rd party PPA to get a chromium package with all the dependencies pulled implicitly, but for security reasons we do not want to use 3rd party PPAs unless strictly necessary.
See https://github.com/pzakha/appliance-build/commit/f847b624cb1c1d53da57ee6ae1a84309350c7956 for the alternative approach.

## Testing
- ab-pre-push for creating internal-testing variant: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/6506/ (pass)
- UI-precommit on tip master: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/integration-tests/job/ui-precommit/job/pre-push/1774/console (failed due to a suspected recent regression on master)
- UI-precommit on master from yesterday morning: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/integration-tests/job/ui-precommit/job/pre-push/1775/ (pass)

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.unittest-internal/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.unittest-internal/tasks/main.yml
@@ -15,11 +15,6 @@
 #
 
 ---
-#
-# libxss1 and libgtk-3-0 are required dependencies for
-# chromium-browser. They are missing in the chromium-browser
-# package. Manually install them here.
-#
 - apt:
     name:
       - adoptopenjdk-java8-jdk
@@ -29,9 +24,56 @@
       - docker.io
       - git
       - python3-minimal
-      - chromium-browser
-      - libxss1
+    state: present
+
+#
+# The UI-precommit suite requires some chromium-browser dependendencies to be
+# installed. However on 20.04 chromium is provided by a snap, and the
+# "chromium-browser" package neither provides any dependencies nor the actual
+# chromium browser, but rather directs users to install the chromium snap.
+# While we could get chromium from a thrid party PPA, we do not actually need
+# chromium for UI-precommit to work, but only some of the libraries that
+# came as dependencies. As such we install here all the lib dependencies
+# that were brought in by the chromium-browser package on 18.04 as per:
+# https://packages.ubuntu.com/bionic/chromium-browser.
+#
+- apt:
+    name:
+      - libasound2
+      - libatk-bridge2.0-0
+      - libatk-bridge2.0-0
+      - libatspi2.0-0
+      - libc6
+      - libcairo2
+      - libcups2
+      - libdbus-1-3
+      - libdrm2
+      - libexpat1
+      - libgbm1
+      - libgcc1
+      - libgdk-pixbuf2.0-0
+      - libglib2.0-0
       - libgtk-3-0
+      - libnspr4
+      - libnss3
+      - libpango-1.0-0
+      - libpangocairo-1.0-0
+      - libwayland-client0
+      - libx11-6
+      - libx11-xcb1
+      - libxcb1
+      - libxcomposite1
+      - libxcursor1
+      - libxdamage1
+      - libxext6
+      - libxfixes3
+      - libxi6
+      - libxkbcommon0
+      - libxrandr2
+      - libxrender1
+      - libxshmfence1
+      - libxss1
+      - libxtst6
     state: present
 
 - user:


prakashsurya (approved):

ckdelphix:
Some background:
We use [karma-chrome-launcher](https://github.com/karma-runner/karma-chrome-launcher) and [Puppeteer](https://github.com/puppeteer/puppeteer) to run ui-precommit tests.
Puppeteer automatically downloads a recent version of Chromium. And karma-chrome-launcher uses the headless Chrome downloaded by Puppeteer.
But Puppeteer doesn't download Chromium's dependencies. That's the reason that we install Chromium to the system to make sure all its dependencies are installed properly for the Chromium that Puppeteer downloads.

I think with this new approach, it is possible in the future, if a new version of Chromium adds new dependencies, we need to manually add them to our list.

As we need to make sure the UI runs fine on the latest version of browser, we cannot lock down the Chromium version.

I think the 3rd party PPA solution may be better that we don't need to worry about Chromium version updates and dependencies list changes. Another option maybe we don't want to use Ubuntu 20.04 for ui-precommit, but lock it down to Ubuntu 18.04 just for ui-precommit. As ui-precommit just runs UI unit tests, no Delphix engine backend is needed so we may not really need to run it on the latest version of OS.

pzakha:
Thanks for the background @ckdelphix.

I'm not sure if using Ubuntu 18.04 for ui-precommit would solve any issues since as we've seen in the past the `chromium-browser` package provided by Ubuntu sometimes missed some dependencies, and even if it didn't it could run a different version of chrome than what's fetched by Puppeteer and thus miss some dependencies. 

The 3rd party PPA would probably be the most convenient way of doing this once and forgetting about it, but our latest security policy dictates that we should not use third party PPAs unless strictly necessary, and even then we'd need to do some additional vetting & locking down to prevent potential attack vectors.

We can sync-up with the security team to see if it's okay to make an exception here, but given that we have a viable (although less convenient) alternative here, I doubt that this will be approved. Let me know if you'd prefer that, or if we should ship this as-is to unblock UI-precommit.

sebroy:
@ckdelphix regarding:
> Another option maybe we don't want to use Ubuntu 20.04 for ui-precommit, but lock it down to Ubuntu 18.04 just for ui-precommit.

I wholeheartedly agree with this, and would go further to say that the OS used to run unit tests need not be Ubuntu, and furthermore should not be built using appliance-build given that this framework is designed to build Delphix product images. It should be possible to generate a "unit test" golden image for use in running unit tests that is generated using a simpler mechanism that is less costly to maintain. The requirements are much narrower given that unit tests can run on any OS and on any platform.

ckdelphix (approved):

prakashsurya:
> furthermore should not be built using appliance-build given that this framework is designed to build Delphix product images.

I'm not sure I'd go that far.. I don't see any reason why we shouldn't use this framework for non-product images.. e.g. we do that for our dcenter images.

>  It should be possible to generate a "unit test" golden image for use in running unit tests that is generated using a simpler mechanism that is less costly to maintain.

Sure, we could do that, and we used to do that.. we chose to deviate from that prior solution, though.. here's a link to the doc written by @basil that goes over how/why we wound up in the current situation of using appliance-build for this image: [Unit Test Image Maintainability](https://docs.google.com/document/d/1V1UgR0gejzkCxzONdE-Sadeyhekln-VnFoGGQQXS99s/edit?usp=sharing)

sebroy:
@prakashsurya I see dcenter images as different, as they are meant to be used for "dog fooding" product code, which by definition requires that they include product code. Unit tests, by definition, have no such requirement, they can run on Windows...

That said, I understand that appliance-build addressed a number of maintenance issues with the previous golden images. There are a few issues that linger in my mind that we should discuss (perhaps offline as a group) if we are to continue with this:
1. There is no testing to alert anyone that a given build will break the UI unit tests that rely on that variant.
2. If we had such tests, we'd have a separately problem which is that updates to Ubuntu which could cause some incompatibility with a unit test framework would grind the entire product lifecycle to a halt unnecessarily (because there is no requirement that the unit test environment be updated at any frequency).
3. Having a product team's (Systems Platform) responsibility include providing a test environment for other teams to run unit tests on is unnatural to me.

prakashsurya:
I don't disagree with any of that, just wanted to help highlight why were in this situation to begin with. We need to balance the new issues due to this new solution, with the old issues due to the old solution. If we, now, only focus on the new problems, my concern is we'll just regress back to the original problems. 

re: 1.. I don't see why we couldn't the unittest variant via "git-ab-pre-push".. i.e. run the tests on the image the build produces.. and either build the image by default, or only do this when "-v unittest" is specified.

prakashsurya:
IMO, it's also worth mentioning, even if we continue to use appliance-build to generate the images, I don't see why we have to build and update the dcenter images using the same cadence as the product; e.g. we could pin this image to some 18.04 based version, and then (optionally) move to 20.04 later, long after after the product has already moved to 20.04.

prakashsurya:
```
136     /*                                                                          
137      * While params.IMPORT_VARIANT accepts a space separated list, the current implementation only
138      * allows for a single import variant. The reason we are accepting a list in the parameters is
139      * allow for improving the implementation in a future change by importing multiple variants (for
140      * example, importing both the qa and unittest variant, and running the unit tests with the
141      * unittest variant and the BlackBox tests with the QA variant).            
142      */     
```
from the Jenkins logic for `git-ab-pre-push`, looks like we even intended to support this workflow for testing the "unittest" variant.

PR #630, opened 2021-11-12
DRAFT: DLPX-78371 Upgrade fails because some packages are unexpectedly autoremoved by apt

pzakha:
## Testing
- ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/6569/
- tested manual upgrade from 6.0.11.0 -> trunk using external-standard variant

### DIFF
--- a/upgrade/upgrade-scripts/execute
+++ b/upgrade/upgrade-scripts/execute
@@ -297,6 +297,10 @@ zcat "/usr/share/doc/delphix-entire-$platform/packages.list.gz" |
 	xargs_apt_get install -y --allow-downgrades ||
 	die "failed to install packages listed in packages.list.gz file"
 
+zcat "/usr/share/doc/delphix-entire-$platform/packages.list.gz" |
+	cut -d= -f1 | xargs apt-mark manual ||
+	die "failed to mark as manual packages listed in packages.list.gz file"
+
 #
 # After we've successfully installed the new packages, we need to remove
 # all packages that are no longer required. For example, if the old


upgrade/upgrade-scripts/execute
```diff
@@ -297,6 +297,10 @@ zcat "/usr/share/doc/delphix-entire-$platform/packages.list.gz" |
 	xargs_apt_get install -y --allow-downgrades ||
 	die "failed to install packages listed in packages.list.gz file"
 
+zcat "/usr/share/doc/delphix-entire-$platform/packages.list.gz" |
```
prakashsurya, line 300: Can you clarify why we need to do this? IIRC, we "apt-get install" each package in this list in the command above, which marks each package as manually installed.. right?

pzakha: That's what I thought as well, but it doesn't seem to be the case. I've put a bit more details in JIRA

pzakha: Here's what we see on a system that was upgraded from 6.0.12.0 -> trunk:
```
$ (zcat /usr/share/doc/delphix-entire-aws/packages.list.gz | cut -d= -f1 && apt-mark showauto) | sort | uniq -d
adduser
apparmor
apt
aptitude
aptitude-common
apt-utils
base-files
base-passwd
bind9-host
binfmt-support
binutils
...
x11-common
xauth
xdg-user-dirs
xkb-data
xz-utils
zfs-initramfs
zfsutils-linux
zfs-zed
zlib1g
```

There's a total of 377 packages that are marked as automatically installed despite being in `packages.list`

prakashsurya: Hm, interesting.. OK, fair enough.. 

This also makes me wonder if we should do something like this so our VM images are also in a state where all packages are marked as "manually" installed.. at least for consistency.. but that can be left for a future discussion.

prakashsurya (approved):

sebroy (approved):

PR #636, opened 2021-12-02
DLPX-78582 6.0/stage build needs python3.8 in order to release vSDK 4…

mrburke:
vSDK 4.0.0 relies on the existence of python3.8 during build time. To avoid breaking git-au, we also need it installed on the engine.

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.qa-internal/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.qa-internal/tasks/main.yml
@@ -40,6 +40,9 @@
 # can move the build dependencies from the list below back into the
 # "internal-dev" playbook.
 #
+- apt_repository:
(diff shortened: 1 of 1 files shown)


prakashsurya (approved):
LGTM. Can you also forward port this change to the master branch, albeit without adding the APT repository, since I don't think that's needed.. i.e. just the addition of `python3.8`  to the dependency list, to make it clear that package is required.

sdimitro (approved):

mrburke:
Yep will do

mrburke:
> Yep will do



> LGTM. Can you also forward port this change to the master branch, albeit without adding the APT repository, since I don't think that's needed.. i.e. just the addition of `python3.8` to the dependency list, to make it clear that package is required.

Yep will do

PR #654, opened 2022-01-21
DLPX-78736 Several test_api_and_cli_list_operation tests failed after…

brad-lewis:
After upgrade the td-agent is not masked, but is running and keeping port 8888 tied up.  This is causing the failure.   It seems the td-agent is missing from a list of services that are masked after upgrade.

Testing:

ab-pre-push (pending) - http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/482/


sebroy (approved):

prakashsurya:
can you comment on why we don't mask this via the appliance-build ansible logic too, i.e. for new VM images?

prakashsurya (approved):

PR #658, opened 2022-01-28
DLPX-79367 td-agent service is not masked after upgrade

brad-lewis:
This is the second attempt to fix this issue.  

The failed fix integrated for DLPX-78736 rightly saw that td-agent needed to be masked in the fix_and_migrate_services() function in the upgrade script appliance-build/upgrad/upgrad-scripts/comm.sh.  However; the logic in that script does not work for new services.  It only masks existing services that are in the masked or disabled state.  

I added a separate lists of services that should always be masked after upgrade.   

Upgrade tests: http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/614/

### DIFF
--- a/upgrade/upgrade-scripts/common.sh
+++ b/upgrade/upgrade-scripts/common.sh
@@ -379,8 +379,13 @@ function mask_service() {
 	fi
 }
 
-function is_svc_masked_or_disabled() {
+function is_svc_new_or_masked_or_disabled() {
 	local svc="$1"
+
+	systemctl cat "$svc" &>/dev/null
+	if [ $? -eq 1 ]; then
+		return 0
+	fi
 
 	state=$(systemctl is-enabled "$svc")
 	if [[ "$state" == masked || "$state" == disabled ]]; then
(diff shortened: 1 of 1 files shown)


upgrade/upgrade-scripts/common.sh
```diff
@@ -431,6 +431,18 @@ function fix_and_migrate_services() {
 		fi
 	fi
 
+	#
+	# The services listed below are permanently disabled.  We need to
```
sebroy, line 435: I think this works for the use case we have, but still misses one; which is a new service in the category of "disabled by default but could be enabled dynamically". If we ever have a service like that, neither this code block nor the one below it would correctly mask it. I think to handle all four cases:
1. A new service that is unconditionally disabled
2. A new service that is disabled by default but dynamically enabled
3. An existing service that is unconditionally disabled
4. An existing service that is disabled by default but dynamically enabled

... we could change the `is_svc_masked_or_isabled` function to return true if the service doesn't exist at all (yet), right? Then, we wouldn't need two lists of services.

sebroy (changes requested):

sebroy (approved):

prakashsurya (approved):

brad-lewis:
The upgrade tests are passing:  http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/738/
(dx-integration tests failed with an artifactory issue).

PR #673, opened 2022-04-13
DLPX-80424 [Backport of DLPX-80287 to 6.0.15.0] remove unnecessary "recommended" packages from appliance

prakashsurya:


prakashsurya:
I don't plan to land this until 6.0.15 opens up; opening to get some feedback and trigger the automated build/tests.

I believe the build failure previously was due to 6.0/stage still having 74543b0c46d2f9a8f85e1cb33dc199d02d447bb5; that change isn't necessary now that we're running on 20.04 on our 6.0 release branches, so I've reverted that change and applied the original changes from #670 on top of that revert.

sebroy (approved):

grwilson (approved):

PR #676, opened 2022-04-26
DLPX-80873 Upgrade verification gets stuck when upgrading from 6.0.13…

prakashsurya:
….0 -> 6.0.13.1

George root-caused this to be a hang on an NFS mountpoint; "zfs list" being a victim of the hang. This change implements his workaround which he posted in the bug:

> One thing to note is that we can avoid the ZFS hang by not doing a list on the filesystem path but instead using the dataset name.

### DIFF
--- a/upgrade/upgrade-scripts/upgrade-container
+++ b/upgrade/upgrade-scripts/upgrade-container
@@ -831,12 +831,11 @@ function do_upgrade_container_not_in_place() {
 }
 
 function get_type() {
-	local root origin
+	local origin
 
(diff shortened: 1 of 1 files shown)


prakashsurya:
```
$ sudo /var/dlpx-update/latest/upgrade-container create in-place
Created symlink /etc/systemd/system/systemd-timesyncd.service → /dev/null.
delphix.gJ58qCj

$  sudo /var/dlpx-update/latest/upgrade-container get-type delphix.gJ58qCj
in-place
```

grwilson (approved):

don-brady (approved):

PR #678, opened 2022-05-31
Add azure

pcd1193182:
This PR adds the Azure CLI to our VMs. Thankfully, Ubuntu ships an azure-cli package that Azure doesn't list in their official installation instructions; this reduced the difficulty of adding the utilities to the image. They are added to the minimal common step unlike the awscli package because they are not needed for the bootstrap process, but we still want them on all systems.

http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/2004/console (upgrade failure appears to be an existing issue, I get it across a number of runs of different changes)

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
@@ -68,3 +68,8 @@
   delay: 30
   register: result
   until: result is succeeded
+
+- apt:
+    name:
+      - azure-cli
+    state: present


grwilson:
I think these types of packages are typically added by `delphix-platform` instead of appliance-build. For example: https://github.com/delphix/delphix-platform/blob/217e08a4ba889eb9c4a66a40a5d31cbfdb46a018/debian/rules#L100

grwilson:
I did notice that `awscli` is listed as a common package which seems wrong but shouldn't break anything.

PR #680, opened 2022-06-22
CP-8404 disable default telegraf service

grwilson:
As part of the integration of telegraf metrics into our product, we want to disable the default systemd service that is enabled by default.

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
@@ -82,6 +82,7 @@
     - nginx.service
     - postgresql.service
     - td-agent.service
+    - telegraf.service
 
 #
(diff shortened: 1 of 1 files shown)


grwilson:
ab-pre-push: http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/2204/

Cloned a VM from this build:
```
root@ip-10-110-234-111:/export/home/delphix# systemctl list-unit-files | grep telegraf
telegraf.service                       masked          enabled
```

sebroy (approved):
Did we ship a release with this service enabled? Do we need to handle upgrade?

grwilson:
> Did we ship a release with this service enabled? Do we need to handle upgrade?

Telegraf went into 6.0.15.0 so I want to backport this change so that we don't need to worry about upgrade.

prakashsurya (approved):

PR #689, opened 2022-08-02
TOOL-14783 Remove unused Python dependencies

ghost:
# Context

After the recent changes to the dcenter-gate we no longer use the
Python dependencies installed through apt. Instead we install all
Python dependencies in isolated Python environments.

# Problem

These OS level dependencies are now bloat and can be removed

# Solution

Remove them


prakashsurya (approved):

dbj-delphix (approved):
This is great; one request is a deployment section which I believe should be.

# Deployment plan

Once this build posts, we will upgrade/deploy on `dcol-dev` to ensure there is no breakage. If that goes smooth this change will be ingested by DC systems during their normal upgrade process.

PR #690, opened 2022-08-05
DLPX-82935 Generate mapping of canonical interface names to their corresponding MAC addresses.

naveen-singh-delphix:
# Context:
This PR is part of the the overall functionality to switch from kernel assigned network interface names to MAC address based interface names in AWS environments only. It handles creation of a mapping from the canonical interface names to their corresponding MAC addresses. This map is used by the osadmin service to update the interface names in the netplan file to MAC address based names.
This PR is directly related to the following two reviews which handle other parts of the overall fix
a. `https://github.com/delphix/delphix-platform/pull/385` -- Handles generation of MAC address based interface names on new engine installations
b. `http://reviews.delphix.com/r/83630` -- Enable the generation of new MAC address based interface names cloned instances in AWS

# Problem:
In AWS environment, changing instance types renders the new instance without any network configuration. This is due to the fact that the network interface names are assigned by the kernel based on the path on which they are enumerated during the boot, and this enumeration order can change while booting into the instance after changing its instance type. Depending on the enumeration path, the interface names can be different to the existing ones in the netplan file. The stale network interface cleaner thread in osadmin service ends up cleaning the existing interfaces because they do not match with the new interface names generated on the instance, thus leaving the instance without any network configuration.

# Solution:
This part of the fix addresses the conversion of old canonical network interface names in the netplan file to MAC address based names on the upgraded image. The fix provides an update script that is invoked as part of the PreExecStart stage of the systemd-networkd service and updates the network interface names to MAC address based names. 

# Testing
The fix has been manually tested for engine upgrade scenario.

# Implementation:
The implementation details of the solution.
-->


upgrade/upgrade-scripts/execute
```diff
@@ -28,6 +30,13 @@ function usage() {
 	exit 2
 }
 
+function generate_interface_to_mac_address_map() {
+  ip -o link | grep -v "lo:" | tr -d ":" | awk '{print $2,$17}' | \
```
sebroy, line 34: This feels fragile as it relies on the specific output format of the `ip` command and the order of the fields it emits. You might consider consuming the contents of the `/sys/class/net/<ifname>/address` instead. This is a kernel interface that is unlikely to change.

upgrade/upgrade-scripts/execute
```diff
@@ -459,6 +468,50 @@ else
 			"for '$ROOTFS_CONTAINER'"
 fi
 
+# In AWS environments, changing instance type could result in generation of
```
sebroy, line 471: Please make sure to run this through shfmt

upgrade/upgrade-scripts/execute
```diff
@@ -459,6 +468,50 @@ else
...
+# address based interface names. This makes this logic redundant for upgrades from
+# 6.0.16.0 and onwards. Hence, we must remove this logic as soon as we mark 6.0.16.0 as
+# the minimum version supported release.
+
+# The container does not have visibility into the complete network namespace of the
+# engine. Hence we need to create the map outside the container to ensure the presense
+# of all the interfaces in the map
+if ! systemd-detect-virt --container --quiet; then
+  platform_type=$(cat /var/lib/delphix-appliance/platform)
+  if [[ "$platform_type" == "aws" ]]; then
```
sebroy, line 500: can probably be simplified:
```
if ! systemd-detect-virt --container --quiet && [[ $(get-appliance-platform) == "aws" ]]; then
  // do your thing
fi
```

upgrade/upgrade-scripts/execute
```diff
@@ -459,6 +468,50 @@ else
...
+    # we always create the map file anew to avoid using any stale versions of the same
+    if [[ -e "$INTERFACE_TO_MACADRESS_MAP_FILE_PATH" ]]; then
+      rm "$INTERFACE_TO_MACADRESS_MAP_FILE_PATH" || \
+        die "Could not delete the existing interface to mac address map file"
+    fi
```
sebroy, line 505: It feels like this should be an implementation detail of the `generate_interface_to_mac_address_map` function.

sebroy (commented):
In the PR description, please provide the broader context (pointer to the parent review where this map file is being consumed).

upgrade/upgrade-scripts/execute
naveen-singh-delphix: This has been fixed.

upgrade/upgrade-scripts/execute
naveen-singh-delphix: Thanks. This looks good and has been incorporated.

upgrade/upgrade-scripts/execute
naveen-singh-delphix: Yes, makes sense. Moved this part into the definition of the `generate_interface_to_mac_address_map` function.

upgrade/upgrade-scripts/execute
naveen-singh-delphix: Yes, all the comments from shfmt have been incorporated.

naveen-singh-delphix:
The check `jenkinsci/appliance-build-orchestrator — jenkinsci/appliance-build-orchestrator` failed due to a segmentation fault logged here -- `http://selfservice.jenkins.delphix.com/job/appliance-build-stage0/job/pre-push/2720/console`. This segmentation fault is not related to this PR. I have manually test the changes and also the last manual ab-pre-push run was successful here -- `http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/3147/flowGraphTable`

upgrade/upgrade-scripts/execute
```diff
@@ -28,6 +30,24 @@ function usage() {
...
 }
 
+function generate_interface_to_mac_address_map() {
+	# we always create the map file anew to avoid using any stale versions of the same
+	if [[ -e "$INTERFACE_TO_MACADRESS_MAP_FILE_PATH" ]]; then
+		rm "$INTERFACE_TO_MACADRESS_MAP_FILE_PATH" ||
+			die "Could not delete the existing contents of the mac address map file"
+	fi
+	for i in $(find /sys/class/net/ -type l | awk -F "/" '{print $5}'); do
+		interface_name="$i"
```
sebroy, line 40: This can be simplified as:
```
for i in /sys/class/net/*; do
        ifname=$(basename $i)
        ...
done

File: upgrade/upgrade-scripts/execute [hunk 6]
```diff
@@ -28,6 +30,24 @@ function usage() {
...
+		if [ -e /sys/class/net/"$i"/address ] &&
+			grep -q -v -e "00:00:00:00:00:00" /sys/class/net/"$i"/address; then
```
sebroy, line 42: I don't think you need to worry about excluding the address for the `lo` interface given that this one won't appear in the netplan file. It'll be a harmless entry in your map file, and simplifies this code.

upgrade/upgrade-scripts/execute
```diff
@@ -28,6 +30,24 @@ function usage() {
...
+			mac_address=$(tr -d ":" </sys/class/net/"$i"/address)
```
sebroy, line 43: You might consider not doing that here given that the `:` is part of the MAC address. Instead, you could strip the `:` in the script that generates the right entries in `/etc/netplan`.

sebroy (changes requested):

upgrade/upgrade-scripts/execute
naveen-singh-delphix: Sure. Thanks for suggesting this.

upgrade/upgrade-scripts/execute
naveen-singh-delphix: Yes, I have moved the truncation to the update_netplan_to_mac_address script which is managed by the delphix-platform repository.

upgrade/upgrade-scripts/execute
naveen-singh-delphix: Sure.

upgrade/upgrade-scripts/execute
```diff
@@ -459,6 +478,42 @@ else
...
+# the minimum version supported release.
+
+# The container does not have visibility into the complete network namespace of the
+# engine. Hence we need to create the map outside the container to ensure the presense
+# of all the interfaces in the map
+if ! systemd-detect-virt --container --quiet && [[ $(get-appliance-platform) == "aws" ]]; then
+	# While booting back to the original image post rollback, we want to reuse the original
+	# netplan file that consisted of the canonical interface names generated by the kernel.
+	# Hence, we skip creating the map file in rollback execution path
+	if [[ "$1" != "rollback" ]]; then
```
prakashsurya, line 512: IIRC, I don't think this parameter will ever be "rollback".. perhaps you're thinking of the `upgrade` script.. but the `execute` script is only run on upgrade moving forward.. and it's not aware of a rollback event..

so, I think this conditional can be removed.. but please correct me if I'm wrong..

upgrade/upgrade-scripts/execute
```diff
@@ -459,6 +478,42 @@ else
...
+# The map creation logic does not apply to engines being upgraded from 6.0.16.0 onwards.
+# This is because the netplan file from 6.0.16.0 will by default be populated with mac
+# address based interface names. This makes this logic redundant for upgrades from
+# 6.0.16.0 and onwards. Hence, we must remove this logic as soon as we mark 6.0.16.0 as
+# the minimum version supported release.
+
+# The container does not have visibility into the complete network namespace of the
+# engine. Hence we need to create the map outside the container to ensure the presense
+# of all the interfaces in the map
+if ! systemd-detect-virt --container --quiet && [[ $(get-appliance-platform) == "aws" ]]; then
```
prakashsurya, line 508: will the `generate_interface_to_mac_address_map` actually fail if it's run in the container? just curious if we need to have it be conditional or not.. this check will only prevent us calling the function in the container.. it'll run outside the container no matter what..

upgrade/upgrade-scripts/execute
```diff
@@ -17,6 +17,8 @@
 
 . "${BASH_SOURCE%/*}/common.sh"
 
+INTERFACE_TO_MACADRESS_MAP_FILE_PATH="/etc/interface_to_macaddress_map.out"
```
prakashsurya, line 20: I don't feel strongly, but I think maybe `/var/lib/delphix-upgrade` or `/var/tmp/delphix-upgrade` or something might be better than putting this file in `/etc`.. maybe not a big deal, but this file isn't really a configuration file, but more like a file containing state needed for the `systemd-networkd` service..

upgrade/upgrade-scripts/execute
naveen-singh-delphix: I believe your point is correct. Some time back I was trying to figure out the correct script from which to generate this map, and the upgrade script was one of them.
I will remove this check and test both upgrade and rollback once.

upgrade/upgrade-scripts/execute
naveen-singh-delphix: In my tests, I was never able to get this map generated from inside the container because the container had no visibility into the net namespace.  In the implementation of `generate_interface_to_mac_address_map` function, we call the die() function at the end if the generated map file is empty. 
It is probably not a bad idea to make it explicit that this function should not be called in the container ? Let me know if you feel otherwise.

upgrade/upgrade-scripts/execute
naveen-singh-delphix: The check for "rollback" type has been removed.

upgrade/upgrade-scripts/execute
naveen-singh-delphix: Yes, thanks for suggesting this. I was not aware we had a /var/tmp/delphix-upgrade directory and it seems apt for this purpose.

upgrade/upgrade-scripts/execute
prakashsurya: ah, ok.. so it's a self-inflicted failure, since the file is empty.. I'm OK with that, just was curious.. we can keep it as-is.

prakashsurya (approved):

sebroy (approved):

PR #691, opened 2022-09-22
Create new variant for linux package mirror

sdimitro:
= Context

DevOps asked for a new variant which consists of a plain Ubuntu system with zfs installed similar to DCenter but without extraneous stuff (NFS, LDAP, specific DHCP configuration, etc..). The variant will be used to be the point of access of our linux package mirror backed by S3. The variant will be accessing S3 using the zfs-object-agent which will mount the package mirror data stored in S3 as an object-store pool. For more info refer to http://reviews.delphix.com/r/83379/.

= This Patch

Adds the new variant called `internal-package-mirror` which is an Ubuntu system with the following features:
- A `delphix` user with sudo priviledges
- Delphix kernel bits including ZFS and its userland components
- The devops-gate cloned under `/opt/devops-gate`

= Testing

In-Progress: http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/3364/


prakashsurya (approved):
LGTM

prakashsurya (approved):

sebroy (approved):

PR #699, opened 2022-11-03
TOOL-16451 Add new "buildserver" variant

prakashsurya:
This change adds a new "buildserver" variant, where the intention is to use this variant for building our software (packages, appliance, etc), rather than our current buildserver image which is based on Ubuntu.

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.buildserver-internal/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.buildserver-internal/tasks/main.yml
@@ -1,0 +1,24 @@
+#
+# Copyright 2022 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


prakashsurya:
```
± git-ab-pre-push -v buildserver-internal
Your build is at: http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/3691/
```

prakashsurya:
@sebroy @sdimitro I've tested this, and verified a few packages (e.g. delphix-platform, cloud-init) build on this new variant.. and appliance-build can work also, with #700 applied.. this change shouldn't impact anything on its own, and landing it should let me more easily test everything else (e.g. I can rely on the dcenter group to exist).. let me know what you think.

prakashsurya:
Also, let me know how you feel about creating a new variant, vs. using the `internal-minimal` variant and adding the additional packages that we'd need to that.

I think both are reasonable.. a new variant gives some (potential) isolation, where we could make changes to it, that we wouldn't necessarily want to make to the product.. but it comes at the cost of another variant, which increases build time (e.g. nighly needs to add a new variant) and storage costs (e.g. another dcenter image to maintain)..

sdimitro (approved):
I am all for it! Thanks for working on this!

P.S Probably doesn't matter but the copyrights are from 2019 in the new files, maybe we can update them?

sebroy (approved):

PR #700, opened 2022-11-03
TOOL-16450 Support running on a ZFS-root based buildserver

prakashsurya:


prakashsurya:
I deployed and executed this change on a `delphix-minimal` based buildserver..
```
$ sudo -E ./config/hooks/90-raw-disk-image.binary
...
+ zpool create -d -o version=28 -O canmount=off -O mountpoint=none -O compression=on -R /mnt/delphix.Vxu0ODD -t delphix.Vxu0ODD rpool /dev/mapper/loop2p1
...
+ zpool export delphix.Vxu0ODD
...

$ sudo zpool import
   pool: rpool
     id: 4804766101960319901
  state: ONLINE
status: The pool is formatted using a legacy on-disk version.
 action: The pool can be imported using its name or numeric identifier, though
        some features will not be available without an explicit 'zpool upgrade'.
 config:

        rpool       ONLINE
          loop2p1   ONLINE
```

I'll rely on the automated `ab-pre-push` check to verify it'll still work on our current Ubuntu buildserver.

sdimitro (approved):
Pending final testing

prakashsurya:
This passed ab-pre-push testing, meaning it was able to build the VM disk image on the Ubuntu buildserver, and also boot that disk image and run testing on the resultant VM.. so, I think we're good to go..

sebroy (approved):
Nice!

PR #421, opened 2020-02-07
Avoid removing "etab" file on upgrade

prakashsurya:

PR #716, opened 2023-03-09
DLPX-85081 Avoid removing "rmtab" file during nfs-utils upgrade

don-brady:
### Description
NFS maintains a list of active NFSv3 client mounts in the `rmtab` file.  We now rely on that file when determining if we can disable NFSv3 services, so we need that file to persist across upgrades to the `nfs-utils` package.

Note that a similar change was required in the past for the `etab` file (#421).

### Solution
To avoid this issue, we modify the `dpkg` state, such that dpkg will no longer associate the `rmtab` file with the `nfs-kernel-server` package. This way, when the old package is removed, the `rmtab` file will not be automatically manipulated by dpkg.  We do this prior to invoking `apt-get`  to ensure we remove the association before upgrading the package.

### Testing Done
http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/4744/

### Notes to Reviewers
This change is a follow on to a change in `nfs-utils` package where we removed `rmtab` file as a deliverable ([DLPX-84998](https://delphix.atlassian.net/browse/DLPX-84998)).

### DIFF
--- a/upgrade/upgrade-scripts/execute
+++ b/upgrade/upgrade-scripts/execute
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright 2018 Delphix
+# Copyright 2018, 2023 Delphix
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


upgrade/upgrade-scripts/execute
```diff
@@ -192,8 +192,13 @@ systemctl mask docker.service
 # release that suffers from this defect; we've dis-associated the "etab"
 # file from the "nfs-kernel-server" package in all later releases.
 #
+# As of release 7.0.0.0, we also need to preserve the "rmtab" file so all
```
sebroy: This looks good, but the comment right above says that the code block can be removed once we no longer support upgrading from 6.0. This is extremely confusing because we already do not support upgrading from 6.0.0.0. Two related thoughts:
1. Someone might come through and delete your new code thinking that it's no longer needed
2. The comment needs to be updated
3. The etab code can be removed since we don't support upgrading from 6.0.0.0

sebroy (changes requested):

upgrade/upgrade-scripts/execute
prakashsurya: 3. Yes, I think you're right..

    2. The comment comment might be more clear after etab references are removed, too

prakashsurya (approved):

upgrade/upgrade-scripts/execute
don-brady: Good point!  I updated the comments to be `rmtab` based and remove the `etab` code line.

sebroy (approved):

upgrade/upgrade-scripts/execute
```diff
@@ -160,40 +160,32 @@ fix_and_migrate_services
...
-#
 # We can safely remove this once running a version that no longer
-# supports upgrading from the 6.0.0.0 release, since that's the only
-# release that suffers from this defect; we've dis-associated the "etab"
-# file from the "nfs-kernel-server" package in all later releases.
+# supports upgrading from releases earlier than 9.0.0.0.
 #
 [[ -e /var/lib/dpkg/info/nfs-kernel-server.list ]] &&
-	sed -i '/\/var\/lib\/nfs\/etab/d' /var/lib/dpkg/info/nfs-kernel-server.list
+	sed -i '/\/var\/lib\/nfs\/rmtab/d' /var/lib/dpkg/info/nfs-kernel-server.list
```
ravi-cm, line 188: Don't we need the old change for etab as well? Why is it not required?

ravi-cm: Per offline discussion, `etab` change is not needed since we do not support upgrade from 6.0.0.0 release anymore.

ravi-cm (approved):
🚢 once upgrade test is clean.

PR #725, opened 2023-05-30
DLPX-85893 run upgrade "execute" script from separate service

prakashsurya:
### Problem

When an upgrade fails while running the `execute` script, we don't have
a mechanism for the product to automatically recover.

For example, we've had cases where the OOM killer will kick in, killing
the virtualization service while it's running the `execute` script,
causing the entire upgrade to fail and requiring support intervention on
the system to recover.

### Solution

The solution we've discussed is to run the `execute` portion of the
upgrade in a standalone service, such that it wouldn't be killed when
the OOM killer targets the virtualization service, and such that it
could be automatically restarted in the event that it dies for some
other reason (e.g. a kernel panic).

While this change doesn't provide the full solution, it's a step in that
direction. This change extends the `execute` script such that it performs
the additional steps that the `upgrade` script and/or the virtualization
service currently perform, after they run `execute`. This way, users
of this script now can run `execute` asynchronously, and be assured
everything needed to complete the upgrade will be performed.

The intention is for `execute` to be used by the virtualization service
as it is today, but eventually be called by a yet-to-be-created upgrade
service. Until the upgrade service is created, `execute` can be run via
`systemd-run`, such that it can be started by the virtualization service,
but decoupled from the virtualization service's cgroup limits (which will
help prevent OOMs).

### Why?

The `execute` script needed to be modified for a couple of reasons:

1. The current `execute` script doesn't handle things like updating the
   GRUB bootloader, nor does it handle restarting services. So, if we
   wanted to have a new service to "execute" the upgrade, the service
   would need to handle these additional steps. IMO, it makes sense to
   encapsulate all of these steps in `execute` like I've done here, so that
  when the new service is created, it can simply call this script.

2. Likewise, the virtualization service currently calls the `execute`
   script and needs to wait for it to complete, such that it can perform
   these additional tasks mentioned in the point above. By moving these
   additional tasks into this script, the virtualization service can now run
   the script asynchronously via `systemd-run` (since it no longer needs
   to wait for the script to complete); which allows us to solve the
   problem incrementally.

### Related Work

- https://github.com/delphix/dlpx-app-gate/pull/773

### Testing

- `ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/github/job/delphix/job/appliance-build/job/appliance-build-orchestrator/job/pre-push/211/)
- upgrade v12 to v12 via `FULL` is [here](http://selfservice.jenkins.delphix.com/job/blackbox-chained/4905/)
- upgrade v11 to v12 via `FULL` is [here](http://selfservice.jenkins.delphix.com/job/blackbox-chained/4906/)
- upgrade v12 to v12 via `DEFERRED` is [here](http://selfservice.jenkins.delphix.com/job/blackbox-chained/4907/)
- upgrade v11 to v12 via `DEFERRED` is [here](http://selfservice.jenkins.delphix.com/job/blackbox-chained/4908/)

### DIFF
--- a/upgrade/upgrade-scripts/upgrade
+++ b/upgrade/upgrade-scripts/upgrade
@@ -128,14 +128,26 @@ function upgrade_in_place() {
 	#
 	[[ "$DLPX_UPGRADE_DRY_RUN" == "true" ]] && return
 
+	CURRENT_VERSION=$(get_current_version) || die "failed to get version"
+
+	source_version_information
+
 	set_upgrade_property "UPGRADE_TYPE" "$UPGRADE_TYPE" ||
 		die "failed to set upgrade property 'UPGRADE_TYPE' to '$UPGRADE_TYPE'"
+
+	set_upgrade_property "UPGRADE_VERSION" "$VERSION" ||
+		die "failed to set upgrade property 'UPGRADE_VERSION' to '$VERSION'"
+
+	if [[ -n "$HOTFIX" ]]; then
+		set_upgrade_property "UPGRADE_HOTFIX" "$HOTFIX" ||
+			die "failed to set upgrade property 'UPGRADE_HOTFIX' to '$HOTFIX'"
+	fi
 
 	set_upgrade_property "UPGRADE_BASE_CONTAINER" \
 		"$(get_mounted_rootfs_container_name)" ||
 		die "failed to set upgrade property 'UPGRADE_BASE_CONTAINER'"
 
-	set_upgrade_property "UPGRADE_BASE_VERSION" "$(get_current_version)" ||
+	set_upgrade_property "UPGRADE_BASE_VERSION" "$CURRENT_VERSION" ||
 		die "failed to set upgrade property 'UPGRADE_BASE_VERSION'"
 
 	#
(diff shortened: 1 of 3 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/5627/)

sebroy (commented):
Any testing done? Pointers to PRs that this depends on?

upgrade/upgrade-scripts/upgrade
```diff
@@ -147,41 +159,45 @@ function upgrade_in_place() {
 	cleanup_in_place_upgrade
 	trap - EXIT
 
-	ROOTFS_CONTAINER="$(get_mounted_rootfs_container_name)"
-	[[ -n "$ROOTFS_CONTAINER" ]] ||
-		die "unable to determine currently mounted rootfs container"
+	if compare_versions "$CURRENT_VERSION" "ge" "12.0.0.0-0"; then
+		execute-in-place-upgrade -p
```
palash-gandhi: What's the reason to not ship this new script `execute-in-place-upgrade` as part of the upgrade image itself?

prakashsurya: good question..

I think the downside of if this script was in the upgrade image, is then, in order to execute it, you'd need to know which image/directory to use..

E.g. I was thinking we'd eventually have a new service that could run `execute-in-place-upgrade -p`, and be able to restart if needed (e.g. automatically run after a kernel panic, or OOM kill, or something like that).. if this service needed to execute a script from the unpacked upgrade image, it'd need to know which unpacked upgrade image to use (since we could have multiple unpacked at a time)..

So, I think we could do it that way.. but we'd still need some logic outside of the upgrade image, that we use to determine which image to use, when executing the script.. does that make sense?

And since most of this script is just determining which upgrade image to choose (e.g. most of the options are for choosing the image to use; `-l`, `-i`, `-v`, and `-p`), and then running some scripts from that image (i.e. `execute`, `rootfs-container`, etc), I chose to just have the whole script live outside of the upgrade image..

I'm not necessarily opposed to moving this into the upgrade scripts, though.. I think that's potentially a better solution.. and then, most of the script probably goes away.. and the yet-to-be-created upgrade service that'd run the script, could still "parse" the `upgrade.properties` file to determine which image to use when running the script..

palash-gandhi: Before your changes with these PRs, the `execute` script was always the single point of entry into actually performing the upgrade. The `execute` script was already shipped as part of the upgrade image. Would you consider extending the `execute` script to do all of the things that your new script is doing?
I think it's just simpler, solves the same problems you are trying to solve with these changes. I don't know if you have already thought of what this future service is supposed to do, but I'd presume it would simply be a thin wrapper to just run the `execute` script from the correct image....we could tell the service to use the upgrade properties API/file to determine the correct upgrade image, like you already mentioned.

prakashsurya: > Would you consider extending the execute script to do all of the things that your new script is doing?

Yea, that's what I was thinking as well, after thinking about your original question..

I just pushed another commit to try and do this.. let me know what you think.

With this update, I think I'll drop the `delphix-platform` change, and then the `dlpx-app-gate` change will need to be modified, to run `execute` instead of `execute-in-place-upgrade` (but still use `systemd-run`)..

> we could tell the service to use the upgrade properties API/file to determine the correct upgrade image

Yea.. so once we do the work to add the new service, I think this service could use:
```
EnvironmentFile=/var/dlpx-update/upgrade.properties
```
in its service definition.. and since we're now setting `UPGRADE_VERSION` in the file (with the changes in this PR, and the `dlpx-app-gate` PR), the service can use that field to know which `execute` script to run (i.e. which unpacked image to use).

Does that seem reasonable to you?

palash-gandhi: Yeah, that sounds great.

prakashsurya: Something we need to consider now, though.. is we'll have an "old" virtualization service calling the "new" execute script.. so, I'll have to verify this actually works correctly..

E.g. we'll have the existing `UpgradeManagerImpl` logic calling the new `execute` script.. where the existing virtualization logic does not expect `execute` to perform the reboot or restart the services.. I need to think about if this is going to cause any problems..

prakashsurya: I'm going to mark this as resolved.

sebroy (approved):

palash-gandhi (approved):

PR #726, opened 2023-06-03
DLPX-85904 Remove hardcoded git branch from appliance-build

palash-gandhi:
<details open>

This repo contains a file with a hardcoded value for the upstream/product branch. During release activities, 
the release team has a script (git-merge-repos) that replaces this value for each branch as part of the merge from 1 branch to the other. To automate these merges, this value should be removed and inferred at runtime such that the logic
to replace values during release merges can also be removed.
</details>

<details open>

Just like https://github.com/delphix/linux-pkg/pull/275, remove the hardcoded value and infer it at runtime only if needed.
</details>

<details open>

ab-pre-push with devops-gate changes: http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/5715/

ab-pre-push --no-tests: http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/5728/
</details>


prakashsurya (approved):
I think I'm OK with this..

it means, that in any case that isn't building from `git` and/or doesn't have an upstream branch set (e.g. perhaps `git bisect`), we'd need to set `UPSTREAM_PRODUCT_BRANCH`.. but I don't think those cases are something to concern ourselves with..

I'd probably prefer us to take this a step further, and _require_ `UPSTREAM_PRODUCT_BRANCH` to be set, when it's run from our Jenkins automation.. I think in all cases, we _should_ be setting it explicitly, so as some added assurance we could detect when running via Jenkins, and fail if it's not already set (rather than try to determine the value via the upstream branch)..

I'm good as-is; thanks, I think this is good cleanup.

scripts/aptly-repo-from-debs.sh
```diff
@@ -64,9 +64,7 @@ done
 # Download the delphix upgrade verification debian package, stored in the
 # combined-packages bundle.
 #
-AWS_S3_URI_COMBINED_PACKAGES=$(resolve_s3_uri \
-	"$AWS_S3_URI_COMBINED_PACKAGES" \
-	"linux-pkg/${UPSTREAM_BRANCH}/combine-packages/post-push/latest")
+AWS_S3_URI_COMBINED_PACKAGES=$(resolve_s3_uri "$AWS_S3_URI_COMBINED_PACKAGES")
```
prakashsurya, line 67: I'm confused, what is this line doing? seems like it's not actually doing anything?

What happens when `AWS_S3_URI_COMBINED_PACKAGES` is empty?

prakashsurya: oh.. nvm, I hadn't seen the changes to `resolve_s3_uri` yet..

scripts/common.sh
```diff
@@ -25,22 +25,25 @@ function die() {
...
 
 	if [[ -n "$pkg_uri" ]]; then
 		resolved_uri="$pkg_uri"
-	elif [[ -n "$latest_subprefix" ]]; then
+	else
+		#
+		# Set UPSTREAM_BRANCH. This will determine which version of the linux package
+		# mirror is used.
+		#
+		UPSTREAM_BRANCH=$(get_upstream_or_fail_if_unset) || exit 1
```
prakashsurya, line 38: is this refactor to avoid failing when run from Jenkins, in the case where it doesn't set `UPSTREAM_PRODUCT_BRANCH`? i.e. the hotfix rebuild case?

this way, we don't fail in the rebuild case, with an error saying we need to set `UPSTREAM_PRODUCT_BRANCH`?

if so, I guess I'm OK doing it this way.. the logic feels a big awkward, but I guess it's fine..

palash-gandhi: Exactly, the `UPSTREAM_PRODUCT_BRANCH` is not and does not need to be set for the hotfix case. Which is why I moved this into the non-hotfix code paths. I agree, it's an awkward API.... if a package URL is set, use that or require that a `UPSTREAM_PRODUCT_BRANCH` is set

prakashsurya: feel free to drop this then..

with that said, if it cleaned up the code, I might also be in favor of just not failing when `get_upstream_or_fail_if_unset` is called when running in the context of a Jenkins job.. e.g. Jenkins sets some specific environment variables that we could check to see if we're running in the context of a job.. like, perhaps the `CI=true` environment variable..

I'm good as is, though.. it's not a big deal..

prakashsurya (approved):

scripts/common.sh
palash-gandhi: The non-hotfix case i.e. the regular appliance-builds also run on jenkins where `get_upstream_or_fail_if_unset` should fail.

prakashsurya: That's true.. generally, it should set it.. but I guess it's possible for it to be empty, and it'd be nice to have an error message.. We could put that check in the Jenkins job, though..

But this is fine.. just leave it as-is.. not a big deal..

scripts/common.sh
```diff
@@ -135,3 +138,26 @@ function extract_debs_into_dir() {
 	find "$source_dir" -name '*.deb' -exec mv {} "$target_dir" \;
 	find "$source_dir" -name '*.ddeb' -exec mv {} "$target_dir" \;
 }
+
+function get_upstream_or_fail_if_unset() {
+	if [[ -z "$UPSTREAM_PRODUCT_BRANCH" ]]; then
+		local upstream_branch
+		upstream_branch="$(git rev-parse --abbrev-ref --symbolic-full-name "@{u}" | cut -d'/' -f2-)"
```
jorge-morse, line 145: Does this work in detached HEAD ? Jenkins  GitSCM checkouts result in a detached HEAD on the cloned repo so I think this won't work. I think it would fail with `HEAD: does not point to a branch`

Not sure if this is needed or if on Jenkins it will always be set explicitly via UPSTREAM_PRODUCT_BRANCH so we don't have to determine it. But thought I would mention it.

jorge-morse (approved):
LGTM

scripts/common.sh
palash-gandhi: The function only expects an upstream to be set in cases when the environment variable `UPSTREAM_PRODUCT_BRANCH` is not set. In all the cases that I have looked at, for example, [appliance-build](https://github.com/delphix/devops-gate/blob/main/jenkins/jobs/appliance_build.groovy#L141), the `UPSTREAM_PRODUCT_BRANCH` is always set. Building hotfixes off of older releases i.e. the hotfix-build job does not set this but does set another env variable which causes this block to be skipped.

PR #731, opened 2023-07-12
DLPX-85893 run upgrade "execute" script from separate service

prakashsurya:
### Problem

When an upgrade fails while running the `execute` script, we don't have
a mechanism for the product to automatically recover.

For example, we've had cases where the OOM killer will kick in, killing
the virtualization service while it's running the `execute` script,
causing the entire upgrade to fail and requiring support intervention on
the system to recover.

### Solution

The solution we've discussed is to run the `execute` portion of the
upgrade in a standalone service, such that it wouldn't be killed when
the OOM killer targets the virtualization service, and such that it
could be automatically restarted in the event that it dies for some
other reason (e.g. a kernel panic).

While this change doesn't provide the full solution, it's a step in that
direction. This change extends the `execute` script such that it performs
the additional steps that the `upgrade` script and/or the virtualization
service currently perform, after they run `execute`. This way, users
of this script now can run `execute` asynchronously, and be assured
everything needed to complete the upgrade will be performed.

The intention is for `execute` to be used by the virtualization service
as it is today, but eventually be called by a yet-to-be-created upgrade
service. Until the upgrade service is created, `execute` can be run via
`systemd-run`, such that it can be started by the virtualization service,
but decoupled from the virtualization service's cgroup limits (which will
help prevent OOMs).

### Why?

The `execute` script needed to be modified for a couple of reasons:

1. The current `execute` script doesn't handle things like updating the
   GRUB bootloader, nor does it handle restarting services. So, if we
   wanted to have a new service to "execute" the upgrade, the service
   would need to handle these additional steps. IMO, it makes sense to
   encapsulate all of these steps in `execute` like I've done here, so that
  when the new service is created, it can simply call this script.

2. Likewise, the virtualization service currently calls the `execute`
   script and needs to wait for it to complete, such that it can perform
   these additional tasks mentioned in the point above. By moving these
   additional tasks into this script, the virtualization service can now run
   the script asynchronously via `systemd-run` (since it no longer needs
   to wait for the script to complete); which allows us to solve the
   problem incrementally.

### Testing

- `git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/7225/)

### Related Work

- https://github.com/delphix/appliance-build/pull/725
- https://github.com/delphix/dlpx-app-gate/pull/773
- https://github.com/delphix/appliance-build/pull/728
- https://github.com/delphix/dlpx-app-gate/pull/850
- https://github.com/delphix/dlpx-app-gate/pull/990

### DIFF
--- a/upgrade/upgrade-scripts/upgrade
+++ b/upgrade/upgrade-scripts/upgrade
@@ -128,14 +128,26 @@ function upgrade_in_place() {
 	#
 	[[ "$DLPX_UPGRADE_DRY_RUN" == "true" ]] && return
 
+	CURRENT_VERSION=$(get_current_version) || die "failed to get version"
+
+	source_version_information
+
 	set_upgrade_property "UPGRADE_TYPE" "$UPGRADE_TYPE" ||
 		die "failed to set upgrade property 'UPGRADE_TYPE' to '$UPGRADE_TYPE'"
+
+	set_upgrade_property "UPGRADE_VERSION" "$VERSION" ||
+		die "failed to set upgrade property 'UPGRADE_VERSION' to '$VERSION'"
+
+	if [[ -n "$HOTFIX" ]]; then
+		set_upgrade_property "UPGRADE_HOTFIX" "$HOTFIX" ||
+			die "failed to set upgrade property 'UPGRADE_HOTFIX' to '$HOTFIX'"
+	fi
 
 	set_upgrade_property "UPGRADE_BASE_CONTAINER" \
 		"$(get_mounted_rootfs_container_name)" ||
 		die "failed to set upgrade property 'UPGRADE_BASE_CONTAINER'"
 
-	set_upgrade_property "UPGRADE_BASE_VERSION" "$(get_current_version)" ||
+	set_upgrade_property "UPGRADE_BASE_VERSION" "$CURRENT_VERSION" ||
 		die "failed to set upgrade property 'UPGRADE_BASE_VERSION'"
 
 	#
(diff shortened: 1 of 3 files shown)


prakashsurya:
@seb @palashgandhi I think now that the min version is set to 6.0.17, we can re-integrate this change.

prakashsurya:
Actually, nevermind.. the min version still is 6.0.15.. closing until we get to 6.0.17..

upgrade/upgrade-scripts/upgrade-container
```diff
@@ -815,18 +815,8 @@ function migrate_configuration() {
...
-	run "$IMAGE_PATH/execute" \
-		-p "$(cat /var/lib/delphix-appliance/platform)" ||
+	run "$IMAGE_PATH/execute" ||
 		die "'$IMAGE_PATH/execute' failed in '$CONTAINER'"
-
-	run /bin/systemctl reload delphix-platform ||
-		die "'systemctl reload delphix-platform' failed in '$CONTAINER'"
-
-	run /bin/systemctl restart delphix-platform ||
-		die "'systemctl restart delphix-platform' failed in '$CONTAINER'"
```
palash-gandhi, line 829: Do we not need these because `execute` will restart `delphix.target` one way or the other?

upgrade/upgrade-scripts/upgrade
```diff
@@ -128,14 +128,26 @@ function upgrade_in_place() {
...
+
+	source_version_information
+
 	set_upgrade_property "UPGRADE_TYPE" "$UPGRADE_TYPE" ||
 		die "failed to set upgrade property 'UPGRADE_TYPE' to '$UPGRADE_TYPE'"
 
+	set_upgrade_property "UPGRADE_VERSION" "$VERSION" ||
+		die "failed to set upgrade property 'UPGRADE_VERSION' to '$VERSION'"
+
+	if [[ -n "$HOTFIX" ]]; then
```
palash-gandhi, line 141: Where does this `HOTFIX` variable get set?

palash-gandhi (approved):

upgrade/upgrade-scripts/upgrade
prakashsurya: IIRC, it comes from the `version.info` file in the upgrade image.. and it might not be set if it's not an image for a hotfix.. e.g. if you look at a regular upgrade image, that value will be `HOTFIX=`.. i.e. empty.. but for a hotfix it'll be `HOTFIX=HF-1234`..

prakashsurya: this is similar to the `VERSION` field, which also comes from the same file..

upgrade/upgrade-scripts/upgrade-container
prakashsurya: yea, that's the idea.. we might want/need to reload delphix-platform though.. that's a bit awkward, but I don't think restarting delphix.target will do that for us.. might need to do something like that via the app-gate, too..

sebroy (approved):

upgrade/upgrade-scripts/upgrade-container
prakashsurya: been awhile since I wrote this code.. just wanted to follow up.. we (now) reload delphix-platform via the `execute` script via these changes.. so I think we're good.

PR #735, opened 2023-08-25
build for oem partners

grwilson:
This PR is a draft for an external build we intend to share with our partners. This is just a draft PR since we're only doing the review to ensure that the build is safe to distribute.

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
@@ -41,12 +41,14 @@
     comment: Delphix User
     home: /export/home/delphix
     password:
-      "{{ lookup('env', 'APPLIANCE_PASSWORD') | password_hash('sha512') }}"
+      "$6$oSDBhe8aMbPzcuGf$Zpnq.v8h6hdq/eYXs61oxvCK/b08mx.5uD61xoJq8SrQS\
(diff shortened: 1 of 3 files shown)


live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml
```diff
@@ -41,12 +41,14 @@
     comment: Delphix User
     home: /export/home/delphix
     password:
-      "{{ lookup('env', 'APPLIANCE_PASSWORD') | password_hash('sha512') }}"
+      "$6$oSDBhe8aMbPzcuGf$Zpnq.v8h6hdq/eYXs61oxvCK/b08mx.5uD61xoJq8SrQS\
```
prakashsurya, line 44: do we really want to make these public? they are now...

should we create a new environment variable instead? e.g. `APPLIANCE_PARTNER_PASSWORD`?

grwilson: I do not intent to merge this PR. This is just a special build that won't ever make it into our repo.

sebroy: I don't see the need to change this line of code at all, actually. If our goal is to do a single manual build, we can set APPLIANCE_PASSWORD to whatever we want in the environment of that build.

PR #713, opened 2023-02-03
DLPX-68852 systemd-run hung after running "execute" script during upgrade

prakashsurya:
Note, this is dependent on: https://github.com/delphix/delphix-platform/pull/409

PR #737, opened 2023-09-05
DLPX-87792 Fix timeout when starting an upgrade container

prakashsurya:
### Problem

Sometimes systemd goes out to lunch, and needs to be restarted. We attempted to fix this previously in #713 but seem to have missed a couple spots.. e.g. if systemd is unresponsive when we go to start a container, that start request may get dropped on the floor.

### Solution

This PR address the issue by restarting systemd just prior to starting the container. Also, this also fixes the issue when we terminate the container, too.

### DIFF
--- a/upgrade/upgrade-scripts/upgrade-container
+++ b/upgrade/upgrade-scripts/upgrade-container
@@ -462,6 +462,7 @@ function start() {
 		die "container '$CONTAINER' non-existent or mis-configured"
 	fi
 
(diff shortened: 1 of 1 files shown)


prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/6906/)

palash-gandhi (approved):

david-mendez1 (approved):

PR #738, opened 2023-09-14
TOOL-22969 Add mold to development engines

pcd1193182:
This PR adds the mold linker to our internal development environments, and configures it for use with our development builds. mold is significantly faster than the default GNU linker, which will improve build times significantly. This change has been tested in conjuction with linux-pkg changes, which can be found [here](https://github.com/delphix/linux-pkg/pull/301).

http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/7029/console

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.zfsonlinux-development/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.zfsonlinux-development/tasks/main.yml
@@ -45,6 +45,7 @@
       - llvm-12
       - lsscsi
       - mandoc
+      - mold
       - nfs-kernel-server
       - parted
       - pkg-config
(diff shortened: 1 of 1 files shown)


live-build/misc/ansible-roles/appliance-build.zfsonlinux-development/tasks/main.yml
```diff
@@ -76,3 +77,14 @@
...
     state: directory
     recurse: yes
+
+- file:
+    path: "/export/home/delphix/.cargo/"
+    state: directory
+    owner: delphix
+    group: staff
+- copy:
+    dest: "/export/home/delphix/.cargo/config.toml"
```
prakashsurya, line 87: Is there a system wide config we could use? I'm OK with this, just curious..

prakashsurya (approved):

live-build/misc/ansible-roles/appliance-build.zfsonlinux-development/tasks/main.yml
pcd1193182: I think the only option for a global configuration is to put something in `/.cargo/config.toml`, which feels weird to me, but apparently cargo checks all directories up the hierarchy to the root. I can change to that if people prefer it but putting it in the home directory is more idiomatic, I think.

prakashsurya: OK, yea this way sounds good.

sdimitro (approved):

PR #746, opened 2023-12-07
Add tab completion, path modifications

pcd1193182:
<details open>

Interacting with the systemctl command line tools can be cumbersome. Finding what services exist
and are active can require grepping the output of commands that are not intended for machine parsing.
Remembering and entering the full names of commands is potentially cumbersome if you don't use them
regularly. There are tab completion scripts that intend to solve these problems.

We also have similar problems with the zfs command line tools and the /opt/delphix/server/bin path.
These things add small friction to countless processes that can be easily reduced.
</details>

<details open>

We add tab completion scripts to the standard location for them on ubuntu systems, and then
source those scripts to get their functionality. We also add the /opt/delphix path to the
delphix user's PATH variable.
</details>

<details>

Spun up a development VM with these changes and verified that they all performed as expected.
</details>

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
@@ -119,3 +119,19 @@
   command: "systemctl disable {{ item }}"
   with_items:
     - docker.service
+
+- name: Add systemctl bash completion
+  copy:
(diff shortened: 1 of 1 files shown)


mwatson-delphix (approved):
lgtm. Thanks!

prakashsurya:
Shouldn't we put this in delphix-platform, so it works for existing engines after an upgrade? This will just do it for new installs..

PR #762, opened 2024-06-05
DLPX-91309 Add debug to grub for ESX and OCI platforms

prakashsurya:
Reapply #759 but only on the 2 platforms we've been able to reproduce internally on, and not on the platform that we found that it was problematic (AWS).

- `git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/8693/)


david-mendez1 (approved):
LGTM pending the ab-pre-push upgrade test and git checks.

palash-gandhi (approved):

PR #763, opened 2024-06-11
DLPX-91390 mount /proc during appliance-build when running ansible

prakashsurya:
### Problem

The `ca-certificates-java` package fails to install when it's installed inside of a chroot environment that does not have a `/proc` available within the chroot. As part of installing Java 17, that package is pulled in as a dependency, and fails to install.

### Solution

The solution taken in this PR, is to bind mount `/proc` (and a couple other related directories) into the chroot environment, so that when the `ca-certificates-java` package is installed as a dependency of the `delphix-masking` package, `/proc` will be available and allow the install to succeed.

### Testing

Verified this works in conjunction with https://github.com/delphix/dms-core-gate/pull/1295

### DIFF
--- a/live-build/config/hooks/configuration/80-build-configuration.binary
+++ b/live-build/config/hooks/configuration/80-build-configuration.binary
@@ -29,7 +29,16 @@ if [[ -e /etc/resolv.conf ]]; then
 	cp /etc/resolv.conf binary/etc/resolv.conf
 fi
 
+for dir in /proc /sys /dev; do
+	mount --rbind "$dir" "binary/${dir}"
+	mount --make-rslave "binary/${dir}"
+done
+
 ansible-playbook -vv -c chroot -i binary, ansible/playbook.yml
+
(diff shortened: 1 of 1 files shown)


sebroy:
This looks good, but I think we need a separate Jira issue to track this change. This PR doesn't "Install Java 17 as a runtime dependency", but rather is something that DLPX-91386 requires as a prerequisite task. We don't want DLPX-91386 to get closed as a result of merging this PR.

prakashsurya:
Sure, I can do that.. generally, I try and use the same ticket for all the changes needed to address that JIRA issue.. rather than different DLPXs for different repos, even though all the changes are for the "same bug" (just spread across different repos)..

I usually re-open the bug after a part of the solution is landed, and it get's auto-closed..

I've opened: https://delphix.atlassian.net/browse/DLPX-91390 as suggested..

pjang-dlpx (approved):

sebroy (approved):

PR #765, opened 2024-07-12
TOOL-25768 Add initial "internal-dct" variant

prakashsurya:
This PR adds a new "internal-dct" variant, such that we'll use this new variant to generate an appliance with DCT installed.

- `git ab-pre-push -v internal-dct --no-tests` is [here](http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/9111/)

### DIFF
--- a/live-build/variants/internal-dct/ansible/playbook.yml
+++ b/live-build/variants/internal-dct/ansible/playbook.yml
@@ -1,0 +1,27 @@
+#
+# Copyright 2024 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+---
+- hosts: all
+  gather_facts: no
+  vars:
+    ansible_python_interpreter: /usr/bin/python3
+  roles:
+    - appliance-build.minimal-common
+    - appliance-build.minimal-internal
+    - appliance-build.dct-common
+    - appliance-build.masking-common
+    - appliance-build.virtualization-common
--- a/live-build/misc/ansible-roles/appliance-build.dct-common/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.dct-common/tasks/main.yml
@@ -1,0 +1,20 @@
+#
+# Copyright 2024 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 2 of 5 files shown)


palash-gandhi (approved):
Lgtm, we do not use the recovery-environment so we likely do not want to add it here too, correct?

live-build/variants/internal-dct/ansible/playbook.yml
```diff
@@ -0,0 +1,25 @@
...
+
+---
+- hosts: all
+  gather_facts: no
+  vars:
+    ansible_python_interpreter: /usr/bin/python3
+  roles:
+    - appliance-build.minimal-common
+    - appliance-build.minimal-internal
+    - appliance-build.virtualization-common
```
palash-gandhi, line 27: Nit: Also, since this is an internal variant, do we want to run `virtualization-development` for developer conveniences like checking out some dev repos, including maybe `apigw-services`? @eyalkaspi-delphix thoughts?

palash-gandhi (approved):

nealquigley (approved):

live-build/variants/internal-dct/ansible/playbook.yml
prakashsurya: we could, I thinking I'd leave that up to the DCT team to decide.. it can be easy to add later

prakashsurya:
> Lgtm, we do not use the recovery-environment so we likely do not want to add it here too, correct?

yea, we don't even really use it on the engine appliance, so I think it's fine to leave it out.. but we could add it later if we want to.

prakashsurya:
```
sync-vm-6136-8a5cd044.dcol1 login: delphix
Password:
Welcome to Delphix 26.0.0.0-snapshot.20240715192822602
Last login: Mon Jul 15 22:47:39 UTC 2024 on ttyS0

Cannot make/remove an entry for the specified session
Traceback (most recent call last):
  File "/usr/bin/delphix-startup-screen", line 417, in <module>
    subprocess.run(login_cmd, shell=False, check=True)
  File "/usr/lib/python3.8/subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['/bin/login', '-p']' returned non-zero exit status 1.
```
getting this error when trying to log in, and the mgmt stack isn't coming up.. investigating that..

live-build/variants/internal-dct/ansible/playbook.yml
```diff
@@ -0,0 +1,27 @@
...
+---
+- hosts: all
+  gather_facts: no
+  vars:
+    ansible_python_interpreter: /usr/bin/python3
+  roles:
+    - appliance-build.minimal-common
+    - appliance-build.minimal-internal
+    - appliance-build.dct-common
+    - appliance-build.masking-common
```
eyalkaspi-delphix, line 26: Just curious: do you think anything would break if we don't include masking? On the one hand, not including masking means deviating from other engines, which could cause weird unexpected failures (what if some upgrade logic assumes masking is present? Maybe some support bundle collection logic makes this assumption too?), but on the other hand it's a little weird to include masking which we're not planning to use....

eyalkaspi-delphix (approved):

live-build/variants/internal-dct/ansible/playbook.yml
prakashsurya: I wasn't able to log in without masking, unfortunately.. probably a silly bug, since we don't usually have virt. without masking.. so, for now, I'd suggest we include masking to make forward progress, and look to remove it later if we want to.

live-build/variants/internal-dct/ansible/playbook.yml
eyalkaspi-delphix: Adding `apigw-services` wouldn't be very useful at this point (we'd need a different version of Java and other tooling to be able to build DCT on the engine directly).

Other dev repos seem useful though, I guess this is necessary to be able to run `git appliance-update` for instance on such engines?

eyalkaspi-delphix (approved):

PR #768, opened 2024-08-20
APIGW-9501 Appliance build to take as input DCT S3 URL and app version

eyalkaspi-delphix:
When building release DCT engine images, the appliance build needs two customization points which will be provided by Jenkins jobs (see https://github.com/delphix/devops-gate/pull/3340 for the corresponding devops-gate change):

1. AWS_S3_URI_DCT_PACKAGES, which was introduced in https://github.com/delphix/appliance-build/pull/765 but wasn't read correctly (see changes to download_dct_artifacts) indicates that the DCT debian packages should be pulled from that location instead of fetching the latest images (dev builds pull the latest image).
This location will be passed in by a new Jenkins job (coming soon) for building DCT engine images, which itself might eventually be trigered as a downstream of the dct release process (see https://github.com/delphix/devops-gate/pull/3222 for the changes producing the debian image for dct releases).

2. DELPHIX_APP_VERSION is a new property which will allow the mgmt stack to identify the DCT version (or Hyperscale version or other similar app running on the engine). The proposal is to set this version on the root dataset, mirroring the existing logic for the appliance version and hotfix version, which will then be read by the mgmt stack (see https://github.com/delphix/dlpx-app-gate/pull/2367/files#diff-0de62ad8e4f9b74e6357f3816ae7574bfb73cf5e2c92b09a8872f0bab4b4ab09R71 for a work in progress to read this version in the mgmt stack). Similarly, we are also including the version in upgrade bundle metadata.
The mgmt stack will use this version for 2 purposes:
 a. Display the DCT version in the UI (API, ...) as we'll communicate to customers in terms of DCT version (and not engine version)
 b. Allow upgrades using an upgrade image wich has the same appliance version, if the app version is different, as long as the app version is more recent (that is, the engine will allow upgrading to DCT engine with app version Y from DCT engine version with app X iff Y > X, even if both have been built on top of the same appliance version. For this situtation, today, the mgmt stack would only allow a "finished deferred" upgrade as it does not understand that some packages might be different). Incidentally, this will also prevent a non DCT engine to be upgraded to a DCT engine or vice versa as the app version check would detect a mismatch).
 
 <details open>

Ran appliance-build on my test Jenkins instance with new parameters set: http://ops.jenkins-eyal.dcol2.delphix.com/job/appliance-build-stage1/job/post-push/1/
Cloned the dcenter VM and verified the app version was set as a property of the ZFS dataset:
```
delphix@ip-10-110-206-157:~$ zfs get -Hpo value "com.delphix:app-version" "$(dirname "$(zfs list -Hpo name /)")"
22.0.0-eyal-testing
```
In the job logs, we can see the job fetching the DCT package from the parameter instead of latest
```
2:24:34  + download_dct_artifacts s3://snapshot-de-images/builds/jenkins-ops/dct/develop/post-push/2023 /var/tmp/jenkins/workspace/appliance-build-stage1/post-push/appliance-build/build/tmp.pkgs.p3jb8QEhN4/artifacts
12:24:34  + local dct_artifacts_uri=s3://snapshot-de-images/builds/jenkins-ops/dct/develop/post-push/2023
12:24:34  + local target_dir=/var/tmp/jenkins/workspace/appliance-build-stage1/post-push/appliance-build/build/tmp.pkgs.p3jb8QEhN4/artifacts
12:24:34  + [[ -z s3://snapshot-de-images/builds/jenkins-ops/dct/develop/post-push/2023 ]]
12:24:34  + mkdir /var/tmp/jenkins/workspace/appliance-build-stage1/post-push/appliance-build/build/tmp.pkgs.p3jb8QEhN4/artifacts/dct
12:24:34  + pushd /var/tmp/jenkins/workspace/appliance-build-stage1/post-push/appliance-build/build/tmp.pkgs.p3jb8QEhN4/artifacts/dct
12:24:34  + aws s3 sync s3://snapshot-de-images/builds/jenkins-ops/dct/develop/post-push/2023 .
12:24:34  Completed 122 Bytes/2.0 GiB (884 Bytes/s) with 2 file(s) remaining
12:24:34  download: s3://snapshot-de-images/builds/jenkins-ops/dct/develop/post-push/2023/SHA256SUMS to ./SHA256SUMS
...
12:24:45  download: s3://snapshot-de-images/builds/jenkins-ops/dct/develop/post-push/2023/delphix-dct_21.0.0-1723763188834.dd927eccac5e_amd64.deb to ./delphix-dct_21.0.0-1723763188834.dd927eccac5e_amd64.deb
12:26:24  delphix-dct_21.0.0-1723763188834.dd927eccac5e_amd64.deb: OK

```
Build info included
```
delphix@ip-10-110-206-157:~$ cat  /var/delphix-buildinfo/appliance-build/DELPHIX_APP_VERSION
22.0.0-eyal-testing
```

Upgrade image contains app version
```
delphix@ip-10-110-206-157:~$ tar -xvf internal-dct.upgrade.tar version.info
version.info
delphix@ip-10-110-206-157:~$ grep APP_VERSION version.info
APP_VERSION=22.0.0-eyal-testing
```
</details>

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -54,7 +54,7 @@ for (variant in allVariants) {
         for (envVar in ["DELPHIX_PLATFORMS",
                         "DELPHIX_HOTFIX_VERSION",
                         "DELPHIX_MINIMUM_VERSION",
-                        "DELPHIX_APP_VERSION",
+                        "DELPHIX_PACKAGED_APP_VERSION",
                         "AWS_S3_URI_LIVEBUILD_ARTIFACTS",
                         "AWS_S3_URI_COMBINED_PACKAGES"]) {
             inputs.property(envVar, System.getenv(envVar)).optional(true)
--- a/build.gradle
+++ b/build.gradle
@@ -54,6 +54,7 @@ for (variant in allVariants) {
         for (envVar in ["DELPHIX_PLATFORMS",
                         "DELPHIX_HOTFIX_VERSION",
                         "DELPHIX_MINIMUM_VERSION",
+                        "DELPHIX_APP_VERSION",
                         "AWS_S3_URI_LIVEBUILD_ARTIFACTS",
                         "AWS_S3_URI_COMBINED_PACKAGES"]) {
             inputs.property(envVar, System.getenv(envVar)).optional(true)
--- a/live-build/build.gradle
+++ b/live-build/build.gradle
@@ -101,7 +101,7 @@ for (variant in allVariants) {
 
                 for (envVar in ["APPLIANCE_PASSWORD",
                                 "DELPHIX_APPLIANCE_VERSION",
-                                "DELPHIX_APP_VERSION",
+                                "DELPHIX_PACKAGED_APP_VERSION",
                                 "DELPHIX_HOTFIX_VERSION",
                                 "DELPHIX_PACKAGE_MIRROR_MAIN",
                                 "DELPHIX_PACKAGE_MIRROR_SECONDARY",
--- a/live-build/build.gradle
+++ b/live-build/build.gradle
@@ -101,6 +101,7 @@ for (variant in allVariants) {
 
                 for (envVar in ["APPLIANCE_PASSWORD",
                                 "DELPHIX_APPLIANCE_VERSION",
+                                "DELPHIX_APP_VERSION",
                                 "DELPHIX_HOTFIX_VERSION",
                                 "DELPHIX_PACKAGE_MIRROR_MAIN",
                                 "DELPHIX_PACKAGE_MIRROR_SECONDARY",
--- a/live-build/config/hooks/vm-artifacts/90-raw-disk-image.binary
+++ b/live-build/config/hooks/vm-artifacts/90-raw-disk-image.binary
@@ -167,9 +167,9 @@ if [[ -n "$DELPHIX_HOTFIX_VERSION" ]]; then
 		"$FSNAME/ROOT/$FSNAME"
 fi
 
-if [[ -n "$DELPHIX_APP_VERSION" ]]; then
+if [[ -n "$DELPHIX_PACKAGED_APP_VERSION" ]]; then
 	zfs set \
-		"com.delphix:app-version=$DELPHIX_APP_VERSION" \
+		"com.delphix:packaged-app-version=$DELPHIX_PACKAGED_APP_VERSION" \
 		"$FSNAME/ROOT/$FSNAME"
 fi
 
--- a/live-build/config/hooks/vm-artifacts/90-raw-disk-image.binary
+++ b/live-build/config/hooks/vm-artifacts/90-raw-disk-image.binary
@@ -167,6 +167,12 @@ if [[ -n "$DELPHIX_HOTFIX_VERSION" ]]; then
 		"$FSNAME/ROOT/$FSNAME"
 fi
 
+if [[ -n "$DELPHIX_APP_VERSION" ]]; then
+	zfs set \
+		"com.delphix:app-version=$DELPHIX_APP_VERSION" \
+		"$FSNAME/ROOT/$FSNAME"
+fi
+
 if [[ -n "$DELPHIX_MINIMUM_VERSION" ]]; then
 	zfs set \
 		"com.delphix:minimum-version=$DELPHIX_MINIMUM_VERSION" \
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -144,7 +144,7 @@ function download_dct_artifacts() {
 	mkdir "$target_dir/dct"
 	pushd "$target_dir/dct" &>/dev/null || exit 1
 
-	aws s3 sync "$DCT_S3_DIR/$DCT_PACKAGE_PREFIX" .
+	aws s3 sync "$dct_artifacts_uri" .
 	sha256sum -c SHA256SUMS
 
 	popd &>/dev/null || exit 1
--- a/scripts/create-build-info-package.sh
+++ b/scripts/create-build-info-package.sh
@@ -94,6 +94,9 @@ check_env DELPHIX_APPLIANCE_VERSION
 echo "$DELPHIX_APPLIANCE_VERSION" >"$target/appliance-build/DELPHIX_APPLIANCE_VERSION"
 check_env AWS_S3_HOTFIX_METADATA
 echo "$AWS_S3_HOTFIX_METADATA" >"$target/appliance-build/HOTFIX_METDATA"
+if [[ -n "$DELPHIX_APP_VERSION" ]]; then
+	echo "$DELPHIX_APP_VERSION" >"$target/appliance-build/DELPHIX_APP_VERSION"
+fi
 
 #
 # Build the package
--- a/scripts/create-build-info-package.sh
+++ b/scripts/create-build-info-package.sh
@@ -94,8 +94,8 @@ check_env DELPHIX_APPLIANCE_VERSION
 echo "$DELPHIX_APPLIANCE_VERSION" >"$target/appliance-build/DELPHIX_APPLIANCE_VERSION"
 check_env AWS_S3_HOTFIX_METADATA
 echo "$AWS_S3_HOTFIX_METADATA" >"$target/appliance-build/HOTFIX_METDATA"
-if [[ -n "$DELPHIX_APP_VERSION" ]]; then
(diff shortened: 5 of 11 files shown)


prakashsurya (approved):
seems reasonable..

I'm a little conflicted with the term `APP` since I could imagine that being confused w.r.t. if that stands for "application" or "appliance".. but I don't have any suggestion otherwise..

this means we have the "current version", "hotfix version", and now "application version".. also "minimum version"..

do we plan to have a definition for what "application" means for any given appliance or upgrade image?

I'm OK with this as-is, I think it's just a little vague in my head w.r.t. how we plan to use "app version" moving forward.. e.g. will masking or virtualization ever use it? should they? if they did, and they wanted to have different version cadences but still be delivered on a single appliance (as they do now), what would the version be? or maybe it'd be w/e masking is, since we don't support running an engine using both..?

anyways, short of having a version for all the "app components" (e.g. virt, masking, dct, etc), I think using a single (and purposefully vague) "app version" makes sense.. we can perhaps adjust the meaning of it as we go..

while we're here, do we think we need/want to also add a "app name" or something too? e.g. to denote it's specifically a DCT image/engine.. I think that falls apart if we make that decision dynamically (e.g. like masking or virt).. since such an image could become _either_ masking, or virt.. DCT is special here, in that it's a different variant..

maybe we want to put the variant name into the image? since each VM image and upgrade image, no matter what application, will be unique.. e.g. `external-standard` could be either masking or virt. application.. perhaps DCT is added to that in the future.. whereas `external-dct` is always assumed to be specific for the DCT application..

with all that said, I'm more or less thinking out loud.. so, feel free to push as-is.. just some things that came to mind looking over the new `app-version` field, so I thought I'd write them down..

palash-gandhi (approved):
I agree with what Prakash said about naming. How about calling this `DELPHIX_CLOSED_APPLIANCE_VERSION`? It's general enough that it can be used with other products but also narrows it down to products that want to use the DE as a closed appliance and can be indicative of a field that is not required otherwise

eyalkaspi-delphix:
> seems reasonable..
> 
> I'm a little conflicted with the term `APP` since I could imagine that being confused w.r.t. if that stands for "application" or "appliance".. but I don't have any suggestion otherwise..
> 
> this means we have the "current version", "hotfix version", and now "application version".. also "minimum version"..
> 
> do we plan to have a definition for what "application" means for any given appliance or upgrade image?
> 
> I'm OK with this as-is, I think it's just a little vague in my head w.r.t. how we plan to use "app version" moving forward.. e.g. will masking or virtualization ever use it? should they? if they did, and they wanted to have different version cadences but still be delivered on a single appliance (as they do now), what would the version be? or maybe it'd be w/e masking is, since we don't support running an engine using both..?
> 
> anyways, short of having a version for all the "app components" (e.g. virt, masking, dct, etc), I think using a single (and purposefully vague) "app version" makes sense.. we can perhaps adjust the meaning of it as we go..
> 
> while we're here, do we think we need/want to also add a "app name" or something too? e.g. to denote it's specifically a DCT image/engine.. I think that falls apart if we make that decision dynamically (e.g. like masking or virt).. since such an image could become _either_ masking, or virt.. DCT is special here, in that it's a different variant..
> 
> maybe we want to put the variant name into the image? since each VM image and upgrade image, no matter what application, will be unique.. e.g. `external-standard` could be either masking or virt. application.. perhaps DCT is added to that in the future.. whereas `external-dct` is always assumed to be specific for the DCT application..
> 
> with all that said, I'm more or less thinking out loud.. so, feel free to push as-is.. just some things that came to mind looking over the new `app-version` field, so I thought I'd write them down..

Thank you, as always, for taking to review this.

I personally don't have a preference for the name of the variable. Palash is suggesting "Closed appliance version" in a previous comment (I am not sure why "closed appliance" better conveys what we are doing here, but agree that I haven't found a name which clarifies what this is). 

> do we plan to have a definition for what "application" means for any given appliance or upgrade image?
This is loosely defined as "The version of the application which is running on this engine, but of which versioning is distinct".

> while we're here, do we think we need/want to also add a "app name" or something too?
I could argue either way. One the one hand, it might make sense to add something so that the "app version" or whatever we call it unambiguously refers to a specific application name, but on the other hand I am a bit skeptical to add a property here which we wouldn't use anywhere. As impemented today, the variant is already set on the engine and upgrade image, and the mgmt stack has been modified to set the "engine type" based on the variant (so the engine UI, APIs, support bundles, ... clearly indicate it's a DCT engine). With this design, to some degree, the variant is now playing this role (among other things, and only for DCT engines). Would you suggest we add a new parameter here, and change the mgmt stack to ignore the variant and use it instead?

> maybe we want to put the variant name into the image? since each VM image and upgrade image, no matter what application, will be unique.. e.g. `external-standard` could be either masking or virt. application.. perhaps DCT is added to that in the future.. whereas `external-dct` is always assumed to be specific for the DCT application..

I am not following your suggestion?

> with all that said, I'm more or less thinking out loud.. so, feel free to push as-is.. just some things that came to mind looking over the new `app-version` field, so I thought I'd write them down..

I appreciate the conversation, I'll gather thoughts from the team to get to a conclusion.

eyalkaspi-delphix:
> > seems reasonable..
> > I'm a little conflicted with the term `APP` since I could imagine that being confused w.r.t. if that stands for "application" or "appliance".. but I don't have any suggestion otherwise..
> > this means we have the "current version", "hotfix version", and now "application version".. also "minimum version"..
> > do we plan to have a definition for what "application" means for any given appliance or upgrade image?
> > I'm OK with this as-is, I think it's just a little vague in my head w.r.t. how we plan to use "app version" moving forward.. e.g. will masking or virtualization ever use it? should they? if they did, and they wanted to have different version cadences but still be delivered on a single appliance (as they do now), what would the version be? or maybe it'd be w/e masking is, since we don't support running an engine using both..?
> > anyways, short of having a version for all the "app components" (e.g. virt, masking, dct, etc), I think using a single (and purposefully vague) "app version" makes sense.. we can perhaps adjust the meaning of it as we go..
> > while we're here, do we think we need/want to also add a "app name" or something too? e.g. to denote it's specifically a DCT image/engine.. I think that falls apart if we make that decision dynamically (e.g. like masking or virt).. since such an image could become _either_ masking, or virt.. DCT is special here, in that it's a different variant..
> > maybe we want to put the variant name into the image? since each VM image and upgrade image, no matter what application, will be unique.. e.g. `external-standard` could be either masking or virt. application.. perhaps DCT is added to that in the future.. whereas `external-dct` is always assumed to be specific for the DCT application..
> > with all that said, I'm more or less thinking out loud.. so, feel free to push as-is.. just some things that came to mind looking over the new `app-version` field, so I thought I'd write them down..
> 
> Thank you, as always, for taking to review this.
> 
> I personally don't have a preference for the name of the variable. Palash is suggesting "Closed appliance version" in a previous comment (I am not sure why "closed appliance" better conveys what we are doing here, but agree that I haven't found a name which clarifies what this is).
> 
> > do we plan to have a definition for what "application" means for any given appliance or upgrade image?
> > This is loosely defined as "The version of the application which is running on this engine, but of which versioning is distinct".
> 
> > while we're here, do we think we need/want to also add a "app name" or something too?
> > I could argue either way. One the one hand, it might make sense to add something so that the "app version" or whatever we call it unambiguously refers to a specific application name, but on the other hand I am a bit skeptical to add a property here which we wouldn't use anywhere. As impemented today, the variant is already set on the engine and upgrade image, and the mgmt stack has been modified to set the "engine type" based on the variant (so the engine UI, APIs, support bundles, ... clearly indicate it's a DCT engine). With this design, to some degree, the variant is now playing this role (among other things, and only for DCT engines). Would you suggest we add a new parameter here, and change the mgmt stack to ignore the variant and use it instead?
> 
> > maybe we want to put the variant name into the image? since each VM image and upgrade image, no matter what application, will be unique.. e.g. `external-standard` could be either masking or virt. application.. perhaps DCT is added to that in the future.. whereas `external-dct` is always assumed to be specific for the DCT application..
> 
> I am not following your suggestion?
> 
> > with all that said, I'm more or less thinking out loud.. so, feel free to push as-is.. just some things that came to mind looking over the new `app-version` field, so I thought I'd write them down..
> 
> I appreciate the conversation, I'll gather thoughts from the team to get to a conclusion.

After a conversation with the team, we've agreed upon "Packaged app version" as the term used throughout the build and mgmt stack to refer to the DCT version. I've updated this PR to reflect this. We've also concluded that a separate "packaged app name" wasn't beneficial at this point as we're using the variant and engine type for that already, and might revisit that in the future if the need arises.

PR #769, opened 2024-08-20
DLPX-92522 appliance-build changes for LTS upgrade to 24.04

palash-gandhi:
These are changes required to build a 24.04 based bootstrap VM that can
be used to build packages for the 24.04 LTS upgrade project. Here are the 
changes:
1. `awscli` is no longer available via apt & is supplied via snap. As such, we've decided to remove it from the product.
2. https://github.com/delphix/linux-pkg/pull/325 removes our reliance on an internal version of adoptopenjdk. appliance-build should install `openjdk-8-jdk-headless`. Note that appliance-build needs to install it because building packages other than virtualization (like make-jpkg, delphix-sso-app etc) also require a JDK installed. If it was only `virtualization` that needed it, it would have been appropriate to make the JDK a dependency of `virtualization`.
3. `virtualization` needs `docker.io` installed and running and requires the `delphix` user added to the `docker` group
4. The apt package `qemu` is replaced by `qemu-system`
5. I occasionally ran into build issues like [this](https://ops-jenkins.eng-tools-prd.aws.delphixcloud.com/job/linux-pkg/job/os-upgrade/job/build-package/job/delphix-rust/job/pre-push/10/console) where dpkg couldn't acquire a lock that was held by `unattended-upgrades` so I am adding logic to remove that package and disable that service on the buildserver.
6. Update references of "focal" to "noble"
7. I was also having trouble with ansible-check, so I chose to replace the internal gradle task with a Github action.

### Testing done
appliance-build jobs:
internal-dev: https://ops-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build/job/os-upgrade/job/pre-push/120/console
internal-buildserver: https://ops-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build/job/os-upgrade/job/pre-push/123/console
internal-dcenter: https://ops-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build/job/os-upgrade/job/pre-push/124/console


bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
```diff
@@ -36,15 +41,38 @@
       - livecd-rootfs
       - make
       - man
-      - openjdk-8-jre-headless
+      - openjdk-8-jdk
```
sebroy: should we be using the `-headless` package here?

live-build/misc/ansible-roles/appliance-build.buildserver-internal/tasks/main.yml
```diff
@@ -17,7 +17,7 @@
 ---
 - apt:
     name:
-      - adoptopenjdk-java8-jdk
+      - openjdk-8-jdk
```
sebroy: I think we want the headless JDK here. Also, why do we need this here and in the bootstrap ansible role? Same question goes for all of the other packages duplicated here. This is more of a question for @prakashsurya .

live-build/misc/ansible-roles/appliance-build.unittest-internal/tasks/main.yml
```diff
@@ -17,7 +17,7 @@
 ---
 - apt:
     name:
-      - adoptopenjdk-java8-jdk
+      - openjdk-8-jdk
```
sebroy: This should be headless.

live-build/variants/internal-dcenter/package-lists/dcenter.list.chroot
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 #
 
-adoptopenjdk-java8-jdk
+openjdk-8-jdk
```
sebroy: This should be headless.

sebroy (changes requested):

live-build/misc/ansible-roles/appliance-build.buildserver-internal/tasks/main.yml
prakashsurya: the bootstrap role configures the build server.. i.e. the VM we use to run appliance-build..

this role configures the build server VM image we use for various builds..

if the implication is that we should just configure the build server to have all necessary dependencies pre-installed, via here.. and then we don't need the bootstrap configuration.. that design could work, but this role would need to know the dependencies of all builds that'll occur on the image, to ensure they're all pre-installed..

I think the reason we pre-install java specifically here, is only due to our Jenkins infrastructure needed a Java runtime to connect the build VM as a Jenkins slave (i.e. the VM image needs to have ssh and java for Jenkins to connect with it).. it's not being installed due to it being needed to run any repo's build system.. the overlap between Jenkins and appliance-build's dependencies is just coincidental..

i.e. this role is just the bare minimum to enable the build VM to be used in Jenkins.. the bootstrap role is to install w/e build dependencies are needed to run w/e build is being performed..

hope that makes sense..?

live-build/variants/internal-dcenter/package-lists/dcenter.list.chroot
palash-gandhi: done

live-build/misc/ansible-roles/appliance-build.unittest-internal/tasks/main.yml
palash-gandhi: done

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
palash-gandhi: done

live-build/misc/ansible-roles/appliance-build.buildserver-internal/tasks/main.yml
palash-gandhi: @sebroy Can I mark this as resolved?

sebroy: Yep, this makes sense.

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
```diff
@@ -36,15 +40,46 @@
...
+    state: stopped
+  register: result_systemd_stop
+  failed_when: "result_systemd_stop is failed and 'Could not find the requested service' not in result_systemd_stop.msg"
+
+- name: Remove unattended-upgrades package
+  ansible.builtin.apt:
+    name:
+      - unattended-upgrades
+    state: absent
+    purge: true
```
sebroy, line 61: I'm wondering how this gets installed to begin with...

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
```diff
@@ -41,28 +44,38 @@
...
+- name: Update the number of nfs threads
+  ansible.builtin.lineinfile:
     path: /etc/default/nfs-kernel-server
     regexp: "{{ item.regexp }}"
     line: "{{ item.line }}"
   with_items:
-    - { regexp: '^RPCNFSDCOUNT=', line: 'RPCNFSDCOUNT=64' }
-    - { regexp: '^RPCMOUNTDOPTS=', line: 'RPCMOUNTDOPTS="--num-threads=5 --manage-gids"' }
+    - { regexp: "^RPCNFSDCOUNT=", line: "RPCNFSDCOUNT=64" }
+    - { regexp: "^RPCMOUNTDOPTS=", line: 'RPCMOUNTDOPTS="--num-threads=5 --manage-gids"' }
```
sebroy: 24.04 no longer references the `/etc/default/nfs-kernel-server` file.

sebroy (changes requested):

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
palash-gandhi: Marking as resolved per our Slack conversation:


> Palash Gandhi
>   1:10 PM
> It's not installed on the engine:
> delphix@localhost:~$ dpkg -l | grep unattended
> delphix@localhost:~$
> But it does exist within the bootstrap-24.04 image that is a temporary buildserver sourced from Ubuntu till we have an os-upgrade buildserver
> 
> 
> Sebastien Roy
>   1:12 PM
> Got it, so when we switch to our own build server, we'll no longer need that ansible task (or, it will be a no-op)? (edited) 
> 
> 
> Palash Gandhi
>   1:15 PM
> Correct, it will be a no-op, but I'd still rather leave it there for future LTS upgrades. Each LTS upgrade will likely begin with sourcing an Ubuntu image first & thus will require uninstall unattended-upgrades. This will minimize the manual tasks required to bootstrap future LTS upgrade projects..

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
palash-gandhi: Great catch, thanks.

palash-gandhi: Fixed

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
```diff
@@ -36,15 +34,46 @@
       - livecd-rootfs
       - make
       - man
-      - openjdk-8-jre-headless
+      - openjdk-8-jdk-headless
       - pigz
-      - qemu
+      # The VSDK plugin requires python3.11
```
prakashsurya, line 39: why does the VSDK requirements change the appliance-build's build requirements? IIRC, this configuration is to install the necessary packages to run appliance-build, not a DE's runtime requirements..

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
```diff
@@ -36,15 +34,46 @@
...
+      - python3.11
+      - qemu-system
       - rename
       - shellcheck
       - vim
       - zfsutils-linux
     state: present
+    update_cache: true
+
+- name: Stop unattended-upgrades systemd service in preparation for package removal
```
prakashsurya, line 49: why do we need this now, when we haven't needed it in the past?

scripts/run-live-build.sh
```diff
@@ -17,6 +17,8 @@
 
 . "${BASH_SOURCE%/*}/common.sh"
 
+check_env DELPHIX_PACKAGE_MIRROR_MAIN DELPHIX_PACKAGE_MIRROR_SECONDARY
```
prakashsurya, line 20: does this mean, when running manually, the user needs to pass in these environment variables now, when previously we didn't have to?

I'm not strictly opposed to this, but it does make it more difficult to run manually when that workflow is desired.

At one point, we could run it manually with no parameters and it'd auto-determine the correct values, and we keep regressing on that functionality.. which is fine, I guess, but makes it harder on some workflows that need to run this manually for debugging (e.g. build failures, etc.).

prakashsurya (approved):
looks reasonable.. curious about the changes to the bootstrap role, though..

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
palash-gandhi: This is not a runtime requirement. To build the virtualization package i.e. the gradle task of the app-gate requires python3.11

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
palash-gandhi: I used this role to bootstrap a vanilla Ubuntu image to use as a temporary build server till a DE-based buildserver is built.
From a recent slack conversation about this:

> Sebastien Roy
> [1:08 PM](https://perforce.slack.com/archives/C07HY8FURB4/p1736975290891139)
> [@Palash Gandhi](https://perforce.slack.com/team/U06SB34AGVA)
>  How was unattended-upgrades installed to begin with? Is that erroneously installed on the engine as well?
> 
> 
> Palash Gandhi
>   1:10 PM
> It's not installed on the engine:
> delphix@localhost:$ dpkg -l | grep unattended
> delphix@localhost:$
> But it does exist within the bootstrap-24.04 image that is a temporary buildserver sourced from Ubuntu till we have an os-upgrade buildserver
> 
> 
> Sebastien Roy
> [1:12 PM](https://perforce.slack.com/archives/C07HY8FURB4/p1736975574815659)
> Got it, so when we switch to our own build server, we'll no longer need that ansible task (or, it will be a no-op)? 
> 
> 
> Palash Gandhi
>   1:15 PM
> Correct, it will be a no-op, but I'd still rather leave it there for future LTS upgrades. Each LTS upgrade will likely begin with sourcing an Ubuntu image first & thus will require uninstall unattended-upgrades. This will minimize the manual tasks required to bootstrap future LTS upgrade projects..

scripts/run-live-build.sh
palash-gandhi: Yes, running manually will require these params to be set. There are now 2 sets of mirrors: v1 & v2. While we could infer this if an upstream branch is set and then decide which mirror to use, I'd rather perform this check here to ensure that this is set by all of our build automation (which already sets these afaict).
I have never had to use the manual workflow and hence assumed that it's not common.

live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
```diff
@@ -29,14 +29,13 @@
...
-    dest: /etc/default/nfs-kernel-server
-    force: yes
+- name: Create default NFS configuration
+  copy:
+    src: etc/
+    dest: /etc/
     owner: root
     group: root
-    state: link
+    mode: 0444
```
sebroy: looks like we need to clean this up?

palash-gandhi: removed

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
```diff
@@ -41,28 +44,39 @@
...
+- name: Mask named and dhcp services
+  ansible.builtin.systemd_service:
+    name: isc-dhcp-server.service
+    masked: true
 
-- command: systemctl mask named.service isc-dhcp-server.service isc-dhcp-server6.service
+- name: Mask named and dhcp services
+  ansible.builtin.systemd_service:
+    name: isc-dhcp-server6.service
+    masked: true
```
sebroy: Minor nit: Can you combine these three tasks by using the `with_items` ansible construct?

sebroy (approved):

live-build/misc/ansible-roles/appliance-build.dcenter/tasks/main.yml
palash-gandhi: done

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
prakashsurya: > To build the virtualization package i.e. the gradle task of the app-gate requires python3.11

we don't build the virtualization package as part of this repo, though.. is this needed to install the virtualization package?

PR #772, opened 2024-09-10
DLPX-86541 CIS: /dev/shm filesystem and mount options

dbshah12:
# Problem

Any user can upload and execute files inside the /dev/shm similar to the /tmp partition. Configuring /dev/shm allows an administrator to set the noexec option on the mount, making /dev/shm useless for an attacker to install executable code. It would also prevent an attacker from establishing a hardlink to a system setuid program and wait for it to be updated. Once the program was updated, the hardlink would be broken and the attacker would have his own copy of the program. If the program happened to have a security vulnerability, the attacker could continue to exploit the known flaw.

Screenshot 2024-09-11 at 8 11 59 PM

# Solution

- For new installations, during installation create a custom partition setup and specify a separate partition for /dev/shm.
- For systems that were previously installed, create a new partition and configure /etc/fstab as appropriate.

# Implementation

- For new installation did the below change in `live-build/config/hooks/vm-artifacts/90-raw-disk-image.binary`

```
cat <<-EOF >>"$DIRECTORY/etc/fstab"
	tmpfs /dev/shm tmpfs defaults,noexec,nosuid,nodev 0 0
EOF

mount -o remount /dev/shm
```

- For the upgrade did the below change in 

```
FSTAB_FILE_PATH="$DIRECTORY/etc/fstab"
DEV_SHM_LINE="tmpfs /dev/shm tmpfs defaults,noexec,nosuid,nodev 0 0"

# Check if the /dev/shm exec permission block line exists in the file
if ! grep -Fxq "$DEV_SHM_LINE" "$FSTAB_FILE_PATH"; then
	# If the /dev/shm exec permission block line doesn't exist, append it to the file
  echo "$DEV_SHM_LINE" >>"$FSTAB_FILE_PATH"
  mount -o remount /dev/shm
fi
```

# Testing

-

### DIFF
--- a/live-build/config/hooks/vm-artifacts/90-raw-disk-image.binary
+++ b/live-build/config/hooks/vm-artifacts/90-raw-disk-image.binary
@@ -298,6 +298,12 @@ cat <<-EOF >"$DIRECTORY/etc/fstab"
 	rpool/crashdump  /var/crash     zfs defaults,x-systemd.before=zfs-import-cache.service,x-systemd.before=kdump-tools.service 0 0
 EOF
 
+cat <<-EOF >>"$DIRECTORY/etc/fstab"
+	tmpfs /dev/shm tmpfs defaults,noexec,nosuid,nodev 0 0
+EOF
+
+mount -o remount /dev/shm
+
 #
 # Now we need to install the bootloader. In order to do that, we'll chroot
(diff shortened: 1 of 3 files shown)


dbshah12:
This change moved to delphix-platform in https://github.com/delphix/delphix-platform/pull/494

PR #774, opened 2024-09-19
DLPX-86532 & DLPX-86542 - CIS: /tmp filesystem and mount options & CIS: /var/tmp filesystem and mount options

dbshah12:
It depended on https://github.com/delphix/pipeline-shared/pull/306 - MERGED to pass test cases.

# Problem
- The '/var/tmp' directory and `/tmp` directory are world-writable directories used for temporary storage by all users and some applications. Since the directories '/var/tmp' and `/tmp` are used for temporary storage and are intended to be world-writable these directories, these directories should have a separate partition to avoid the risk of resource exhaustion and appropriate security options such as `nodev`, `noexec` and `nosuid` set for the partition to avoid security vulnerabilities.

# Solution
- Created a separate mount point for the `/tmp` and `/var/tmp` directories. And added them inside the `/etc/fstab` to restrict the `suid, dev and exec` permissions. 
- Also updated for the upgrade to handle the below 2 scenarios: 
    - Upgrade from a system that does not have a separate `/tmp` and `/var/tmp` filesystem: In this case, do nothing related to `/tmp` and `/var/tmp`; proceed with the upgrade as is. Even after the upgrade, the system will still not be a CIS compliant.
    - Upgrade from a system that does have a separate `/tmp` and `/var/tmp` filesystem: In this scenario, upgrade with the mount, ensuring that it remains CIS compliant.

# Testing
- Ran ab-pre-push https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build-orchestrator-pre-push/9600/ - PASSED
 
### Manual
- Created VM from the ab-pre push image and checked the mount for the `/tmp` and `/var/tmp`
Screenshot 2024-10-14 at 11 42 18 AM

- Checked the above two upgrade scenarios, and they are working as expected. For CIS-compliant systems, the upgrade maintains CIS compliance, and for non-CIS-compliant systems, the upgrade proceeds without changes to the compliance status.

### DIFF
--- a/upgrade/upgrade-scripts/upgrade-container
+++ b/upgrade/upgrade-scripts/upgrade-container
@@ -24,7 +24,10 @@ CONTAINER=
 
 TMP_DATASETS_EXIST=false
 
-# Check if both datasets exist
+# Verify whether both /tmp and /var/tmp ZFS datasets exist for the specified container.
+# If both datasets are present, the system is considered CIS compliant.
+# In such cases, handle the /tmp and /var/tmp mounts appropriately during the upgrade process.
+# To ensure this handling, set the TMP_DATASETS_EXIST variable to true.
 if zfs list "rpool/ROOT/$CONTAINER/tmp" >/dev/null 2>&1 && zfs list "rpool/ROOT/$CONTAINER/vartmp" >/dev/null 2>&1; then
 	TMP_DATASETS_EXIST=true
 fi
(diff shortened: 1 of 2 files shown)


upgrade/upgrade-scripts/upgrade-container
```diff
@@ -566,10 +613,10 @@ function destroy() {
 		#
 		# In order to safely perform the recursive destroy below,
 		# we need to ensure the filesystems are unmounted in the
-		# correct order. Since the "log", "data", and "home"
-		# datasets are mounted inside the "root" dataset, we need
-		# to unmount these two datasets before attempting to
-		# unmount (and/or destroy) the "root" dataset.
+		# correct order. Since the "log", "data", "home", "tmp" and
+		# "vartmp" datasets are mounted inside the "root" dataset,
+		# we need to unmount these two datasets before attempting
```
nealquigley: [nit] I guess we can remove "two" from this sentence?

nealquigley (commented):
This is not really my wheelhouse so definitely get someone with more experience with the build to review as well.

nealquigley (approved):

upgrade/upgrade-scripts/upgrade-container
dbshah12: Removed

sebroy (approved):

PR #784, opened 2025-01-23
DLPX-93224 Remove unnecessary tasks related to python3.11 added by previous os-upgrade PRs

palash-gandhi:
<details open>

In my previous PRs on this branch, I added `python3.11` as a dependency in the bootstrap role.
Prakash pointed out that this was not required since it is already listed as a build dependency of
the virtualization package and will be installed at build time. As such, I am undoing my previous changes
around python3.11 and the associated PPA.
Prakash also pointed out that unattended-upgrades should be stopped and removed before we install
any packages.

This PR also addresses some other bugs found during testing using the resulting DE-based buildserver:
1. appliance-build fails in [this task](https://github.com/delphix/appliance-build/blob/develop/live-build/misc/ansible-roles/appliance-build.minimal-common/tasks/main.yml#L33) while trying to hash a password. Apparently, ansible needs `passlib` to be installed. I tried installing it directly in the role but ansible still could not find the module. I tried installing it in the chroot env via [apt](https://ops-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build-stage1/job/post-push/194142/console) and also via [pip](https://ops-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build-stage1/job/post-push/194163/console), none of those worked. I finally resorted to installing it via bootstrap.
2. appliance-build fails when trying to run `aws` commands because it's not available in the path. It was found that `snapd` is not installed on DE-based buildservers and is in fact installed via a dependency by the bootstrap role. The bootstrap role then goes on to install `aws-cli` via snap but the Jenkins job continues to use the same shell which is why the path is not updated. When I tried to update the path and run the job, I realized that snap didn't allow the `aws` command to run because the home directory on a DE-based buildserver is not in `/home`:
```
delphix@ip-10-110-219-152:~$ aws
Sorry, home directories outside of /home needs configuration.
See https://forum.snapcraft.io/t/11209 for details.
```
I finally resorted to installing awscli via pip, system-wide.
</details>

<details open>

https://ops-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build/job/os-upgrade/job/pre-push/137/console
</details>

### DIFF
--- a/bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
+++ b/bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
@@ -51,6 +51,7 @@
       - openjdk-8-jdk-headless
       - pigz
       - python3-passlib
+      - python3-pip
       - qemu-system
       - rename
       - shellcheck
@@ -59,12 +60,13 @@
     state: present
     update_cache: true
 
-# aws-cli is distributed via snap on 24.04. While the package is not required by the product,
+# awscli is distributed via pip or snap on 24.04. While the package is not required by the product,
 # it is required by appliance-build itself.
-- name: Install aws-cli snap package
-  community.general.snap:
-    name: aws-cli
-    classic: true
+- name: Install awscli python package
+  ansible.builtin.pip:
+    name: awscli
+    break_system_packages: true
+  become: true
 
 - name: Load ZFS kernel module.
   community.general.modprobe:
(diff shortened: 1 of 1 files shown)


prakashsurya (approved):
LGTM, thanks.

sebroy (approved):

PR #795, opened 2025-02-25
DLPX-93479 ntpsec-rotate-stats.timer should follow ntpsec.service

palash-gandhi:
<details open>

The `ntpsec-rotate-stats.timer` is a systemd timer responsible for rotating NTP stats. It is currently enabled, even when NTP is disabled. We also never enable NTP stats, so we do not need this service. It fails to start when `ntpsec.service` is not enabled, causing noise.
</details>

<details open>

Disable it.
</details>

<details open>

ab-pre-push: https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build-orchestrator-pre-push/10462/
```
delphix@pg-mask-ntpsec-timer:~$ sudo systemctl status ntpsec-rotate-stats.timer
○ ntpsec-rotate-stats.timer
     Loaded: masked (Reason: Unit ntpsec-rotate-stats.timer is masked.)
     Active: inactive (dead)
    Trigger: n/a
```
</details>

### DIFF
--- a/live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
+++ b/live-build/misc/ansible-roles/appliance-build.virtualization-common/tasks/main.yml
@@ -81,6 +81,7 @@
     - delphix-fluentd.service
     - delphix-masking.service
     - ntpsec.service
+    - ntpsec-rotate-stats.timer
     - snmpd.service
 
 #
--- a/upgrade/upgrade-scripts/common.sh
+++ b/upgrade/upgrade-scripts/common.sh
@@ -492,6 +492,7 @@ function fix_and_migrate_services() {
 		nfs-mountd.service
 		nginx.service
 		ntpsec.service
+		ntpsec-rotate-stats.timer
 		postgresql.service
 		rpc-statd.service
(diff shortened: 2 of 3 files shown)


david-mendez1 (approved):

sebroy:
@palash-gandhi can you explain what this means?
> We also never enable NTP stats, so we do not need this service.

What are NTP stats, and how would we enable them if we used them? In other words, what files is this timer supposed to be rotating, and why doesn't our usage of NTP write to those?

sebroy:
I think I found the answer to my question. In the `ntp.conf` file, there's a `statsdir` property that has to be set (and the directory it's set to has to exist) for the daemon to log stats in that directory. If that isn't set (and I'm guessing our configuration doesn't have this set since you said that we never enable NTP stats), then no files need rotating.

sebroy (approved):

prakashsurya (commented):
don't we need to disable it on upgrade too? where is that done?

palash-gandhi:
> don't we need to disable it on upgrade too? where is that done?

Good catch. Added.

PR #815, opened 2025-05-06
DLPX-94130 LTS 24.04: upgrade from 20.0.0.0 -> 2025.2.0.0 -> os-upgrade branch failed

prakashsurya:
<details open>
The delphix-osadmin services removes the netplan configuration file when it runs inside an upgrade container, due to the container not having network interfaces. This leads to networking not being configured correctly when/if that upgrade container is promoted to a the host's rootfs.
</details>

<details open>
Disable the delphix-osadmin service from running in the upgrade container.
</details>

### DIFF
--- a/upgrade/upgrade-scripts/upgrade-container
+++ b/upgrade/upgrade-scripts/upgrade-container
@@ -234,10 +234,11 @@ function create_upgrade_container() {
 	# version of Delphix we're upgrading from).
 	#
 	# DLPX-78307 - Similarly, we must disable "google-guest-agent".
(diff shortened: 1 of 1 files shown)


sebroy (approved):
Looks good, and then we can follow that up with something native to that service that prevents it from doing the wrong thing without needing this tweak in the upgrade code..

prakashsurya:
> and then we can follow that up with something native to that service that prevents it from doing the wrong thing without needing this tweak in the upgrade code..

right.. we can add this directly to the osadmin's service file, rather than an override like this..

PR #822, opened 2025-06-02
DLPX-93758 upgrade incorrectly masking already masked service

mszak-perforce:
<details open>

The original logic would check to see if the service was masked, and then mask it. The correct logic is to only mask a service if it does not exist, or if it is disabled.
</details>

<details open>

I changed the shell script to implement the correct logic. However, in the case that the service registry is in a bad state, it's not 100% guaranteed that upgrade will somehow succeed in a way that leaves the system in a working state.
</details>

<details>

This change was tested in two ways:
1. The common.sh script was copied to the target machine, ingested into an active shell, and then fix_and_migrate_services was called. The list of services that are fixed/migrated is below, and they were in the correct state after running (note that systemd-timesyncd.service was not there at the start (i.e. not-found) and was in a masked state afterwards.
		delphix-fluentd.service
		delphix-masking.service
		fluentd.service
		nfs-mountd.service
		nginx.service
		ntpsec.service
		ntpsec-rotate-stats.timer
		postgresql.service
		rpc-statd.service
		rpcbind.service
		rpcbind.socket
		snmpd.service
		systemd-timesyncd.service
		telegraf.service
2. An upgrade image was created using ab-pre-push and the target system was upgraded. SSH'ing into the box found the services in the correct state.
</details>


upgrade/upgrade-scripts/common.sh
```diff
@@ -389,7 +389,16 @@ function mask_service() {
...
+        local svc="$1"
 
+        state=$(systemctl is-enabled "$svc")
+        if [[ "$state" == not-found || "$state" == disabled ]]; then
+                return 0
+        fi
+
+        return 1
+}
 function is_svc_new_or_masked_or_disabled() {
```
sebroy: It would be good to have a uniform nomenclature for these similar functions, and also to eliminate the duplicate code (the existing function can call your new function).

sebroy (changes requested):

upgrade/upgrade-scripts/common.sh
mszak-perforce: Removed dead code, which fixes nomenclature issue as well. Built upgrade image and tested upgrade.

sebroy (approved):

david-mendez1 (approved):

PR #824, opened 2025-06-03
TOOL-13488 appliance-build.bootstrap role fails to connect to ubuntu.com

sebroy:
Reconfigure apt early during the bootstrap phase of appliance-build to make sure that the build VM itself has apt sources pointing at the right package mirror URL. Because the internal-buildserver variant build also runs the `minimal-development` ansible role, I had to remove the ansible code in that role that enables cloud-init-based apt configuration (so that cloud-init doesn't reset the apt sources list on internal-buildserver variants). This then required that I do something explicit for `internal-dev` variants to configure their apt sources to point to the mirror. That's the change in `83-etc-apt-sources-list.binary`. While in that file, I noticed that the code that cleans up apt sources for external variants is wrong: It was only deleting files with a suffix of ".list", leaving behind a file named `ubuntu.sources`. I fixed the code to remove everything in `/etc/apt/sources.list.d/*`.

ab-pre-push: https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build-orchestrator-pre-push/11283/

### DIFF
--- a/bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
+++ b/bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
@@ -1,5 +1,5 @@
 #
-# Copyright 2018-2019 Delphix
+# Copyright 2018-2025 Delphix
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -15,6 +15,27 @@
 #
 
 ---
+- name: Remove default apt sources
+  ansible.builtin.command: bash -c 'rm -rf /etc/apt/sources.list.d/*'
+  changed_when: true
+
+- name: Configure apt sources
+  vars:
+    mirror_url: "{{ lookup('env', 'DELPHIX_PACKAGE_MIRROR_MAIN') }}"
+  block:
+    - name: Configure noble apt source
+      ansible.builtin.apt_repository:
+        repo: "deb {{ mirror_url }} noble main restricted multiverse universe"
+        state: present
+    - name: Configure noble-updates apt source
+      ansible.builtin.apt_repository:
+        repo: "deb {{ mirror_url }} noble-updates main restricted multiverse universe"
+        state: present
+    - name: Configure noble-security apt source
+      ansible.builtin.apt_repository:
+        repo: "deb {{ mirror_url }} noble-security main"
+        state: present
+
 - name: Stop unattended-upgrades systemd service in preparation for package removal
   ansible.builtin.systemd:
     name: unattended-upgrades
--- a/live-build/config/hooks/vm-artifacts/83-etc-apt-sources-list.binary
+++ b/live-build/config/hooks/vm-artifacts/83-etc-apt-sources-list.binary
@@ -1,6 +1,6 @@
 #!/bin/bash -eux
 #
-# Copyright 2018 Delphix
+# Copyright 2018-2025 Delphix
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
```diff
@@ -15,6 +15,27 @@
 #
 
 ---
+- name: Remove default apt sources
+  ansible.builtin.command: bash -c 'rm -rf /etc/apt/sources.list.d/*'
+  changed_when: true
+
+- name: Configure apt sources
+  vars:
+    mirror_url: "{{ lookup('env', 'DELPHIX_PACKAGE_MIRROR_MAIN') }}"
```
prakashsurya, line 24: hm, I'm not opposed to this.. but does this imply, even if the image is configured to point to our repository via the latest link.. we still need to do this for each build, since we don't want to use latest for the builds? e.g. linux-pkg still needs logic like this, appliance-build, and any other build system that uses our buildserver image?

I just don't like having each consumer of the buildserver image have to re-implement this logic.. rather than having the buildserver pre-configured to work correctly, "out of the box"..

sebroy: Good question. It looks like linux-pkg builds have their own existing logic (which isn't affected by this at all) to configure apt sources at build time: https://github.com/delphix/linux-pkg/blob/d9e4a3fad02212273d01435f61f2757f28781898/setup.sh#L28

In general, the philosophy here is that builds configure apt source URLs at build time (either passed in as parameters, or derived based on the branch being built), and the only systems that have a baked-in `/etc/apt/sources.list.d/` configuration are `internal-dev` variants. This change reinforces the philosophy that we've used up to now.

If we change the philosophy to be: configure the internal-buildserver variant with an `/etc/apt/sources.list.d/` configuration that points to the "latest" link for the branch it was created for, and simply use that configuration for all builds, then I think we'll have a bigger effort on our hands, and I'm not sure it will actually work. For example, if we do that, I don't think that builds can be repeatable (because "latest" moves forward while mirror snapshots do not). For example, if we want to re-build a prior release from a tag (e.g. to build a hotfix), and the internal-buildserver image snapshot in dcoa used to build that still points to "latest", then that build may produce different results, or at worst may not work

tonynguien (approved):
Makes sense to me.

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
prakashsurya: > For example, if we want to re-build a prior release from a tag (e.g. to build a hotfix), and the internal-buildserver image snapshot in dcoa used to build that still points to "latest", then that build may produce different results, or at worst may not work

I think I understand the concern, but that's no different than what we currently do for builds.. i.e. we use the "latest" link of w/e branch (or released version) we're building.. to ensure reproducible builds (e.g. for hotfixes), we "pin" that latest link based on the version we're building..

e.g. so to do this "right", we'd need the sources to be version specific.. and use the "latest" that corresponds to the released version, or unreleased branch, that corresponds to the image's version/branch..

we could perhaps even solve this via a runtime service, that checks the running appliances version (e.g. using `get-appliance-version`) and configures the sources list dynamically (and correctly) based on the result..

> In general, the philosophy here is that builds configure apt source URLs at build time

I'll go ahead and approve this, but still think this philosophy "bad".. and leads to bugs like this.. where we fix it in one build, but leave it broken in other builds (usually because we don't realize it's broken)..

IMO, it'd be much better if each unique build didn't have to worry configuring the sources correctly.. but as you mention, might be more work than you're trying to take on here..

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
```diff
@@ -15,6 +15,27 @@
...
+  block:
```
prakashsurya, line 25: why use `block` here? I don't think I've used that before, and don't see it used in any of the existing ansible tasks, so just curious.. is that needed so the variable is available within this block? and perhaps, only this block?

prakashsurya (approved):

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
sebroy: Exactly, the block enables the sharing of variables declared under the `vars` section.

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
sebroy: Yeah, I see what you mean, and I agree, except for the one remaining difference that I'm not sure about: For appliance-build, the jenkins code figures out what latest is pointing at first, and then passes in the resulting snapshot URL to the job. You and I had speculated that this was maybe done to prevent weird behavior if the latest link was changed mid-build. I still don't know if that's working around an actual problem with using "latest" for appliance-build, or if it's needless paranoia.

I'll land this, and then we can spend more time re-designing the whole approach given your concerns.

PR #827, opened 2025-06-30
TOOL-28181 internal-dev not configured with mirror apt sources

sebroy:
The fix for TOOL-1348 included logic to configure the internal-dev variant apt sources, but there’s a missing chroot in that configuration logic, which causes the configuration to apply to the build server instead of the VM being created.

This was tested by creating an internal-dev VM and verifying that `/etc/apt/sources.list.d/delphix.list` is present on it.

### DIFF
--- a/live-build/config/hooks/vm-artifacts/83-etc-apt-sources-list.binary
+++ b/live-build/config/hooks/vm-artifacts/83-etc-apt-sources-list.binary
@@ -33,9 +33,9 @@ chroot binary bash -c 'rm -f /etc/apt/sources.list /etc/apt/sources.list.d/*'
 # sources configuration that points to our internal Ubuntu mirror.
 #
 if [[ $APPLIANCE_VARIANT = "internal-dev" ]]; then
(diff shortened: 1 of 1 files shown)


mszak-perforce (approved):
Looks good.

prakashsurya (approved):

david-mendez1 (approved):

live-build/config/hooks/vm-artifacts/83-etc-apt-sources-list.binary
```diff
@@ -33,9 +33,9 @@ chroot binary bash -c 'rm -f /etc/apt/sources.list /etc/apt/sources.list.d/*'
 # sources configuration that points to our internal Ubuntu mirror.
 #
 if [[ $APPLIANCE_VARIANT = "internal-dev" ]]; then
-	cat <<-EOF >/etc/apt/sources.list.d/delphix.list
```
dbjwhs-perforce, line 36: nice find

dbjwhs-perforce (approved):
LGTM

PR #835, opened 2025-07-14
DLPX-93758 upgrade incorrectly masking already masked service

mszak-perforce:
<details open>

The original logic would check to see if the service was masked, and then mask it. The correct logic is to only mask a service if it does not exist, or if it is disabled.
</details>

<details open>

I changed the shell script to implement the correct logic. However, in the case that the service registry is in a bad state, it's not 100% guaranteed that upgrade will somehow succeed in a way that leaves the system in a working state.
</details>

<details>

1. An upgrade image was created using ab-pre-push and the target system running 2023.3 was upgraded. SSH'ing into the box found the services in the correct state.
2. An upgrade image was created using ab-pre-push and the target system running 2023.4 was upgraded. SSH'ing into the box found the services in the correct state.
  
</details>


upgrade/upgrade-scripts/common.sh
```diff
@@ -390,16 +391,26 @@ function mask_service() {
...
+function should_mask_service() {
 	local svc="$1"
 
-	systemctl cat "$svc" &>/dev/null
+	state=$(systemctl is-enabled "$svc")
 	if [ $? -eq 1 ]; then
-		return 0
+		# we should not mask the service if we have any issue determining
+		# whether it is enabled or not, except for an error returned on
+		# 2023-3 indicating the service is not yet found,
```
sebroy: There's no 2023-3 version. Did you mean 2025.3?

upgrade/upgrade-scripts/common.sh
```diff
@@ -390,16 +391,26 @@ function mask_service() {
...
+		if [[ "$state" == *"No such file or directory"* ]]; then
+			return 0
+		fi
+		return 1
 	fi
-
-	state=$(systemctl is-enabled "$svc")
-	if [[ "$state" == masked || "$state" == disabled ]]; then
+	if [[ "$state" == not-found || "$state" == disabled ]]; then
+		# only mask services that are not-found (2024-4 and greater) or
```
sebroy: There's no 2024-4. Did you mean 2025.4?

sebroy (approved):

upgrade/upgrade-scripts/common.sh
mszak-perforce: Fixed.

upgrade/upgrade-scripts/common.sh
mszak-perforce: Fixed

nealquigley (approved):

PR #842, opened 2025-07-30
DLPX-94639 [Jenkins] - JDK17 - Run unitTcsSuite on dlpx-internal-unittest-develop via dxtest

VenkatanadhanG:
[Jira](https://perforce.atlassian.net/browse/DLPX-94639) 

# Description
Adding JDK 17 while keeping JDK 8 as the default to support [CP-9424.](https://perforce.atlassian.net/browse/CP-9424)

The goal is to have JDK 17 installed in the dlpx-internal-unittest-develop images (**where delphix-virtualization pkg is not installed**), to enable running unit-tcs builds.

# Manual Testing
Following below steps with this code change:
- Created a unittcs image with this code change and validated the java version installed and default version of java.

Output:
```
vg ~ % ssh delphix@vg-jdk17-unittcs-1.dlpxdc.co                                                                                                          15/09/25 18:03:49
The authenticity of host 'vg-jdk17-unittcs-1.dlpxdc.co (10.110.221.74)' can't be established.
ED25519 key fingerprint is SHA256:t2VSi5NLfjL9t9cJCmKKEWC7X1y2qOruhYcV6YtTa9k.
This key is not known by any other names.
Are you sure you want to continue connecting (yes/no/[fingerprint])? yes
Warning: Permanently added 'vg-jdk17-unittcs-1.dlpxdc.co' (ED25519) to the list of known hosts.
(delphix@vg-jdk17-unittcs-1.dlpxdc.co) Password: 
delphix@ip-10-110-221-74:~$ 
delphix@ip-10-110-221-74:~$ cd /opt/
containerd/ gdb-python/ 
delphix@ip-10-110-221-74:~$ ls -l /opt/
total 3
drwx--x--x 4 root root 4 Sep 15 12:22 containerd
drwxr-xr-x 5 root root 5 Sep 15 09:46 gdb-python
delphix@ip-10-110-221-74:~$ 
delphix@ip-10-110-221-74:~$  sudo update-alternatives --config java 
There are 2 choices for the alternative java (providing /usr/bin/java).

  Selection    Path                                            Priority   Status
------------------------------------------------------------
  0            /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java   1800      auto mode
  1            /usr/lib/jvm/java-17-openjdk-amd64/bin/java      1711      manual mode
* 2            /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java   1800      manual mode

Press <enter> to keep the current choice[*], or type selection number: 
```


dbshah12 (approved):

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
```diff
@@ -99,3 +100,9 @@
     name: delphix
     groups: docker
     append: true
+
+- name: Set java to java-8-openjdk-amd64
+  community.general.alternatives:
+    name: java
+    path: /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java
+    priority: 1800
```
dbshah12: [nit] any reason behind defining priority as `1800`?

VenkatanadhanG: The priority should be higher than 1711, which is set up installing jdk17. So used 1800.

```
There are 2 choices for the alternative java (providing /usr/bin/java).

  Selection    Path                                            Priority   Status
------------------------------------------------------------
  0            /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java   1800      auto mode
* 1            /usr/lib/jvm/java-17-openjdk-amd64/bin/java      1711      manual mode
  2            /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java   1800      manual mode
```

suyash-jai (approved):

sebroy:
This code change does not result in installing Java 17 on any appliance variant. The bootstrap role is responsible for bootstrap configuration of the build server prior to a build starting. If we want Java 17 on appliances, then the right way to do that would be to add a dependency from whatever package relies on Java 17. For example, the delphix-masking package already depends on Java 17, so that package dependency is declared from dms-core-gate as part of building that package. The same should be done from dlpx-app-gate when building the delphix-virtualization package given that it's that package that consumes Java 17.

If the goal is actually to add a build dependency (if so, please make that clear in the Jira and/or PR description), then similarly, the build dependency should be declared in the context of that particular package's build, ideally (the dlpx-app-gate has a list of debian package build dependencies already that you can add to).

prakashsurya:
I agree with @sebroy .. at first glance, I don't think this PR is necessary, nor doing what was intended.

VenkatanadhanG:
We have made the required changes in delphix-virtualization package and merged it to `projects/jdk17-upgrade` https://github.com/delphix/dlpx-app-gate/pull/3386/files.

The intention of this PR is to get jdk17 on dlpx-internal-unittest-develop (delphix-virtualization pkg is not installed) image to run unit tcs, we had an option of adding it in `appliance-build/bootstrap/roles/appliance-build.bootstrap/tasks/main.yml` or specifically on `appliance-build/live-build/misc/ansible-roles/appliance-build.unittest-internal/tasks/main.yml`

Since applying in bootstrap applied to all image type, I choose to make these changes here.

Please let me know if I have misunderstood, or if there is an option to achieve it from elsewhere.

sebroy:
@VenkatanadhanG and I met 1-1 to discuss this. The correct way to add a package dependency for the unittest variant is to add it to the ansible role for that variant.

sebroy (approved):

ShibasishDelphix (approved):

dbshah12 (approved):

PR #849, opened 2025-09-06
DLPX-95341 Can't boot OCI images if CPU count is 8+

tonynguien:
<details open>

On OCI, VM images with `systemd-boot` can't boot properly when a VM is created with 8 or more OCPUs. The reported error is "?Error loading \vmlinuz: Not found" which is weird because the file is present in the EFI partition.

</details>

<details open>

I noticed OCI Ubuntu images use GRUB and don't have this issue.

So the approach here is to install `GRUB` as the primary boot loader for OCI images so our OCI images can function properly for all machine sizes. `systemd-boot` is still on the image but will not be active.

In the future, we can remove GRUB if/when systemd-boot works correctly with higher CPU count. Another solution could be to unify and use only GRUB if we don't see the GRUB corruption issue with EFI GRUB. 
</details>

<details open>

Confirmed
1) OCI VM with 9 CPUs boot successfully and
2) upgrading 2025.6 AWS VM (systemd-boot) to new code works as expected.

</details>

### DIFF
--- a/upgrade/upgrade-scripts/rootfs-container
+++ b/upgrade/upgrade-scripts/rootfs-container
@@ -133,12 +133,20 @@ function update_systemd_boot_entries() {
 	options root=ZFS=rpool/ROOT/${CONTAINER}/root ro $GRUB_CMDLINE_LINUX_DEFAULT
 	EOF
 
-	# Single user entry
+  #
+  # Single user entry
+  # Set console option, ESX and HyperV use VGA (tty0) and cloud platforms use serial (ttyS0).
+  #
+  if [[ "$APPLIANCE_PLATFORM" == "esx" || "$APPLIANCE_PLATFORM" == "hyperv" ]]; then
+    console_options="console=tty0"
+  else
+    console_options="console=ttyS0"
+  fi
 	cat <<-EOF >"${EFI_DIR}/loader/entries/delphix-single.conf"
 	title   Delphix Engine Single User mode
 	linux   /vmlinuz
 	initrd  /initrd.img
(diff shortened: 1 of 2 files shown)


upgrade/upgrade-scripts/rootfs-container
```diff
@@ -194,14 +202,29 @@ function set_bootfs_not_mounted() {
 		cp "/var/lib/machines/${CONTAINER}/boot/initrd.img" "$EFI_DIR"
 		cp "/var/lib/machines/${CONTAINER}/boot/vmlinuz" "$EFI_DIR"
 		chroot "/var/lib/machines/$CONTAINER" bootctl --esp-path="$EFI_DIR" update
+
+    # Update GRUB boot loader and config
+    if [[ -f "$EFI_DIR/EFI/GRUB/grubx64.efi" ]]; then
+      opts="--root-directory=/mnt --bootloader-id=GRUB --no-nvram"
+      if [[ "$PLATFORM" == "esx" ]] || [[ "$PLATFORM" == "oci" ]]; then
```
david-mendez1: The `--debug-install=all` should not be needed anymore for ESX since it was for debugging the GRUB corruption issue when using BIOS/GRUB.

I'm not certain if that debug flag can be removed altogether until we're certain that `free magic is broken` will not surface with EFI/GRUB.

david-mendez1 (approved):

upgrade/upgrade-scripts/rootfs-container
tonynguien: OK. I'll remove the ESX test.

sebroy (approved):

PR #858, opened 2026-02-25
DLPX-96649 Harden execute for background runs to avoid stopped job state

mszak-perforce:
<details open>

upgrade-scripts/execute can be run in the background, but it still inherits terminal stdin. In that mode, subcommands that touch stdin can receive job-control behavior from the terminal and the process can end up in a stopped state (for example via SIGTTIN), leaving upgrade execution hanging instead of progressing non-interactively.
</details>

<details open>

Treat execute as strictly non-interactive by detaching it from terminal input. The change redirects stdin to [null] at startup, so background runs do not depend on a controlling terminal and cannot be stopped due to stdin/job-control interaction. This is a minimal hardening change that preserves existing command behavior while making background execution reliable.
</details>

<details>

Me: Verified that upgrade worked on szak-upgrade-test-b.dlpxdc.co
AI:
- Reviewed the one-file change in upgrade/upgrade-scripts/execute and verified it introduces only stdin redirection (exec </dev/null) with explanatory comments.
- Verified the change is positioned early in script startup (after argument/root checks, before the main execution path), so all downstream subcommands inherit non-interactive stdin.
- Manually validated expected behavior for background usage: running [execute](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) in the background no longer enters a stopped job state caused by terminal input handling (SIGTTIN path).
- Confirmed no functional option/path changes beyond this hardening (no argument parsing or business-logic changes).

</details>

### DIFF
--- a/upgrade/upgrade-scripts/execute
+++ b/upgrade/upgrade-scripts/execute
@@ -136,6 +136,13 @@ shift $((OPTIND - 1))
 [[ $# -ne 0 ]] && usage "too many arguments specified"
 [[ "$EUID" -ne 0 ]] && die "must be run as root"
 
+#
+# This script is non-interactive. Redirect stdin from /dev/null so
+# background execution cannot be stopped by terminal input job-control
+# signals (e.g. SIGTTIN) from subcommands.
+#
+exec </dev/null
+
 if [[ -z "$opt_p" ]]; then
 	opt_p="$(get-appliance-platform)" ||
 		usage "platform must be specified"


upgrade/upgrade-scripts/execute
```diff
@@ -136,6 +136,13 @@ shift $((OPTIND - 1))
 [[ $# -ne 0 ]] && usage "too many arguments specified"
 [[ "$EUID" -ne 0 ]] && die "must be run as root"
 
+#
+# This script is non-interactive. Redirect stdin from /dev/null so
+# background execution cannot be stopped by terminal input job-control
+# signals (e.g. SIGTTIN) from subcommands.
+#
+exec </dev/null
```
prakashsurya, line 144: seems reasonable, do we want to adopt this behavior for all the upgrade scripts? e.g. put it in the `common.sh` file, perhaps next to where we configure stdout/stderr to go to `syslog` on line 65?

prakashsurya (approved):

nealquigley (approved):

upgrade/upgrade-scripts/execute
mszak-perforce: I think that's a good idea, but a backlog item. Let's take this win, and pick up that change next.

PR #867, opened 2026-05-12
DLPX-97204 Upgrade 2026.3→2026.4 fails: dpkg file conflict on /etc/modprobe.d/disable-algif_aead.conf between delphix-platform-aws and kmod 2ubuntu7.2

david-mendez1:
<details open>

Provide a clear description of the high-level problem you are trying to
solve. The problem statement should be written in terms of a specific
symptom that affects users or the business. The problem statement should
not be written in terms of the solution. If possible, include a minimal
reproducible example (MRE) with steps to reproduce, expected results,
and actual results.
</details>

<details open>

Provide a clear description of the high-level solution you have chosen.
If there were other possible solutions that you considered and rejected,
mention those along with the corresponding reasoning. Do not describe
implementation details when writing about the solution; these should go
into the implementation section instead.
</details>

<details>

Provide a clear description of how this change was tested. At minimum
this should include proof that a computer has executed the changed
lines. Ideally this should include an automated test or an explanation
as to why this pull request has no tests.
</details>

### DIFF
--- a/upgrade/upgrade-scripts/execute
+++ b/upgrade/upgrade-scripts/execute
@@ -300,6 +300,18 @@ DEBIAN_FRONTEND=noninteractive dpkg --configure -a \
 [[ -e /var/lib/dpkg/info/nfs-kernel-server.list ]] &&
 	sed -i '/\/var\/lib\/nfs\/rmtab/d' /var/lib/dpkg/info/nfs-kernel-server.list
 
+#
+# DLPX-97204: delphix-platform-$opt_p 1.0.0-delphix.2026.05.05.22.17 (2026.3)
+# ships /etc/modprobe.d/disable-algif_aead.conf, but kmod 31+20240202-2ubuntu7.2
+# in the 2026.4 upgrade pool also claims it as a conffile. Disown the file so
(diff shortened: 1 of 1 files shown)


upgrade/upgrade-scripts/execute
```diff
@@ -300,6 +300,18 @@ DEBIAN_FRONTEND=noninteractive dpkg --configure -a \
 [[ -e /var/lib/dpkg/info/nfs-kernel-server.list ]] &&
 	sed -i '/\/var\/lib\/nfs\/rmtab/d' /var/lib/dpkg/info/nfs-kernel-server.list
 
+#
+# DLPX-97204: delphix-platform-$opt_p 1.0.0-delphix.2026.05.05.22.17 (2026.3)
```
prakashsurya, line 304: I don't think this is the right way to solve this.. this issue is self-inflicted, correct?

I think we should either rename the file in delphix-platform, so we don't have the conflict, assuming we haven't released a version with that file already..

or perhaps, remove the file from delphix-platform on develop.. not sure if this would solve the issue, but we should try it out..

prakashsurya (changes requested):

PR #870, opened 2026-06-08
TOOL-30005 ucf: add masking-common role — required by Delphix mgmt JVM startup

malikparvez:
## Problem

The internal/external UCF (Unified Continuous Forensics / CDS) appliance
variants did not exist in `appliance-build`. UCF needs Delphix's closed
appliance configuration and admin console as a base, with the UCF app
layered on top. Standing this up surfaced several build failures:

- No `internal-ucf` / `external-ucf` variants existed.
- The UCF S3 package bucket lives in a different AWS account than
  Delphix's `linux-build-publish` IAM user, so `aws s3 sync` hit
  `AccessDenied` using the ambient credentials.
- The new variant dirs were missing the conventional
  `ansible/roles -> ../../../misc/ansible-roles` symlink, so the chroot
  `ansible-playbook` hook failed to resolve `appliance-build.minimal-common`.
- `virtualization-development` (clones `dlpx-app-gate` via git, which isn't
  in the chroot) and `qa-internal` are Delphix-product-specific and broke
  the UCF build.
- The Delphix management JVM fails to start without the `masking-common`
  role present.

## Solution

Add `internal-ucf` and `external-ucf` variants for CDS and the supporting
build plumbing:

- **New variants** `internal-ucf` / `external-ucf` (`build.gradle`,
  playbooks, `ansible/roles` symlink).
- **Cross-account S3** — honor optional `AWS_UCF_ACCESS_KEY_ID` /
  `AWS_UCF_SECRET_ACCESS_KEY` for the UCF `aws s3 sync` only; falls back to
  ambient credentials so DCT/Hyperscale paths are unaffected. Paired with a
  devops-gate PR exposing these via the `aws-ucf-build` Jenkins credential.
- **Role set** trimmed to what UCF actually needs: `minimal-common`,
  `minimal-internal`, `virtualization-common`, `ucf-common`; dropped
  `virtualization-development` and `qa-internal`.
- **`ucf-common`** role installs the CDS `puppetcloudops` package (uploaded
  to S3 by the CDS team's GitHub Actions, pulled into the apt repo by
  `build-ancillary-repository.sh`).
- **`masking-common`** role added — required by the Delphix mgmt JVM startup.

## Testing Done

- Built the `internal-ucf` variant locally; chroot `ansible-playbook` hook
  resolves all roles and completes.
- Verified UCF `aws s3 sync` uses the `AWS_UCF_*` credentials when set and
  falls back to ambient credentials when unset (DCT/Hyperscale paths
  unchanged).
- `check-shellcheck` and style checks pass; `.ansible-lint-ignore` baselined
  for the new ucf files consistent with every existing variant/role.

Tracking: PUPCLD-3705


scripts/common.sh
```diff
@@ -149,6 +149,50 @@ function download_dct_artifacts() {
...
+		AWS_ACCESS_KEY_ID="$AWS_UCF_ACCESS_KEY_ID" \
+		AWS_SECRET_ACCESS_KEY="$AWS_UCF_SECRET_ACCESS_KEY" \
+			aws s3 sync "$ucf_artifacts_uri" .
+	else
+		aws s3 sync "$ucf_artifacts_uri" .
+	fi
+
+	#
+	# UCF packages are produced by an external GitHub Actions workflow
+	# and may not ship a SHA256SUMS file; verify only if present.
```
prakashsurya: I think UCF really needs to ship a SHA256SUMS file (or _some_ way to verify the files we download).. the build needs to be able to verify the downloaded files that's it's putting into the image.. the checksum verification below shouldn't be conditional, or else we won't detect package corruption..

prakashsurya (changes requested):

scripts/common.sh
malikparvez: sure let me check on this, to get SHA256SUMS in our deb package

malikparvez: I have addressed this.

prakashsurya (approved):

prakashsurya:
please squash this to a single commit; we don't need/want to preserve all the interim commits in the repo history.

veijay-kushaal (approved):

PR #871, opened 2026-06-10
DLPX-97456 use non-recursive bind for /domain0 in upgrade container

prakashsurya:
<details open>

The upgrade container is created with `Bind=/domain0` in its `.nspawn` file, and `Bind=` is recursive by default, so every host mount underneath `/domain0` is inherited into the container's mount namespace. On a large engine that's tens of thousands of per-VDB mounts ([ESCL-6013](https://perforce.atlassian.net/browse/ESCL-6013): ~14.4k), all tracked as mount units by the container's systemd.

Starting with Ubuntu 24.04 (systemd 255), `daemon-reload` is quadratic in the number of units, so each of the ~136 postinst-triggered reloads during the in-container package phase takes 80-636s; the verify spends ~6.4h of a ~6.8h job inside reloads. Further, while PID 1 is reloading it doesn't service D-Bus, so bus clients time out after 25s; this can fail the verify outright (e.g. `systemctl enable` in a postinst script), and blocks support from accessing the container (`machinectl shell`, `systemd-run --machine`).
</details>

<details open>

The solution taken in this PR is to use the `norbind` option on the `/domain0` bind, so that only the top-level `/domain0` filesystem is mounted in the container, and the per-VDB child mounts are excluded. The `norbind` option dates to systemd v233, so it's supported by the nspawn on the oldest hosts we upgrade from (Ubuntu 20.04 / systemd 245).

Note the bind remains propagation-slave, so mounts created on the host after container start still propagate in; that's bounded by host churn during the verify window (tens of mounts, not 14k). It's not clear that's worth eliminating (e.g. by making the bind source private), so I've left it alone for now; we can always circle back if it proves to be a problem.

Also note, with this change, the per-VDB mountpoints appear as empty directories inside the container; whether the upgrade-verify JAR reads anything below those paths is being confirmed as part of [DLPX-97456](https://perforce.atlassian.net/browse/DLPX-97456).
</details>

<details open>

Pre-push (with upgrade testing enabled; `BUILD_ARTIFACTS=ALL`, `RUN_TESTS=True`, `TEST_UPGRADE_FROM_VERSION=2026.4.0.0`):
https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build-orchestrator-pre-push/14212/

Additionally, validated manually on a `dlpx-2025.5.0.2` VM with 14k filesystems mounted on the host under `/domain0`, creating/starting an upgrade container with these scripts, before and after this change:

```
                          Bind=/domain0 (before)     Bind=/domain0:/domain0:norbind (after)
mounts inside container:  14,051                     51
daemon-reload:            144.6s / 105.9s / 118.6s   0.45s / 0.28s / 0.27s
```

`/domain0` itself is still mounted and readable inside the container after the change. For comparison, a `dlpx-24.0.0.0` VM (systemd 245, where the reload cost is linear) reloads in ~6.4s at the same mount count, which matches the behavior seen in the customer's container journal prior to the in-container systemd upgrade.

Also verified through the upgrade API on a fresh `dlpx-release` `2026.3.0.0` engine: applied this change to `/var/dlpx-update/latest/upgrade-container`, created two child mounts under `/domain0` to stand in for per-VDB mounts (a `tmpfs` at `/domain0/norbind-test` holding a `MARKER` file, plus the existing `/domain0/repave/metadata/snapshot-based` zfs mount), then kicked off a `DEFERRED` upgrade to `2026.4.0.0` via `POST .../system/version/<ref>/apply`. The verify phase created the container with the `norbind` bind, so comparing the host and container mount namespaces .. e.g.

```
# host (/proc/1/mountinfo)              # upgrade container (/proc/<leader>/mountinfo)
/domain0                                /domain0
/domain0/repave/metadata/...            (repave + norbind-test children absent)
/domain0/norbind-test (tmpfs)
```

`/domain0` is mounted and readable in the container, while neither host child mount appears in the container's namespace; the `MARKER` file on the `tmpfs` is not visible inside the container either, confirming the child mount is excluded rather than just empty. The verify completed successfully (`status: VERIFIED`). The `DEFERRED` apply was then blocked only by the unrelated `upgrade.platform.minimum.memory` check (12GB required, the small test VM had ~7GB), which is independent of this change.
</details>

### DIFF
--- a/upgrade/upgrade-scripts/upgrade-container
+++ b/upgrade/upgrade-scripts/upgrade-container
@@ -363,9 +363,19 @@ function create_upgrade_container() {
 	# flyway migration) is compatible with the new version of the
 	# Virtualization application.
 	#
+	# We use "norbind" to make this a non-recursive bind mount, so
+	# that only the top-level "/domain0" filesystem is mounted in
+	# the container, rather than it and all of the (potentially,
+	# tens of thousands of) per-VDB filesystems mounted underneath
+	# it. The "daemon-reload" of systemd inside the container (e.g.
+	# run via the packaging scripts during upgrade) scales with the
+	# number of mounts in the container, so we must avoid these
+	# per-VDB mounts in the container, to avoid multi-minute reload
+	# times on systems with many VDBs; see DLPX-97456 for details.
+	#
 	if [[ -d "/domain0" ]]; then
 		cat >>"/etc/systemd/nspawn/$CONTAINER.nspawn" <<-EOF ||
-			Bind=/domain0
+			Bind=/domain0:/domain0:norbind
 		EOF
 			die "failed to add '/domain0' to container config file"
 	fi
--- a/upgrade/upgrade-scripts/upgrade-container
+++ b/upgrade/upgrade-scripts/upgrade-container
@@ -356,29 +356,28 @@ function create_upgrade_container() {
 		die "failed to add '$UPDATE_DIR' to container config file"
 
 	#
-	# We also need to make "/domain0" available from within the
-	# container, since that's required by the "upgrade-verify" JAR
-	# that's run later. This allows the Virtualization upgrade
-	# verification to verify the contents of it's database (e.g. run
-	# flyway migration) is compatible with the new version of the
-	# Virtualization application.
+	# Note that we intentionally do *not* bind-mount "/domain0" into
+	# the container. The "upgrade-verify" JAR needs to verify the
+	# contents of its database (e.g. run flyway migration) against the
+	# new version of the Virtualization application, but the root
+	# "/domain0" filesystem holds no files of its own; the data lives
+	# in child datasets (e.g. the MDS database) that are cloned and
+	# mounted into "/domain0" within the container's own mount
+	# namespace at verify time. Since "zfs" is available inside the
+	# container (via the "/dev/zfs" bind above), those mounts land in
+	# the container without "/domain0" needing to be exposed from the
+	# host at all.
 	#
-	# We use "norbind" to make this a non-recursive bind mount, so
-	# that only the top-level "/domain0" filesystem is mounted in
-	# the container, rather than it and all of the (potentially,
-	# tens of thousands of) per-VDB filesystems mounted underneath
-	# it. The "daemon-reload" of systemd inside the container (e.g.
-	# run via the packaging scripts during upgrade) scales with the
-	# number of mounts in the container, so we must avoid these
-	# per-VDB mounts in the container, to avoid multi-minute reload
-	# times on systems with many VDBs; see DLPX-97456 for details.
+	# A recursive "Bind=/domain0" was previously used, but "Bind=" is
+	# recursive by default, so every per-VDB filesystem mounted under
+	# "/domain0" on the host (tens of thousands on a large engine) was
+	# inherited into the container's mount namespace. systemd's
+	# "daemon-reload" scales with the number of mount units, so on
+	# Ubuntu 24.04 (systemd 255, where the reload became quadratic)
+	# the in-container package phase spent hours in reloads. Dropping
+	# the bind entirely avoids both the per-VDB mounts and the cost of
+	# the bind itself; see DLPX-97456 for details.
 	#
-	if [[ -d "/domain0" ]]; then
-		cat >>"/etc/systemd/nspawn/$CONTAINER.nspawn" <<-EOF ||
-			Bind=/domain0:/domain0:norbind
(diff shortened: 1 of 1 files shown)


sebroy:
I understand the problem, but I'm missing a bit of context.

Why bother bind-mounting domain0 into the container at all if we're not mounting any other filesystem inside domain0 (because afaik, there are actually no files that live in the root domain0 filesystem)? Does the upgrade code explicitly mount other things in there? Why couldn't it do that using a regular directory instead of a bind-mounted domain0?

prakashsurya:
@sebroy you're likely correct.. quick test, in a container, without that bind mount:
```
# /opt/delphix/server/bin/dx_manage_pg clone -s MDS-test
Cloning dataset...
done.

# mount | grep domain0
domain0/MDS-test on /domain0/MDS-test type zfs (rw,nosuid,nodev,noexec,relatime,xattr,noacl,casesensitive)
```

I'm not certain what creates the `/domain0` to begin with, but the container had it already when it was created.. so, this probably should work.. maybe the bind mount was never necessary to begin with..

prakashsurya:
oh, but it's failing to start:
```
root@ip-10-110-232-69:~# /opt/delphix/server/bin/dx_manage_pg start -s MDS-test
Starting postgres on 54441...
renaming /domain0/MDS-test/db/postmaster.pid to /domain0/MDS-test/db/postmaster.pid.original
Job for delphix-postgres@MDS-test.service failed because the control process exited with error code.
See "systemctl status delphix-postgres@MDS-test.service" and "journalctl -xeu delphix-postgres@MDS-test.service" for details.

dx_manage_pg: failed to start delphix-postgres@MDS-test
```

```
Jun 10 22:32:02 ip-10-110-232-69 svc-postgres[3406]: Configuring pg data /domain0/MDS-test/db true
Jun 10 22:32:02 ip-10-110-232-69 svc-postgres[3410]: cat: /domain0/MDS-test/db/PG_VERSION: Permission denied
Jun 10 22:32:02 ip-10-110-232-69 svc-postgres[3408]: dx_pg_upgrade: could not read /domain0/MDS-test/db/PG_VERSION
Jun 10 22:32:02 ip-10-110-232-69 svc-postgres[3406]: dx_pg_pre_start: failed to upgrade postgres
```

I can read that file manually, though:
```
# cat /domain0/MDS-test/db/PG_VERSION
14
```

hm..

prakashsurya:
looks like without the bind mount, the permissions on `/domain0` in the container aren't right:
```
# ls -l / | grep domain0
drwxr-x---    3 root root   3 Jun 10 22:47 domain0

# sudo -u postgres cat /domain0/MDS-tests/db/PG_VERSION
cat: /domain0/MDS-tests/db/PG_VERSION: Permission denied
```

we could probably solve this with a `chmod` somewhere..

prakashsurya:
@sebroy this passed upgrade testing.. see [here](https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/upgrade-testing/4487/)

should we move forward with this as-is, and investigate the removal of the bind mount in a follow up change.. or keep this pending until we can finish that investigation?

prakashsurya:
> should we move forward with this as-is

and I mean, the `norbind` option..

but also, thinking a bit more.. even if we "fix" this for upgrade, since this is a systemd regression w.r.t. the performance with so many mounts, I feel like we'll just end up hitting this issue on the host system, after the upgrade.. either during normal operation (if anything runs the reload command), or during their next upgrade, after verification.. i.e. when it upgrades packages on the host, and ends up running the reload command during the package phase..

perhaps we should view the bind mount as a "feature" w.r.t. upgrade verification.. rather than a bug? I'd be curious what you think, @sebroy

prakashsurya:
closing this for now, we can revisit if needed later.

prakashsurya:
picked this back up in https://github.com/delphix/appliance-build/pull/890, against develop.

the reasoning for closing this still holds (demotion collapses the mount count on the host too, rather than only hiding it from the container), but it leaves two gaps. demotion landed in `dlpx-app-gate` develop with the sweep gated off (`STORAGE.TIMEFLOW_DEMOTION.SWEEP_ENABLED` defaults to `false`), so nothing collapses the mount count unless an operator asks for it. and `upgrade-container` runs from the target version's upgrade image, so an engine sitting on 2025.5.0.2+HF-1582 loses this the moment it upgrades onto a develop-based release.

the change is unchanged from what's here; #890 is the same hunk, byte-identical to what ships in HF-1582 and HF-1588.

PR #869, opened 2026-06-01
DLPX-86523 CIS: /home filesystem and mount options

prakashsurya:
<details open>
<summary><h2> Problem </h2></summary>

The appliance's home ZFS dataset is mounted at `/export/home`, without the `nodev` or
`nosuid` mount options. CIS requires a dedicated `/home` filesystem carrying those
options, so the current layout fails the relevant controls for the `/home` partition
and its mount options.
</details>

<details open>
<summary><h2> Solution </h2></summary>

The solution taken in this PR is to mount the home dataset at `/home`, with
`nodev,nosuid`, on both fresh installs and in-place upgrades.

**Fresh installs / build:**
- Create and mount the home dataset at `/home` (raw-disk-image hook and
  upgrade-container template), with `nodev,nosuid` on the `/home` fstab entry.
- Update ansible roles and the FAQ for the new path, and set the `delphix` user's home
  to `/home/delphix`.

**In-place upgrades** (`upgrade/upgrade-scripts`):
- `migrate_export_home_to_home()` (`common.sh`) repoints the home dataset's
  `/etc/fstab` entry, and any affected `/etc/passwd` home directories, from
  `/export/home` to `/home`, then mounts `/home`. The pre-existing `/export/home`
  mount is left live until the next reboot, so running processes that hold it open are
  not disrupted and a busy unmount cannot fail the upgrade. The function self-guards on
  the fstab entry, so it is a no-op once migrated, and on systems that already use
  `/home`.
- `harden_home_mount_options()` (`common.sh`) ensures the `/home` fstab entry carries
  `nodev,nosuid`. It is idempotent.
- `execute` calls both functions before the package phase, so package maintainer
  scripts (e.g. the `delphix-platform` `postinst`) operate on the home dataset already
  mounted at `/home`, rather than on a `/home` directory on the root filesystem that
  the dataset mount would later shadow. The migration runs before the hardening, which
  depends on the `/home` entry the migration creates. Neither call is host-only. The
  functions self-guard and no-op inside upgrade containers.

**Dev images:**
- The `delphix-ldap` role (internal-dev / internal-dcenter only) no longer adds the
  `/home auto_home -nobrowse` autofs map. With the home dataset at `/home`, that
  automount reasserts `/home` on its timeout, shadowing the dataset and breaking
  home-directory access and SSH login. Customer variants never applied it.
</details>

<details open>
<summary><h2> Companion change </h2></summary>

This must land together with two companion PRs that move the remaining `/export/home`
references to `/home`:

[...]

PR #879, opened 2026-07-08
DLPX-86523 internal-buildserver: unhardened /home mount options

prakashsurya:
## Summary

The CIS `/home` hardening (DLPX-86523, #869) mounts the home ZFS dataset at
`/home` with `nodev,nosuid` for every variant. On the `internal-buildserver`
variant that breaks appliance-build: those builds run from a working tree under
`/home` and need to create device nodes and run setuid binaries there, which
`nodev`/`nosuid` block.

This mounts `/home` with plain `defaults` on the `internal-buildserver` variant
only. The raw-disk-image hook now computes a `HOME_MOUNT_OPTS` variable that
defaults to the hardened `defaults,nodev,nosuid` and is overridden to `defaults`
for that one variant, then interpolates it into the `/home` fstab entry. Every
other variant keeps `defaults,nodev,nosuid` byte-for-byte identical to today.

Scope is initial image generation only; the buildserver variant is never
upgraded, so no upgrade-scripts change is required. The variant is internal-only
and never shipped to customers, so relaxing the CIS hardening on it is
acceptable.

### ab-pre-push build #14495: IN PROGRESS

- Build: https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build-orchestrator-pre-push/14495/
- Tested: `appliance-build@995ce3e` (against `develop`)
- Variants: `internal-buildserver` (requested explicitly — the default variant set omits it, so this change is exercised directly) and `internal-qa` (covers the unchanged CIS-hardened default path)
- Status: running — result not yet available; re-run recording in completed mode once finished.

### DIFF
--- a/live-build/config/hooks/vm-artifacts/90-raw-disk-image.binary
+++ b/live-build/config/hooks/vm-artifacts/90-raw-disk-image.binary
@@ -313,8 +313,21 @@ rsync --info=stats3 -WaAX binary/* "$DIRECTORY/"
 # exist in /etc/fstab since that will ensure that it gets mounted
 # automatically whenever we boot into the crash kernel.
 #
+# The /home dataset is normally mounted with nodev and nosuid to satisfy
+# the CIS requirement for a hardened /home filesystem. The internal
+# buildserver variant is an exception: appliance-build runs its image and
+# upgrade builds out of a working tree under /home, and those builds create
+# device nodes and run setuid binaries there -- which the nodev and nosuid
+# options would block. So we mount its /home with the default (unhardened)
+# options. This variant is internal only and is never shipped to customers,
+# so relaxing CIS hardening here is acceptable.
+#
+HOME_MOUNT_OPTS="defaults,nodev,nosuid"
+if [[ "$APPLIANCE_VARIANT" == "internal-buildserver" ]]; then
+	HOME_MOUNT_OPTS="defaults"
+fi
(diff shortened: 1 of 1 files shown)


lyriclake (approved):

dbjwhs-perforce (approved):
LGTM

prakashsurya:
Tested appliance-build using the buildserver image built from this change, and it works.

Imported the `internal-buildserver` variant from this branch's ab-pre-push (`appliance-build-orchestrator-pre-push` #14495, appliance-build #6949) into a DCoA group, and confirmed the build area is no longer mounted `nodev`:

```
$ findmnt -no OPTIONS /home
rw,relatime,xattr,noacl,casesensitive
$ grep ' /home ' /etc/fstab
rpool/ROOT/.../home /home zfs defaults,x-systemd.before=zfs-import-cache.service 0 0
$ zfs get -Ho value devices,setuid <home-dataset>
on
on
```

Then ran a full `appliance-build/develop/post-push` (`internal-dev`, `aws`) with that image as the build host, and it completed successfully (~187 min), producing `internal-dev-aws.vmdk`.

For context: without this change, regenerating the develop buildserver produces an image whose build area (`/home`) is mounted `nodev`, which breaks live-build's `debootstrap` (`Cannot install into target ... mounted with noexec or nodev`) and fails the appliance build. That is the failure behind the buildserver being disabled in TOOL-28483; this change resolves it.

PR #889, opened 2026-08-03
DLPX-98271 Retire legacy autofs /home during upgrade so the home dataset stays mounted

david-mendez1:
## Problem

Upgrading to a release containing DLPX-86523 fails during the upgrade container's package phase on any engine that automounts `/home` via autofs (e.g. LDAP-integrated engines with a `/home auto_home` entry in `/etc/auto.master`). The `delphix-virtualization` postinst dies at `mkdir -m 700 /home/cli/.ssh` -> "No such file or directory", cascading to `delphix-masking`, then `execute` -> `upgrade-container:do_upgrade_container:700`.

DLPX-86523 mounts the home dataset at `/home` in the container (via fstab -> `home.mount`), but the container root is a clone of the source engine, which still runs `autofs.service` with a `/home auto_home` map. At container boot autofs takes over `/home` and deactivates the zfs mount; the container has no network so the LDAP map never resolves and `/home` is left empty. This also fails during upgrade *verify* (verify runs `execute` in a container). Standard CI/QA engines do not use autofs/LDAP home directories, which is why this was not caught.

## Solution

Add `disable_autofs_home()` (in `common.sh`), called from `execute` before `migrate_export_home_to_home` / the package phase:

- Remove the `/home` entry from `/etc/auto.master`(.d) -- durable: autofs will not reclaim `/home` even if restarted.
- **Stop** autofs rather than reloading it -- a running daemon asynchronously expires and re-unmounts `/home` ~1s after the remount, racing its cleanup thread.
- Mount the dataset via `systemctl start home.mount` so systemd owns/tracks it (a bare `mount /home` is left inactive and a later daemon-reload can unmount it), then assert `/home` is the zfs dataset.
- No-op on engines without an autofs `/home` (fresh installs, non-LDAP engines); other autofs maps (e.g. `/net`) are left intact.

## Testing Done

- Root-caused and reproduced on internal engines 10.43.1.30 and 10.43.1.31 (2026.1.0.0 -> 2026.4.0.0, both with LDAP-backed autofs `/home`); the container boot journal confirmed autofs deactivating `home.mount` followed by the `mkdir /home/cli/.ssh` failure.
- Prototyped on 10.43.1.30: an initial reload-based approach failed (a still-running autofs re-unmounted `/home` ~1s later); the committed stop-based approach got the upgrade past the previously-failing `delphix-virtualization`/`delphix-masking` configure step.
- Confirmed the change is a no-op on a non-autofs control engine (`dm-2025-5-0-2`): `disable_autofs_home` finds no `/home auto_home` entry and returns early.
- `bash -n` clean on `common.sh` and `execute`.
- Follow-ups before merge: a clean end-to-end upgrade on an autofs/LDAP engine (the prototype run stopped later at an unrelated memory precondition), and adding such an engine to the upgrade regression matrix to close the coverage gap.

### DIFF
--- a/upgrade/upgrade-scripts/common.sh
+++ b/upgrade/upgrade-scripts/common.sh
@@ -684,3 +684,73 @@ function harden_home_mount_options() {
 		die "failed to add nodev,nosuid to /home entry in /etc/fstab"
 	fi
 }
+
+#
+# Retire any legacy autofs management of /home before the home-dataset
+# migration and the package phase. Engines that automount home directories
+# via autofs (e.g. LDAP-integrated ones) carry a "/home auto_home" entry in
+# /etc/auto.master and run autofs.service. Inside the upgrade container that
+# service starts and takes over /home, deactivating the home-dataset mount
+# established from the container's fstab. The upgrade container has no
+# network, so the autofs map never resolves and /home is left empty; package
+# maintainer scripts that write under /home (e.g. the delphix-virtualization
+# postinst creating /home/cli/.ssh) then fail.
+#
+# Remove only the /home autofs mapping -- other maps (e.g. /net) are left
+# intact -- and ensure /home is the home dataset. This is a no-op on engines
+# that do not automount /home (fresh installs, non-LDAP engines), so it is
+# safe to call unconditionally.
+#
+function disable_autofs_home() {
+	local removed=false
+	local f
+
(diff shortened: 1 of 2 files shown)


david-mendez1:
This PR is only intended as a workaround if a develop variant hits this issue during upgrade.

PR #873, opened 2026-06-17
CP-13456 Add SBOM generation design spec

sebroy:
## Problem
Delphix appliance images ship without a machine-consumable SBOM. The only
bill-of-materials today is a home-grown CSV (devops-gate generate_bom.py /
combine_bom.py) built by brittle text-parsing of per-package copyright files —
license-centric, no PURLs, and not consumable by vulnerability scanners.

## Solution
Adds a design/feasibility spec (docs/specs/2026-06-15-sbom-generation-design.md)
for automating per-image CycloneDX 1.6 SBOM generation:
- Producers (linux-pkg packages; masking/virtualization/DCT/Hyperscale source
  builds) emit per-package CycloneDX SBOMs alongside their artifacts.
- appliance-build assembles one CycloneDX document per platform x variant image
  (3rd-party debs flat; 1st-party deep, with a Syft deb-scan fallback).
- Customer-facing publishing switches from the CSV to the CycloneDX file.
- The home-grown generate_bom/combine_bom path is decommissioned.

Spec only — no implementation in this PR.

## Testing Done
Documentation only. Spec reviewed iteratively; the ASCII architecture diagram
alignment was verified programmatically.

PR #892, opened 2026-08-10
DLPX-98618 appliance-build: generate per-image CycloneDX BOM via Syft chroot scan

justsanjeev:
<details open>
<summary><h2> Background </h2></summary>

Phase 1 ([DLPX-98618](https://perforce.atlassian.net/browse/DLPX-98618)) of the SBOM rollout under epic
[CP-13455](https://perforce.atlassian.net/browse/CP-13455), per the design spec
(`docs/specs/2026-06-15-sbom-generation-design.md`) and implementation plan
(`docs/specs/2026-06-23-sbom-generation-implementation-plan.md`) merged in #873.
</details>

<details open>

Delphix appliance images ship without a machine-consumable SBOM. The only bill-of-materials
today is a home-grown CSV (`devops-gate`'s `generate_bom.py`/`combine_bom.py`) built by
text-parsing per-package copyright files -- license-centric, no PURLs, and not consumable by
vulnerability scanners (Grype, Dependency-Track, etc.).
</details>

<details open>

`scripts/run-live-build.sh` now runs Syft against the populated `binary` rootfs (dpkg
cataloger only -- the sole source that can authoritatively attribute a file on disk to the
package that placed it there) right after the existing `binary/SHA256SUMS` check, emitting
`<variant>-<platform>.cdx.json` (CycloneDX 1.6 JSON), then validates it with
`cyclonedx-cli validate --fail-on-errors` so a malformed SBOM fails the build rather than
shipping silently. Wired into the existing Gradle output graph and S3 artifact sync -- no new
publishing plumbing.

Deep language-level cataloging (jars, npm modules, Rust crates bundled inside a larger
installed package) is intentionally out of scope here -- `dpkg` can't attribute that content
without risking misattribution. It's covered separately by per-package sidecar SBOMs produced
upstream in `linux-pkg` (CP-13465), merged in at a later stage (CP-13466).

Neither `syft` nor `cyclonedx-cli` are available through the Delphix package mirror --
`bootstrap/roles/appliance-build.bootstrap/tasks/main.yml` now provisions both explicitly,
pinned to an exact version and checksum-verified:
- **Syft `v1.46.0`** -- installed from the pinned `.deb` release asset via `apt`, checksum
  from Syft's own published `_checksums.txt`.
- **cyclonedx-cli `v0.32.0`** -- ships only as a bare per-platform binary upstream (no `.deb`,
  no official checksums file); checksum computed locally at pin time. Installed at
  `/usr/local/bin/cyclonedx-cli` -- that exact name matters, since it's the command
  `run-live-build.sh` invokes, not upstream's own `cyclonedx` naming.

Both pinned one release behind "latest" deliberately: Syft ships almost weekly so recency
alone isn't a meaningful signal, and cyclonedx-cli's `v0.33.0`/`v0.33.1` were a same-day
hotfix pair.

Purely additive -- the existing CSV BOM keeps working exactly as before. Decommissioning it
is a separate, later ticket (CP-13468), gated on this and the following phases being
validated first.
</details>

<details open>

⏺ Validation result: the SBOM matches the dpkg list exactly.                                                                                                                                                                       
                                                                                                                                                                                                                                   
  - dpkg_list.txt: 684 installed (ii) packages                                                                                                                                                                                     
  - internal-qa-aws.cdx.json: 685 components                                                                                                                                                                                       
  - 684/684 dpkg packages found in the SBOM, exact name+version match, 0 missing, 0 version mismatches                                                                                                                             
  - The 1 extra SBOM entry is ubuntu 24.04 (type: "operating-system") — that's Syft's OS/distro identification component, not a dpkg package, so it's expected and not a discrepancy.                                              
  
**Attachments** - 'dpk -l` output and sbom: 
[dpkg_list.txt](https://github.com/user-attachments/files/31789113/dpkg_list.txt)
[internal-qa-aws.cdx.json](https://github.com/user-attachments/files/31789116/internal-qa-aws.cdx.json)
                                                                                                                                                                                                                                 
    

**Jenkins Link:**
[appliance-build-orchestrator-pre-push](http://selfservice.dlpx98618.dcol1.delphix.com/job/appliance-build-orchestrator-pre-push/14/)  - 🟢 

[appliance-build-stage0](http://selfservice.dlpx98618.dcol1.delphix.com/job/appliance-build-stage0/job/pre-push/10/) - 🟢  The Cyclone DX Sbom artifacts are attached with Jenkins Job.

<img width="890" height="463" alt="Screenshot 2026-09-03 at 5 30 40 PM" src="https://github.com/user-attachments/assets/ab1f9b01-d877-4ec1-ae98-fde55b28330f" />



S3 Cyclonedx sbom artifats
````
sanjeev.rohilla@APAC-SANJEEV~/syft  add-packaging-code % aws s3 ls s3://dev-de-images/builds/jenkins-selfservice.sanjeev.rohila/appliance-build/develop/pre-push/10/                
                           PRE external-standard-aws/
                           PRE hostchecker2/
                           PRE input-artifacts/
                           PRE internal-dev-aws/
                           PRE internal-qa-aws/
                           PRE metadata/
                           PRE toolkit-devkit/
                           PRE upgrade-artifacts/
sanjeev.rohilla@APAC-SANJEEV~/syft  add-packaging-code % aws s3 ls s3://dev-de-images/builds/jenkins-selfservice.sanjeev.rohila/appliance-build/develop/pre-push/10/internal-qa-aws/
2026-09-02 18:01:54        240 MD5SUMS
2026-09-02 18:01:54        368 SHA256SUMS
2026-09-02 18:01:54    1538897 internal-qa-aws.cdx.json
2026-09-02 18:03:11 6086165824 internal-qa-aws.debs.tar.gz
2026-09-02 18:01:54      21393 internal-qa-aws.packages.list
2026-09-02 18:01:54 7674152960 internal-qa-aws.vmdk
sanjeev.rohilla@APAC-SANJEEV~/syft  add-packaging-code % aws s3 ls s3://dev-de-images/builds/jenkins-selfservice.sanjeev.rohila/appliance-build/develop/pre-push/10/internal-dev-aws/
2026-09-02 18:16:20        244 MD5SUMS
2026-09-02 18:16:20        372 SHA256SUMS
2026-09-02 18:16:20    1841098 internal-dev-aws.cdx.json
2026-09-02 18:18:06 6203335421 internal-dev-aws.debs.tar.gz
2026-09-02 18:16:20      25360 internal-dev-aws.packages.list
2026-09-02 18:16:20 9840958464 internal-dev-aws.vmdk
sanjeev.rohilla@APAC-SANJEEV~/syft  add-packaging-code % aws s3 ls s3://dev-de-images/builds/jenkins-selfservice.sanjeev.rohila/appliance-build/develop/pre-push/10/external-standard-aws/
2026-09-02 17:58:24        264 MD5SUMS
2026-09-02 17:58:24        392 SHA256SUMS
2026-09-02 17:58:24    1338564 external-standard-aws.cdx.json
2026-09-02 17:59:28 5984082570 external-standard-aws.debs.tar.gz
2026-09-02 17:58:24      18702 external-standard-aws.packages.list
2026-09-02 17:58:24 7440211456 external-standard-aws.vmdk
sanjeev.rohilla@APAC-SANJEEV~/syft  add-packaging-code % 

````


</details>

<!--
<details open>
<summary><h2> Notes to Reviewers </h2></summary>

</details>

<details open>
<summary><h2> Future Work </h2></summary>
</details>
-->

### DIFF
--- a/bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
+++ b/bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
@@ -81,33 +81,6 @@
     state: present
     update_cache: true
 
-# Syft and cyclonedx-cli are used by scripts/run-live-build.sh to generate and validate a
-# CycloneDX SBOM for each built image. Neither is available through the Delphix package
-# mirror -- they're upstream GitHub-release binaries, not Ubuntu packages -- so each is
-# pinned to an explicit version and checksum-verified here instead of installed via apt.
-- name: Download pinned Syft .deb
-  ansible.builtin.get_url:
-    url: https://github.com/anchore/syft/releases/download/v1.46.0/syft_1.46.0_linux_amd64.deb
-    dest: /tmp/syft_1.46.0_linux_amd64.deb
-    checksum: sha256:003f82f0d70350a2e7795ecc43f86b50e2c883055ffaf82b309a0a274c62915d
-    mode: "0644"
-
-- name: Install Syft
-  ansible.builtin.apt:
-    deb: /tmp/syft_1.46.0_linux_amd64.deb
-
-# cyclonedx-cli ships only a bare per-platform binary upstream (no .deb, and no official
-# checksums file); the checksum below was computed locally from the downloaded binary at
-# the time this version was pinned. The destination filename must be exactly
-# "cyclonedx-cli" -- that's the command name run-live-build.sh invokes, not upstream's own
-# convention of renaming the asset to "cyclonedx".
-- name: Download pinned cyclonedx-cli binary
-  ansible.builtin.get_url:
-    url: https://github.com/CycloneDX/cyclonedx-cli/releases/download/v0.32.0/cyclonedx-linux-x64
-    dest: /usr/local/bin/cyclonedx-cli
-    checksum: sha256:454879e6a4a405c8a13bff49b8982adcb0596f3019b26b0811c66e4d7f0783e1
-    mode: "0755"
-
 # awscli is distributed via pip or snap on 24.04. While the package is not required by the product,
 # it is required by appliance-build itself.
 - name: Install awscli python package
--- a/bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
+++ b/bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
@@ -81,6 +81,33 @@
     state: present
     update_cache: true
 
+# Syft and cyclonedx-cli are used by scripts/run-live-build.sh to generate and validate a
+# CycloneDX SBOM for each built image. Neither is available through the Delphix package
+# mirror -- they're upstream GitHub-release binaries, not Ubuntu packages -- so each is
+# pinned to an explicit version and checksum-verified here instead of installed via apt.
+- name: Download pinned Syft .deb
+  ansible.builtin.get_url:
+    url: https://github.com/anchore/syft/releases/download/v1.46.0/syft_1.46.0_linux_amd64.deb
+    dest: /tmp/syft_1.46.0_linux_amd64.deb
+    checksum: sha256:003f82f0d70350a2e7795ecc43f86b50e2c883055ffaf82b309a0a274c62915d
+    mode: "0644"
+
+- name: Install Syft
+  ansible.builtin.apt:
+    deb: /tmp/syft_1.46.0_linux_amd64.deb
+
+# cyclonedx-cli ships only a bare per-platform binary upstream (no .deb, and no official
+# checksums file); the checksum below was computed locally from the downloaded binary at
+# the time this version was pinned. The destination filename must be exactly
+# "cyclonedx-cli" -- that's the command name run-live-build.sh invokes, not upstream's own
+# convention of renaming the asset to "cyclonedx".
+- name: Download pinned cyclonedx-cli binary
+  ansible.builtin.get_url:
+    url: https://github.com/CycloneDX/cyclonedx-cli/releases/download/v0.32.0/cyclonedx-linux-x64
+    dest: /usr/local/bin/cyclonedx-cli
+    checksum: sha256:454879e6a4a405c8a13bff49b8982adcb0596f3019b26b0811c66e4d7f0783e1
+    mode: "0755"
+
 # awscli is distributed via pip or snap on 24.04. While the package is not required by the product,
 # it is required by appliance-build itself.
 - name: Install awscli python package
--- a/scripts/run-live-build.sh
+++ b/scripts/run-live-build.sh
@@ -17,7 +17,7 @@
 
 . "${BASH_SOURCE%/*}/common.sh"
 
-check_env DELPHIX_PACKAGE_MIRROR_MAIN DELPHIX_PACKAGE_MIRROR_SECONDARY
+check_env DELPHIX_PACKAGE_MIRROR_MAIN DELPHIX_PACKAGE_MIRROR_SECONDARY DELPHIX_APPLIANCE_VERSION
 
 TOP=$(git rev-parse --show-toplevel 2>/dev/null)
 
@@ -170,6 +170,30 @@ lb config \
 	--mirror-binary-backports "$DELPHIX_PACKAGE_MIRROR_MAIN"
 
 lb build
+
+#
+# Syft and cyclonedx-cli are consumed exactly the same way as every other
+# first-party Delphix package: as .deb's in the ancillary repository (see
+# build-ancillary-repository.sh), already being served locally above. The
+# one difference is that these are build-host-only tooling for the SBOM
+# step below, never installed into the appliance chroot -- so, unlike
+# "lb config"'s --mirror-chroot/--mirror-binary (which points the chroot's
+# own bootstrap at config/archives/localhost.list), these are installed
+# directly onto the build host itself via apt, pointed at that same
+# ancillary repository, before the still-running local Aptly server
+# serving it is killed just below.
+#
+# Reuse config/archives/localhost.list itself (rather than re-typing the
+# same "deb [trusted=yes] http://localhost:8080 noble main" line here) so
+# this can never silently drift out of sync with what the chroot uses.
+#
+cp config/archives/localhost.list /etc/apt/sources.list.d/ancillary-repository.list
+apt-get update \
+	-o Dir::Etc::sourcelist="sources.list.d/ancillary-repository.list" \
+	-o Dir::Etc::sourceparts="-" \
+	-o APT::Get::List-Cleanup="0"
+apt-get install -y --no-install-recommends delphix-syft delphix-cyclonedx-cli
+rm -f /etc/apt/sources.list.d/ancillary-repository.list
 
 kill -9 $APTLY_SERVE_PID
 
--- a/scripts/run-live-build.sh
+++ b/scripts/run-live-build.sh
@@ -199,7 +199,16 @@ fi
 # is covered separately by per-package sidecar SBOMs produced upstream
 # in linux-pkg, merged in at a later stage.
 #
-syft scan dir:binary \
+# SYFT_FILE_METADATA_SELECTION=none suppresses Syft's default behavior of
+# emitting a separate CycloneDX "file" component (with SHA-1/SHA-256
+# hashes) for every individual file owned by every cataloged package.
+# That's unrelated to --select-catalogers (which only scopes *package*
+# catalogers) and, left at its default, balloons the output to one
+# "file" entry per file on the appliance -- tens of thousands of them --
+# instead of the flat, per-package pkg:deb document this step is meant
+# to produce.
+#
+SYFT_FILE_METADATA_SELECTION=none syft scan dir:binary \
 	--select-catalogers dpkg \
 	--source-name "$ARTIFACT_NAME" \
 	--source-version "$DELPHIX_APPLIANCE_VERSION" \
--- a/scripts/run-live-build.sh
+++ b/scripts/run-live-build.sh
@@ -17,7 +17,7 @@
 
 . "${BASH_SOURCE%/*}/common.sh"
 
-check_env DELPHIX_PACKAGE_MIRROR_MAIN DELPHIX_PACKAGE_MIRROR_SECONDARY DELPHIX_APPLIANCE_VERSION
+check_env DELPHIX_PACKAGE_MIRROR_MAIN DELPHIX_PACKAGE_MIRROR_SECONDARY
 
 TOP=$(git rev-parse --show-toplevel 2>/dev/null)
 
@@ -171,30 +171,6 @@ lb config \
 
 lb build
 
-#
-# Syft and cyclonedx-cli are consumed exactly the same way as every other
-# first-party Delphix package: as .deb's in the ancillary repository (see
-# build-ancillary-repository.sh), already being served locally above. The
-# one difference is that these are build-host-only tooling for the SBOM
-# step below, never installed into the appliance chroot -- so, unlike
-# "lb config"'s --mirror-chroot/--mirror-binary (which points the chroot's
-# own bootstrap at config/archives/localhost.list), these are installed
-# directly onto the build host itself via apt, pointed at that same
-# ancillary repository, before the still-running local Aptly server
-# serving it is killed just below.
-#
-# Reuse config/archives/localhost.list itself (rather than re-typing the
-# same "deb [trusted=yes] http://localhost:8080 noble main" line here) so
-# this can never silently drift out of sync with what the chroot uses.
-#
-cp config/archives/localhost.list /etc/apt/sources.list.d/ancillary-repository.list
-apt-get update \
-	-o Dir::Etc::sourcelist="sources.list.d/ancillary-repository.list" \
-	-o Dir::Etc::sourceparts="-" \
-	-o APT::Get::List-Cleanup="0"
-apt-get install -y --no-install-recommends delphix-syft delphix-cyclonedx-cli
-rm -f /etc/apt/sources.list.d/ancillary-repository.list
-
 kill -9 $APTLY_SERVE_PID
 
 #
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by sebroy** [comment]
This is a fantastic start, and it's encouragingly simple. The only issue I have is with the way we're fetching the tools and pinning them to specific versions.

File: bootstrap/roles/appliance-build.bootstrap/tasks/main.yml [hunk 1]
```diff
@@ -81,6 +81,33 @@
...
+# the time this version was pinned. The destination filename must be exactly
+# "cyclonedx-cli" -- that's the command name run-live-build.sh invokes, not upstream's own
+# convention of renaming the asset to "cyclonedx".
+- name: Download pinned cyclonedx-cli binary
+  ansible.builtin.get_url:
+    url: https://github.com/CycloneDX/cyclonedx-cli/releases/download/v0.32.0/cyclonedx-linux-x64
+    dest: /usr/local/bin/cyclonedx-cli
+    checksum: sha256:454879e6a4a405c8a13bff49b8982adcb0596f3019b26b0811c66e4d7f0783e1
+    mode: "0755"
+
```
sebroy, line 110: We generally discourage installing build dependencies from the public Internet at runtime. This introduces instabilities in our builds. Separately, it feels like hard-coding versions in here is an anti-pattern (this becomes immediate tech-debt).

Instead, we should do one of:
1. If there's a debian package that can be fetched from somewhere, let's add it to our mirror so that our mirror always has the latest version that we can install (see recent examples of how package sources have been added to our mirrir in devops-gate).
2. If that's not possible, then use linux-pkg to build our own package that we can install. For cyclonedx-cli, we could either package the standalone binary itself, or build it from source (either way works).

sebroy (changes requested):

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
justsanjeev: HI Seb,

I went through your suggestions, below  is my finding to your suggestions -

**Option 1 (mirror an existing apt repo)** — Its not viable for either tool. I found the actual mechanism: [jenkins/scripts/linux_mirror_sync_ubu_repo/dlpx-branch-packages.json](https://github.com/delphix/devops-gate/blob/main/jenkins/scripts/linux_mirror_sync_ubu_repo/dlpx-branch-packages.json) is the real "list of package sources".

It's how Docker's, Postgres', Microsoft's, and InfluxData's own official apt repos get mirrored into Delphix's internal Aptly-based mirror ([ansible/roles/internal/dlpx.linux-package-mirror-v2](https://github.com/delphix/devops-gate/tree/main/ansible/roles/internal/dlpx.linux-package-mirror-v2)). 

But this only works by mirroring a repo that already exists upstream. I checked both tools' own install docs directly: 
- Syft's README only offers curl | sh, Homebrew, Docker, Scoop, Chocolatey, Nix; 
- cyclonedx-cli's only offers raw GitHub release binaries, Homebrew, and Docker. Neither publishes an actual apt/deb repository — there's nothing to add an entry for.


**Option 2 (linux-pkg builds our own package)** — Its viable, and there's an exact precedent. packages/delphix-go/ and packages/host-jdks/ in linux-pkg do exactly what we need: fetch a pre-built third-party binary/tarball, checksum-verified, from Delphix's own Artifactory (not hit upstream at build time) via fetch-and-run-installer.sh/fetch-file-from-artifactory.sh, then repackage it as a .deb via dpkg_buildpackage_default. This is "download once to local hosting, install from there" — just implemented as an Artifactory proxy/cache rather than the apt mirror.

Once packaged this way, it plugs into the existing pipeline automatically: add syft/cyclonedx-cli to package-lists/build/main.pkgs, and it flows through the same combine-packages → S3 manifest → appliance_build_stage0.groovy 
  mechanism every other 1st-party Delphix package already uses — completely bypassing both the apt mirror and any live GitHub fetch.

 delphix-go is in linux-pkg/package-lists/build/main.pkgs — the same list the plan is to add syft/cyclonedx-cli to — but it's a pure build-time toolchain package (packages the Go compiler itself, used to compile other         
  1st-party packages like zfs/ptools). 
  
I checked: delphix-go never appears in any of appliance-build's own chroot package lists (live-build/config/package-lists/delphix-platform.list.chroot.in. So despite linux-pkg building it and it being available via the combined-packages S3 flow, it never actually lands in binary/, and it's invisible to syft scan dir:binary today.       

So the correct implementation, mirroring delphix-go exactly: add syft/cyclonedx-cli to linux-pkg's [main.pkgs](https://github.com/delphix/linux-pkg/blob/develop/package-lists/build/main.pkgs) -- so they get built and published like any other 1st-party package.

Never reference them from any of appliance-build's .list.chroot files. The build-host provisioning step (today's ansible bootstrap role) would fetch the resulting .deb and apt install it onto the ephemeral build VM directly — same mechanic as today's  GitHub-sourced install, just swapping the source URL from github.com to Delphix's own S3/Artifactory-hosted package. As long as that stays true, these tools stay exactly where the design spec always intended — build-host-only, never inside the shipped image — and won't show up as pkg:deb components in the SBOM.

On the basis of all this information- I have created an another minimal spec as well.

justsanjeev: I am marking this thread resolved. We have done the changes as per suggested plan.

scripts/run-live-build.sh
```diff
@@ -187,6 +187,39 @@ if [[ ! -f binary/SHA256SUMS ]]; then
 	exit 1
 fi
 
+#
```
prakashsurya: I think this logic should be moved into a live-build hook, instead of adding to this script.. live-build allows us to run arbitrary scripts, so we could add a new script that's specific to running these commands..

e.g. add a new script in this directory: https://github.com/delphix/appliance-build/tree/develop/live-build/config/hooks .. not sure which of `configuration` or `vm-artifacts` it should go in yet, but I think that'd be a better, and more isolated, solution.. that is more consistent with the existing logic..

bootstrap/roles/appliance-build.bootstrap/tasks/main.yml
```diff
@@ -81,6 +81,33 @@
     state: present
     update_cache: true
 
+# Syft and cyclonedx-cli are used by scripts/run-live-build.sh to generate and validate a
```
prakashsurya: I think this logic still needs to be removed, and replaced with installing our downstream built packages, right?

justsanjeev: Fixed in 9f5534a -- removed the ansible.builtin.get_url tasks that curled Syft/cyclonedx-cli from github.com. They're now installed via apt from the local ancillary repository (same .deb-based mechanism as every other first-party Delphix package), pointed at delphix-syft/delphix-cyclonedx-cli built via linux-pkg (delphix/linux-pkg#414). Confirmed they never appear in any live-build chroot package list, so they stay build-host-only and are never shipped inside the appliance image.

scripts/run-live-build.sh
justsanjeev: Looked into this concretely rather than deferring it. The reason it's inline rather than a hook: the `if [[ ! -f binary/SHA256SUMS ]]; then exit 1; fi` check right above this (existing code, not part of this PR) exists specifically because `lb build`'s own exit code is known to be unreliable -- per the comment above it, it can return success even when the build actually failed. `binary/SHA256SUMS` is generated by live-build itself as one of the very last things it does internally, *after* all custom `.binary` hooks have already run (confirmed nothing in this repo creates it -- it's a live-build-internal artifact). So a hook has no way to gate on that check; it would run before the file exists.

Traced the rest too: the local Aptly `aptly serve` process (this step's dependency for installing `delphix-syft`/`delphix-cyclonedx-cli`) is started before `lb config`/`lb build` and only killed after, so it would still be reachable from inside any hook -- the install mechanics themselves aren't the blocker, just this specific safety net. Since that check exists because of a previously-observed real failure mode, I'd rather not silently drop it by moving this into a hook. Keeping it inline for now.

prakashsurya: why is that file relevant to this discussion? sorry, I don't know how these utilities work...

justsanjeev:
With ref to the latest successful Build: http://selfservice.dlpx98618.dcol1.delphix.com/job/appliance-build-orchestrator-pre-push/9/ after doing changes, bringing back the PR to **Ready for review.**

scripts/run-live-build.sh
justsanjeev: Sorry, should've been clearer -- I mean `binary/SHA256SUMS`, the file checked two lines above this diff (`if [[ ! -f binary/SHA256SUMS ]]; then exit 1; fi`).

Context: `lb build` (the live-build tool that produces the appliance's rootfs into the `binary/` directory) has a known quirk where it can return a "success" exit code even when the build actually failed internally. So this script doesn't trust `lb build`'s own exit code -- instead, right after `lb build` finishes, it checks whether `binary/SHA256SUMS` exists. That file is something live-build itself writes out as one of the very last things it does when a build genuinely succeeds, so its presence is used here as the real "did this actually work" signal.

Why that matters for the hook question: a live-build hook (like the ones in `live-build/config/hooks/`) runs *during* `lb build`, before live-build has finished and before it's written that file. So if I moved the Syft/cyclonedx-cli step into a hook, it would have no way to check `binary/SHA256SUMS` first -- that specific safety check only works because this code runs *after* `lb build` returns, outside of any hook. That's the reason I kept it inline instead of moving it, even though a hook would otherwise be a cleaner fit.

justsanjeev:
Kicked off a new build with the latest changes (`9f5534a`): http://selfservice.dlpx98618.dcol1.delphix.com/job/appliance-build-orchestrator-pre-push/10/

justsanjeev:
A new Build has been kicked off, which has already completed the sbom creation
- ab-pre-push -> http://selfservice.dlpx98618.dcol1.delphix.com/job/appliance-build-orchestrator-pre-push/11/
- appliance-build-stage0 -> http://selfservice.dlpx98618.dcol1.delphix.com/job/appliance-build-stage0/job/pre-push/8/ 

The sbom creation part is completed. The review is up, Kindly provide your reviews.

Thanks

scripts/run-live-build.sh
```diff
@@ -171,6 +171,30 @@ lb config \
...
+# same "deb [trusted=yes] http://localhost:8080 noble main" line here) so
+# this can never silently drift out of sync with what the chroot uses.
+#
+cp config/archives/localhost.list /etc/apt/sources.list.d/ancillary-repository.list
+apt-get update \
+	-o Dir::Etc::sourcelist="sources.list.d/ancillary-repository.list" \
+	-o Dir::Etc::sourceparts="-" \
+	-o APT::Get::List-Cleanup="0"
+apt-get install -y --no-install-recommends delphix-syft delphix-cyclonedx-cli
+rm -f /etc/apt/sources.list.d/ancillary-repository.list
```
sebroy: One minor issue with the placement of this code is that the syft and cyclonedx-cli packages only need to be installed once rather than for each variant/platform combination (run-live-build.sh gets called for each of those). As such, this code might be better placed in `live-build/build.gradle`. It could go at the end of the existing `ancillaryRepository` task. Note that you don't need a running aptly server, and you don't need to modify the system's apt sources config to install these. You can install the debs directly from the directory they're in (e.g. using `dpkg -i /.../something.deb`).

scripts/run-live-build.sh
sebroy: I tend to agree with Prakash on this point. Error handling of various hooks is already a core part of the lb subsystem. If your hook runs, then you can assume that the previous hooks were successful. A `config/hooks/vm-artifacts/93-generate-sbom.binary` hook feels like a fairly elegant approach.

sebroy (changes requested):

scripts/run-live-build.sh
justsanjeev: - The apt-get install delphix-syft delphix-cyclonedx-cli step moved out of scripts/run-live-build.sh (which runs once per variant/platform combination) and into scripts/build-ancillary-repository.sh (which runs once and is   
    shared across every variant/platform combination).                                                                                                                                                                             
  - Installation now goes straight from the .deb files just produced in $WORK_DIRECTORY/debs via apt-get install -y --no-install-recommends "$deb", instead of standing up a local Aptly APT source                                
    (/etc/apt/sources.list.d/ancillary-repository.list + apt-get update) on the build host each time.                                                                                                                              
  - The actual SBOM generation/validation logic (syft scan + cyclonedx-cli validate) moved into a new live-build hook, live-build/config/hooks/configuration/95-generate-sbom.binary, which just consumes the already-installed tooling — it doesn't reinstall anything now.                                    
  
  syft/cyclonedx-cli are now installed exactly once per build (in the ancillary-repository step) rather than once per platform/variant combination in run-live-build.sh

scripts/run-live-build.sh
justsanjeev: Thank you both for pushing on this -- you were right, and I appreciate you not letting it go.

I went back and actually traced through the exact `live-build` version this repo uses (Ubuntu Noble's `live-build 3.0~a57-1ubuntu49.1`). `lb_binary` runs with `set -e` and calls `lb binary_hooks` before `lb binary_checksums`. `lb_binary_hooks` explicitly does `./"${HOOK}" || { Echo_error ...; exit 1; }` for every `config/hooks/*.binary` script, so if our new hook fails, `binary_hooks` exits non-zero, `set -e` stops `lb_binary` right there, and `lb binary_checksums` (which writes `binary/SHA256SUMS`) never runs. So the existing `if [[ ! -f binary/SHA256SUMS ]]` check in `run-live-build.sh` still catches a hook failure correctly, from the outside, exactly as before -- I was wrong that moving this into a hook would lose that safety net.

I've moved it into `live-build/config/hooks/configuration/95-generate-sbom.binary` in 7f77ef2. One small deviation from your suggestion, Seb: I put it under `configuration/` rather than `vm-artifacts/`, and numbered it after everything else, so it still runs for the `upgrade-artifacts` run type (which also needs the `.cdx.json` output per `build.gradle`, and which `run-live-build.sh` never copies `vm-artifacts/` hooks for) and so it always runs last regardless of run type. Happy to adjust if you see it differently.

Sorry for the back-and-forth on this one, and thanks again for catching it.

sebroy (approved):

PR #893, opened 2026-08-13
CP-13600 Add syft/cyclonedx-cli provisioning design spec

justsanjeev:
## Problem
Delphix appliance images ship without a machine-consumable SBOM. The only bill-of-materials
today is a home-grown CSV (devops-gate generate_bom.py / combine_bom.py) built by
brittle text-parsing of per-package copyright files -- license-centric, no PURLs, and not
consumable by vulnerability scanners.

CP-13464 (PR #892) implements the appliance-build side of the fix, but review feedback
flagged its provisioning approach -- downloading syft/cyclonedx-cli directly from github.com
at build time -- as against convention (no internet build deps, no hardcoded versions).

## Solution
Adds a design spec (docs/specs/2026-08-13-syft-cyclonedx-cli-provisioning-design.md)
evaluating both alternatives suggested in review:
- Mirroring an apt repo: not viable, neither tool publishes one upstream.
- Packaging via linux-pkg (proposed): modeled on the existing delphix-go/host-jdks pattern --
  fetch the pre-built release binary from Delphix's own Artifactory, checksum-verified,
  package as a .deb, consumed through the existing combine-packages/S3 flow like any other
  first-party package.

Includes an explicit guardrail, verified against the delphix-go precedent, that these
packages must never be referenced by any appliance-build chroot package list, so they stay
build-host-only tooling and never ship inside the appliance image.

Spec only -- no implementation in this PR.

## Testing Done
Documentation only.

### DIFF
--- a/docs/specs/2026-08-13-syft-cyclonedx-cli-provisioning-design.md
+++ b/docs/specs/2026-08-13-syft-cyclonedx-cli-provisioning-design.md
@@ -1,0 +1,143 @@
+# Syft / cyclonedx-cli Build-Host Provisioning — Design
+
+- **Companion to:** `docs/specs/2026-06-15-sbom-generation-design.md` (CP-13456)
+- **Epic:** [CP-13455](https://perforce.atlassian.net/browse/CP-13455) — CycloneDX SBOM for Delphix engine product images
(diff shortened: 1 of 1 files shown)


sebroy:
That looks perfect, let's go with this.

sebroy (approved):

PR #902, opened 2026-09-08
DLPX-98670 protect running kernel's dependency closure from upgrade purge

prakashsurya:
## Summary

- `execute`'s final package purge protected the running kernel's own packages by name only (`apt-mark showauto | grep -v "$(uname -r)"`), but `apt-get purge` also removes a package's reverse dependents — so a dropped dependency that doesn't itself carry the kernel release in its name (e.g. `wireless-regdb`, which `linux-modules-extra-$(uname -r)` depends on) cascades the purge back up and takes the kernel package with it.
- The failure is silent at apply time (the module is already loaded) and only surfaces as a management-stack outage at some arbitrary later boot (see DLPX-98412).
- Protects the running kernel's full installed dependency closure (via `apt-cache depends --recurse --installed`) instead of just package names, and asserts after the purge that none of those packages went missing, so a gap fails the upgrade loudly instead of deferring it to a later boot.

Jira: https://perforce.atlassian.net/browse/DLPX-98670

## Test plan

- [x] `shellcheck -x` clean on both changed files (only pre-existing, unrelated `SC1091`/`SC2119` info notices remain)
- [x] Manually verified the new `common.sh` functions against this dev VM's real package state (a `linux-modules-extra`/`wireless-regdb`-shaped kernel), confirming the dependency closure protects the vulnerable package and `verify_packages_installed` accepts a healthy system / rejects a missing one
- [ ] No automated test added — this directory (`upgrade/upgrade-scripts/tests/`) has no CI cadence today (existing tests there are manual, root-required, run-by-hand); a temporary reproduction script was used locally to validate the fix and then dropped rather than added as unreachable dead code

### DIFF
--- a/upgrade/upgrade-scripts/common.sh
+++ b/upgrade/upgrade-scripts/common.sh
@@ -365,6 +365,63 @@ function xargs_apt_get() {
 		"$@"
 }
 
+#
+# List installed packages whose name contains "$1" (typically the
+# running kernel release, "$(uname -r)").
+#
+function get_installed_packages_matching() {
+	[[ -n "$1" ]] || die "pattern not specified"
+	local PACKAGES
+	PACKAGES="$(dpkg-query -Wf '${Package}\n')" || die "dpkg-query failed"
+	grep -F "$1" <<<"$PACKAGES" || true
+}
+
+#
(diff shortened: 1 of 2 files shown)


hongwongdelphix (approved):
LGTM

PR #905, opened 2026-09-15
DLPX-98978 internal-claude: pre-install the repositories the engine is built from

prakashsurya:
<details open>

The `internal-claude` variant clones six repositories into `~delphix/src`:

```yaml
    - { org: delphix, name: dlpx-app-gate, version: develop }
    - { org: delphix, name: dlpx-qa-gate, version: develop }
    - { org: delphix, name: dms-core-gate, version: develop }
    - { org: delphix, name: zfs, version: develop }
    - { org: delphix, name: cd-aidlc, version: main }
    - { org: delphix, name: git-utils, version: main }
```

That covers the product gates and the tooling, but it leaves out most of what
the appliance is actually assembled from. `appliance-build`, `linux-pkg`, and
`delphix-platform` are all absent, as is every repository `linux-pkg` builds a
package from, so an agent asked why a package looks the way it does has to
work out which repository owns it and clone it first, using whatever
credentials it can find on the image.

The list is also spelled out twice in the role, once to clone and once to
rewrite the remote URL, which is survivable at six entries and not at fifty.
</details>

<details open>

The solution taken in this PR is to derive the engine-input half of the list
from `linux-pkg` rather than curating it by hand, so that it can follow along
as packages come and go: every repository named in a package's
`DEFAULT_PACKAGE_GIT_URL`, plus `appliance-build` and `linux-pkg` themselves,
`delphix-platform`, `delphix-kernel`, and `delphix-upgrade-verification`.
Alongside those sit the build and CI repositories the image's existing skills
already drive (`devops-gate`, `dcoa`, `pipeline-shared`, `qa-infra`,
`github-infra`, and both claude-plugins repos), and the QA repositories
(`dlpx-qa-gate`, `qa-appdata-toolkits`, `golden-images`). The API gateway repositories
(`apigw-services`, `apigw-ui`, `apigw-support`) ride along too, in their own group:
none of the three is a linux-pkg package, so unlike most of the list they fall outside
the derivation above, and the group comment says so once rather than per entry.
52 in total.

Two `linux-pkg` packages are deliberately absent, because they have no git
source to clone: `grub2` fetches a prebuilt from the `release-de-images` S3
bucket and `misc-debs` fetches prebuilt `.deb`s from Artifactory, so both set
`DEFAULT_PACKAGE_GIT_URL=none`.

Branches are not uniformly `develop`, which is why each entry carries its own:
35 `develop`, 12 `main`, 4 `master`, and `dcoa` is `stage`.

The five per-platform kernels clone with `depth: 1`. At full depth they are
~20GB packed, roughly an order of magnitude more than everything else here
combined, on a variant whose rpool is 70GB (`90-raw-disk-image.binary` gives
`internal-*` 70GB), and that 20GB would come off GitHub on every image build.
Arguably it'd be better to clone them blobless (`--filter=blob:none`) so that
the history survives, but `ansible.builtin.git` has no option for it and
that'd mean dropping those five to a `command` task. Thus, shallow for now;
`git fetch --unshallow` recovers the history on the image.

That shallowness is what the third task is for. A `--depth` clone is also a
single-branch clone, so git narrows `remote.origin.fetch` to the one branch
cloned, and neither unshallowing nor a checkout of any other branch would see
anything past it:

```
$ git config --get remote.origin.fetch     # after a --depth=1 clone
+refs/heads/develop:refs/remotes/origin/develop
```

Restoring the wildcard refspec afterwards makes both work, and is a no-op for
the 47 full clones, which already hold that value.

The list itself moves to the role's `defaults/main.yml` as
`claude_internal_repos`, with `org` defaulting to `delphix` and an optional
`depth`, so the clone loop and the two URL-rewrite loops all read one
definition. The clone task also retries, since the list is long enough, and
the gates large enough, that one transient GitHub failure would otherwise
discard an entire image build.
</details>

<details open>

### Local verification

- All 52 `(repo, branch)` pairs resolve against the GitHub API, rather than
  being assumed to be `develop`. That is what caught `delphix-go`, `dwarves`,
  `qa-infra`, and `golden-images` being `master`, and `dcoa` being `stage`.
- The three clone tasks were run for real against three small repositories,
  one of them shallow, with the task bodies copied verbatim and only `dest`
  repointed at a temp directory. Asserted afterwards, and all passing: `depth`
  applies only where requested, no token survives in any remote URL, and every
  repository ends with the wildcard fetch refspec.

  ```
  challenge-response   shallow=false (expected false) commits=23 url=git@github.com:delphix/challenge-response.git
  savedump             shallow=false (expected false) commits=40 url=git@github.com:delphix/savedump.git
  connstat             shallow=true  (expected true ) commits=1  url=git@github.com:delphix/connstat.git
  ```

  The `commits=1` against `commits=23`/`commits=40` is the part worth having:
  it shows `depth: "{{ item.depth | default(omit) }}"` actually omitting, not
  silently applying a depth everywhere.
- `ansible-playbook --syntax-check` and `--list-tasks` clean on
  `live-build/variants/internal-claude/ansible/playbook.yml`, with the new task
  appearing in position.
- `ansible-lint` (26.8.0, repo-wide, the same no-args invocation
  `.github/workflows/main.yml` uses) reports 2 fatal `fqcn[action]` failures,
  both pre-existing and in roles this PR does not touch
  (`unittest-internal`, `virtualization-development`). This role's own two
  findings are `command-instead-of-module` warnings marked `# ignored`, one of
  which is the pre-existing `set-url` task.
- `./gradlew check` is not a signal here. It fails on an untouched `develop`
  on this workstation, because the local `shfmt` disagrees with CI's about a
  trailing `\` after `|` in 7 shell files, and it does not read YAML anyway.

### ab-pre-push build #14997 — PASSED (SUCCESS)

- Build: https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build-orchestrator-pre-push/14997/
- Tested: `appliance-build@2fe02e0` (against `develop`), the current head of this PR.
- Variants: `internal-claude`, platform `aws`, imported into DCenter as the same
  variant, `BUILD_ARTIFACTS=ALL`.
- Suites: none ran (`RUN_TESTS=False`), so the build result is the whole signal
  from Jenkins: all 52 clones resolved and succeeded inside the chroot with the
  bot token, and the three tasks completed in position.
- Image: DCoA group `dlpx-psurya-develop`, `ami-0d994dd5b9e41cb47`, vmdk from
  `pre-push/7389`.

#### Post-boot verification, on a clone of that image

What the build cannot show is the post-boot state, so the image was cloned as
`verify-98978-repos`, alongside `baseline-98978` from
`dlpx-internal-claude-develop` (weekly 190, `857ef56`) as the 6-repo control.
Every check below covers all 52 entries of `claude_internal_repos` rather than a
sample of them:

- 52 directories under `/export/home/delphix/src`, exactly the configured set,
  with no missing entry and no extra one.
- Branch matches on all 52, i.e. every per-entry `version` is right: 35
  `develop`, 12 `main`, 4 `master`, and `dcoa` on `stage`.
- The five kernels are shallow (`rev-parse --is-shallow-repository` true,
  `rev-list --count HEAD` of 1), and the other 47 hold full history
  (`dlpx-app-gate` 36680 commits, `dlpx-qa-gate` 18583, `zfs` 14219), so `depth`
  applies only where it was asked for.
- All 52 remotes are `git@github.com:delphix/<repo>.git`, a single unique form
  across the set, so no build-time token survives on the image. A grep across the
  156 git metadata files that could plausibly retain one (`config`, `logs/HEAD`,
  `FETCH_HEAD`, `packed-refs`) finds no token pattern either.
- All 52 carry `remote.origin.fetch = +refs/heads/*:refs/remotes/origin/*`,
  the five shallow ones included, which is what the third task exists for.
- All 52 working trees are clean and on a branch, none detached. All are
  `delphix:staff` with group write, and as `delphix` both a write inside a
  repository and a branch create/delete succeed.

One thing not mentioned above that the image makes plain: `ansible.builtin.git`
defaults `recursive: yes`, so the submodules came along too, `bcc` pulling three
(`blazesym`, `bpftool`, `libbpf`), `dwarves` and `zfs` one each. All five are
public upstream repositories fetched over plain https, so they carry no token,
but they are why `.git/config` holds `https://` URLs on those three.

#### What the checkouts cost the build

Measured against stage1 #64931, the same variant and platform on `f6663dd` with
the old 6-repo list, i.e. the last `internal-claude` pre-push before this change:

| phase | 6 repos | 52 repos | delta |
|---|---|---|---|
| ansible: Clone product repositories | 4m09s | 16m52s | +12m42s |
| ansible: Reset clone URLs to SSH | ~0s | 6.7s | +6.7s |
| ansible: Restore the wildcard fetch refspec | (new task) | 13.5s | +13.5s |
| ansible: `chown -R` of `src` | 10.4s | 2m44s | +2m33s |
| **the checkouts themselves** | **4m20s** | **19m56s** | **+15m36s** |
| rsync of the chroot into the ZFS rpool | 7m19s | 16m20s | +9m01s |
| `syft scan dir:binary` | 3m38s | 26m25s | +22m47s |
| live-build binary SHA256SUMS | 4m09s | 12m36s | +8m27s |
| img to vmdk convert | 6m53s | 11m14s | +4m21s |
| artifact `sha256sum` and `md5sum` | 2m34s | 3m55s | +1m21s |
| S3 publish | 1m37s | 2m19s | +42s |
| **stage1 total** | **60m53s** | **125m51s** | **+65m** |
| **orchestrator, end to end** | **105m44s** (#14974) | **171m58s** (#14997) | **+66m** |

So the checkouts are under a quarter of the added wall clock, and the rest is the
downstream cost of a larger image. The largest single item is `syft`, which is
invoked `--select-catalogers dpkg` and still walks the whole tree, spending 23
extra minutes reading 52 git repositories it will never catalog. Excluding
`/export/home/delphix/src` from the syft scan, and from the live-build SHA256SUMS
pass, would return up to roughly 31 minutes to every `internal-claude` build.
That belongs in its own change rather than here, and is worth a follow-up.

#### What the checkouts cost on disk

| measure | 6 repos | 52 repos | delta |
|---|---|---|---|
| vmdk artifact | 6.14 GiB | 10.4 GiB | +4.24 GiB |
| `rpool` ALLOC on a booted clone | 7.00 GiB | 12.2 GiB | +5.2 GiB |
| rpool used, of the 70GB pool | 10% | 17% (57.3 GiB free) | |
| `src` physical, post-compression | 2.12 GiB | 7.41 GiB | +5.29 GiB |
| `src` logical (`du -sb`) | 2.26 GiB | 11.80 GiB | +9.54 GiB |

The 70GB rpool is nowhere near binding at 17%, so the sizing argument for
`depth: 1` holds with a lot of room to spare. The five shallow kernels are still
4.45 GiB physical between them, which is 60% of `src` and 84% of the whole image
growth, at roughly 910 MiB each. That 910 MiB splits about 280 MiB of `.git`
against 625 MiB of checked-out tree, which bounds the obvious next idea: sharing
objects across the five (one clone with five worktrees, or `--reference`) would
recover only the four redundant `.git` copies, i.e. around 1.1 GiB, since five
checked-out trees are irreducible if all five are to be present. Dropping a
platform's kernel entirely is the only lever worth more than that, and it costs
more than it saves.

Worth recording separately, since 52 repositories make it more visible than 6 did:
`~delphix/src` is not reachable as `~/src` on the image. `HOME` is `/home/delphix`
since appliance-build#869 moved it, the repositories sit at
`/export/home/delphix/src`, and the only thing bridging the two is
`/etc/profile.d/git-utils-path.sh` putting the absolute path on `PATH`. The
baseline image behaves identically, so this is not a regression here, but a
`~/src` symlink would be cheap.
</details>


live-build/misc/ansible-roles/appliance-build.claude-internal/defaults/main.yml
```diff
@@ -0,0 +1,130 @@
...
+  #
+  # The build frameworks, and the platform/kernel metapackages they assemble.
+  #
+  - { name: appliance-build, version: develop }
+  - { name: linux-pkg, version: develop }
+  - { name: delphix-platform, version: develop }
+  - { name: delphix-kernel, version: develop }
+
+  #
+  # The product gates. These are the three large ones -- roughly 1.5GB of the total.
```
prakashsurya: add https://github.com/delphix/apigw-services

prakashsurya: Added as `{ name: apigw-services, version: main }` in this group, and rolled the group's comment to four repos at ~1.7GB. Note the branch is `main`, not `develop`.

It isn't a linux-pkg package, so it sits outside the `DEFAULT_PACKAGE_GIT_URL` derivation the rest of that section comes from. The commit message now names it explicitly rather than leaving it to look derived.

`apigw-ui` (42MB, `develop`) and `apigw-support` exist alongside it; say if you want those as well.

In 68e1779. All 50 `(repo, branch)` pairs still resolve. ab-pre-push #14993 tested the superseded 6395fc3, so it is being re-run against this SHA.

prakashsurya: Correction to what I wrote above: `apigw-ui` is 136MB on `main`, not 42MB on `develop`. I'd crossed it with `virtualization-sdk` from an earlier size listing, and the branch was the part that mattered, since `develop` doesn't exist there and the clone would have failed.

Both added on the verified branches, along with `apigw-support` (93KB, `main`). The three apigw repos now get their own group rather than being folded into the product gates, since none of them is a linux-pkg package and the group comment can say so once instead of per-entry.

52 repositories now, 47 full clones, branches 35 `develop` / 12 `main` / 4 `master` / 1 `stage`. All 52 pairs re-verified against the API. In 2fe02e0; ab-pre-push #14997.

lyriclake (approved):

PR #865, opened 2026-05-07
DLPX-97162 Add internal-claude variant for Claude Code developer environments

prakashsurya:
<details open>
<summary><h2> Problem </h2></summary>

Developers working with Claude Code need a pre-configured appliance image that
is ready to use on first boot — with the Claude CLI installed, product repos
pre-cloned, and standard tooling wired up. Today there is no such image; each
developer must manually install and configure everything on a stock internal VM.
</details>

<details open>
<summary><h2> Solution </h2></summary>

Add a new `internal-claude` appliance variant (AWS-only) that layers a
`claude-internal` Ansible role on top of `minimal-common + minimal-internal +
minimal-development`. On first boot the VM provides:

- **Claude CLI** installed via npm (`/usr/local/bin/claude`)
- **Six product repos** pre-cloned under `~delphix/src/` at their configured
  branches, with remote URLs rewritten to SSH form (no build-time token on disk):
  - `delphix/dlpx-app-gate` @ `develop`
  - `delphix/dlpx-qa-gate` @ `develop`
  - `delphix/dms-core-gate` @ `develop`
  - `delphix/zfs` @ `develop`
  - `delphix/cd-aidlc` @ `main`
  - `delphix/git-utils` @ `main`
- **support-tools group** (gid 4000) with `delphix` as a member
- **NAS mount** at `/nas` via a systemd `nas.mount` unit
  (`After/Wants=network-online.target`); uses a unit file rather than
  `/etc/fstab` because `90-raw-disk-image.binary` overwrites fstab after the
  ansible chroot stage
- **apt sources** modifiable post-boot (cloud-init override in
  `/etc/cloud/cloud.cfg.d/99-delphix-claude.cfg`)
- **git-utils/bin** in the login shell PATH via `/etc/profile.d/`

Also fixes `80-build-configuration.binary` to use `trap cleanup EXIT` so
`/proc`, `/sys`, and `/dev` bind mounts are always torn down even when the
ansible playbook fails mid-run.

The build requires `GITHUB_TOKEN` to be set (asserted at the top of the role);
without it the build fails fast with a clear message.
</details>

<details open>
<summary><h2> Testing Done </h2></summary>

- `./gradlew buildInternalClaudeAws` succeeds with `GITHUB_TOKEN` set —
  produces `internal-claude-aws.{vmdk,debs.tar.gz,packages.list}`
- `./gradlew buildInternalClaudeAws` without `GITHUB_TOKEN` fails fast at the
  assertion task with the expected `fail_msg`
[...]

PR #906, opened 2026-09-17
DLPX-99079 internal-claude: clone the product repositories into /home

prakashsurya:
<details open>

The `appliance-build.claude-internal` role clones its 52 repositories into
`/export/home/delphix/src`, which is no longer the `delphix` user's home
directory. DLPX-86523 moved the home dataset's mountpoint from `/export/home`
to `/home`, and deliberately shipped no `/export/home` compatibility symlink,
so the role's own `file` task is what brings the path into existence, as a
stray directory tree on the root dataset. `appliance-build.minimal-common`
sets `home: /home/delphix`, so on the image `cd ~/src` fails, and the
`~delphix/src` that the role's own task names and its `defaults/main.yml`
comments both refer to is never populated.

Seven occurrences, all in the one role:

```
live-build/misc/ansible-roles/appliance-build.claude-internal/tasks/main.yml:132  # task name
live-build/misc/ansible-roles/appliance-build.claude-internal/tasks/main.yml:134  # dir creation
live-build/misc/ansible-roles/appliance-build.claude-internal/tasks/main.yml:147  # clone dest
live-build/misc/ansible-roles/appliance-build.claude-internal/tasks/main.yml:163  # remote set-url
live-build/misc/ansible-roles/appliance-build.claude-internal/tasks/main.yml:176  # fetch refspec
live-build/misc/ansible-roles/appliance-build.claude-internal/tasks/main.yml:183  # ownership fixup
live-build/misc/ansible-roles/appliance-build.claude-internal/tasks/main.yml:202  # profile.d PATH
```

The variant was added in DLPX-97162 (#865) on 2026-05-08, five weeks before
the home move landed in #869 on 2026-06-16, and it was missed in that sweep.
The sibling development roles are all already correct; e.g.
`appliance-build.masking-development`,
`appliance-build.virtualization-development`, and
`appliance-build.zfsonlinux-development` clone into `/home/delphix`. This role
is the only `/export/home` reference left in `live-build`. The remaining ones
in the tree are in `upgrade/upgrade-scripts/{common.sh,execute}`, which is the
migration code itself and has to name the old path.
</details>

<details open>

The solution taken in this PR is to move the clone root into the role's
`defaults/main.yml` as `claude_internal_src_dir`, rather than repointing seven
inline literals at `/home`. Repeating the path inline is what let the home
move miss this role in the first place, and six tasks consume it now, so the
one-definition form is what keeps the next move from having the same problem.
The remaining `/export/home` mention in the role is the comment on that
default explaining why the path is what it is.

Moving the clones under `/home` exposes a second problem, which is the other
half of this change. `90-raw-disk-image.binary` mounts `/home` with
`nodev,nosuid` for every variant except `internal-buildserver`, which #879
exempted because appliance-build's image and upgrade builds create device
nodes and run setuid binaries in their working tree. DLPX-98978 put
`appliance-build` and `linux-pkg` into the clone list, so an agent on the
image running a build out of `~/src/appliance-build` would hit exactly that.
`internal-claude` therefore gets the same exemption. Both variants are
internal only and never shipped to customers, so relaxing the hardening there
is the same trade that was already accepted for the buildserver.

The exemption is written as a `case` rather than extending the `if`, to match
the variant-matching idiom the same file already uses for `RAW_DISK_SIZE_GB`:

```bash
HOME_MOUNT_OPTS="defaults,nodev,nosuid"
case "$APPLIANCE_VARIANT" in
internal-buildserver | internal-claude)
	HOME_MOUNT_OPTS="defaults"
	;;
esac
```

Worth noting for review: the clones are currently escaping `nodev,nosuid`
only because they land on the root dataset. Moving them to where they belong
is what puts them on the hardened filesystem, which is why these two changes
have to travel together rather than landing separately.
</details>

<details open>

### Local verification

- All six templated references render to the intended path. The role's
  `defaults/main.yml` and `tasks/main.yml` were loaded and the `path`, `dest`,
  `cmd`, and `content` values rendered through Jinja with the real defaults:

  ```
  [Create ~delphix/src directory]            path:    /home/delphix/src
  [Clone product repositories]               dest:    /home/delphix/src/appliance-build
  [Reset clone URLs to SSH]                  cmd:     git -C /home/delphix/src/appliance-build remote set-url origin git@github.com:delphix/appliance-build.git
  [Restore the wildcard fetch refspec]       cmd:     git -C /home/delphix/src/appliance-build config remote.origin.fetch +refs/heads/*:refs/remotes/origin/*
  [Ensure ~delphix/src is owned ...]         path:    /home/delphix/src
  [Add git-utils/bin to PATH ...]            content: export PATH="$PATH:/home/delphix/src/git-utils/bin"
  ```

  The quoting is the part worth having: `path:` had to become
  `"{{ claude_internal_src_dir }}"` because an unquoted `{` opens a YAML flow
  mapping, and the `profile.d` content keeps its single quotes so `$PATH`
  stays literal while the Jinja still expands.
- The mount-option `case` was exercised across the variant list, confirming
  the exemption is exactly two variants wide and nothing else moved:

  ```
  internal-buildserver     defaults
  internal-claude          defaults
  internal-dev             defaults,nodev,nosuid
  internal-qa              defaults,nodev,nosuid
  internal-dcenter         defaults,nodev,nosuid
  external-standard        defaults,nodev,nosuid
  external-dct             defaults,nodev,nosuid
  internal-minimal         defaults,nodev,nosuid
  ```
- `ansible-playbook --syntax-check` clean on
  `live-build/variants/internal-claude/ansible/playbook.yml`.
- `ansible-lint` (26.8.0, repo-wide, the same no-args invocation
  `.github/workflows/main.yml` uses) reports 2 fatal `fqcn[action]` failures,
  both pre-existing and in roles this PR does not touch
  (`unittest-internal:104`, `virtualization-development:115`). This role's own
  two findings are the pre-existing `command-instead-of-module` warnings on
  the two `git -C` tasks, both already marked `# ignored`. No new findings.
- `shellcheck -x` on `90-raw-disk-image.binary` reports 5 `note`-level
  findings, all on pre-existing lines (296, 387, 398, 418) and none at the
  edit. `-S warning` is empty, exit 0.
- `./gradlew check` is not a signal here, same as in #905: `:shfmtCheck` fails
  on an untouched `develop` on this workstation because the local `shfmt`
  (3.8.0) disagrees with CI's about a trailing `\` after `|`, in 7 shell files
  including this one. Confirmed no regression by diffing the tool's own output
  across the edit rather than trusting the task's exit code: 63 complaint
  lines before, 63 after, and none of them name `HOME_MOUNT_OPTS`,
  `APPLIANCE_VARIANT`, or `internal-claude`. The flagged-file set is identical
  to baseline. `bash -n` clean.

### ab-pre-push build #15023 — SUCCESS

- Build: https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build-orchestrator-pre-push/15023/
- Tested: `appliance-build@18aa32b` (against `develop`)
- Variants: `internal-claude`, platform `aws`, imported into DCenter as the same
  variant, `BUILD_ARTIFACTS=ALL`. No suites run (`RUN_TESTS=False`), so the build
  result is the whole signal from Jenkins.
- Result: `SUCCESS` in 2h46m. That is the half a build can cover: the role's clone
  tasks ran to completion against the new `/home/delphix/src` dest, which they could
  not have if the templated path were wrong, since a failed clone fails the build;
  and `90-raw-disk-image.binary` produced and imported a disk image with the new
  `case` in place. What it does not cover is the mount options the booted image
  actually reports -- see below.

### Review feedback

- The `DLPX-86523` reference is out of the `claude_internal_src_dir` comment; the
  comment still explains why the path is `/home` and why the old one would not work.
  Comment-only, so the tree is otherwise identical to what build #15023 tested.

### Not run

- The post-boot check no build can cover, on a clone of the imported image: that the
  52 repositories are in `~delphix/src` and reachable as `cd ~/src`, that
  `/export/home` does not exist, and that `findmnt /home` reports the unhardened
  options. Both variants here are internal only, and the failure mode if any of it is
  wrong is a missing clone tree or a too-hard-mounted `/home` on a development image
  rather than anything shipped, so this is landing without it.
</details>


live-build/misc/ansible-roles/appliance-build.claude-internal/defaults/main.yml
```diff
@@ -15,6 +15,16 @@
 #
 
 ---
+#
+# The directory the repositories below are cloned into. This is the delphix user's home
+# (see the "delphix" user in the appliance-build.minimal-common role), which DLPX-86523
```
david-mendez1: Remove the DLPX-86523 reference

david-mendez1 (approved):

dbjwhs-perforce (approved):
LGTM

live-build/misc/ansible-roles/appliance-build.claude-internal/defaults/main.yml
prakashsurya: Removed. The comment still says the home moved from `/export/home` to `/home` and that there is no compatibility symlink, since that is why the default is what it is, but it no longer names the ticket.

PR #863, opened 2026-04-02
DLPX-96879 Disable Influxdb default package

dbshah12:
<details open>
<summary><h2> Problem </h2></summary>

The `influxdb2` package ships with a default `influxdb.service` systemd unit that starts automatically when the package is installed. The performance-diagnostics package ([delphix/performance-diagnostics#119](https://github.com/delphix/performance-diagnostics/pull/119)) declares `influxdb2` as a runtime dependency and provides its own custom service wrapper (`delphix-influxdb.service`) that handles InfluxDB startup along with one-time initialization (org/bucket/token setup). If the default `influxdb.service` is not disabled, it will auto-start on install/boot and conflict with the custom service.
</details>

<details open>
<summary><h2> Solution </h2></summary>

Add `influxdb.service` to the list of services that are disabled and masked at appliance build time in the `appliance-build.virtualization-common` Ansible role. This follows the same pattern used for other services (e.g., `nginx.service`, `postgresql.service`, `telegraf.service`) that are installed by packages but managed by Delphix-specific service wrappers.

This ensures that when `influxdb2` is installed via the `performance-diagnostics` package, the default upstream service does not auto-start and the `delphix-influxdb.service` unit remains the sole entry point for managing InfluxDB on the appliance.
</details>

<details>
<summary><h2> Testing Done </h2></summary>

Covered by the ab-pre-push run in the dependent PR: [delphix/performance-diagnostics#119](https://github.com/delphix/performance-diagnostics/pull/119).
</details>

PR #907, opened 2026-09-23
DLPX-99011 Mask influxdb2's default influxdb.service during upgrade

dbshah12:
<details open>

The `influxdb2` package (a runtime dependency of `performance-diagnostics`, see [delphix/performance-diagnostics#119](https://github.com/delphix/performance-diagnostics/pull/119)) ships its own default `influxdb.service` systemd unit, enabled automatically by the package's postinst on install — the same situation `telegraf.service` is already in a few lines above this change. `delphix-influxdb.service` is Delphix's own wrapper around the same `influxd` binary, doing one-time org/bucket/token onboarding that `delphix-telegraf.service` depends on.

Neither unit declares any relationship to the other, so whichever wins the boot-time race for port 8086 after `influxdb2` is (re)installed determines the outcome:
- If the stock `influxdb.service` wins, `delphix-influxdb.service` can never bind the port and crash-loops 5x to `FAILED` (`Start request repeated too quickly`).
- The stock instance's own auth store is never onboarded with Delphix's tokens, so Telegraf (configured against Delphix's expected tokens) gets a permanent `401 Unauthorized` from whichever instance is actually listening.

[delphix/appliance-build#863](https://github.com/delphix/appliance-build/pull/863) already masks `influxdb.service` — but only in the `appliance-build.virtualization-common` Ansible role, which runs when a fresh appliance image is **built**. It does not cover an existing, already-deployed engine that upgrades `performance-diagnostics` into pulling in `influxdb2` for the **first time**, which is exactly the path `fix_and_migrate_services()` (this function) exists to handle for `nginx`, `postgresql`, `telegraf`, etc. `influxdb.service` was never added here when the InfluxDB/Telegraf infra (DLPX-96312) landed.

### Field evidence (ESCL-6163)

Two independent FedEx engines hit the identical signature, six days apart:

| | Engine | Bundle | Upgrade event | Outcome |
|---|---|---|---|---|
| 1st occurrence | `FXIAZEUS2NL001` | `/nas/support/fedex_express_uk/01657275/bundle1` (2026-09-16) | `performance-diagnostics` `1.0.0-1delphix.2025.08.27.17.30` → `1.0.0-1delphix.2026.08.28.04.41`, first-ever `influxdb2` install (`dpkg.log`) | Reboot → `influxdb.service` (PID 2652) won port 8086; `delphix-influxdb.service` crash-looped to `FAILED`; Telegraf got permanent 401s |
| 2nd occurrence | `FXIAZEUS2PL011` | `/nas/support/fedex_express_uk/01659990` (2026-09-22) | Same package version transition, same first-ever `influxdb2` install | Identical: `influxdb.service` (PID 2617) won port 8086, `systemctl status` showed it `active (running)` while `delphix-influxdb.service` was `failed`, still logging `Unauthorized: authorization not found` at bundle collection time |

Both engines are Azure/Hyper-V VMs (one object-storage-backed via Azure Blob on `domain0`, one plain Azure-managed-disk local storage) — this is not an OCI-specific or object-storage-specific race. The originally-filed DLPX-99011 diagnosis (missing `After=zfs-object-agent.service`, independent bolt/rpool dataset rollback, `delphix-influxdb-init`'s early-exit swallowing 401s) was investigated further against both support bundles and does not explain either failure: in both cases `systemctl status --all` shows the stock `influxdb.service` unit `enabled` and actively running, holding port 8086, while `delphix-influxdb.service` is the one left `failed`. PR delphix/performance-diagnostics#130 (which implements the DLPX-99011 fixes) would not have prevented either incident.
</details>

<details open>

Apply the same handling `telegraf.service` already gets in `fix_and_migrate_services()`:

1. Add `influxdb.service` to the heredoc list of services that get migrated/re-masked across upgrades. `should_mask_service()` treats a `not-found` unit as "should mask" (see comment above it), so on an engine that has never had `influxdb2` installed — exactly both ESCL-6163 engines' starting state — this masks `influxdb.service` (`systemctl mask --now` succeeds even for a not-yet-existing unit) **before** the upgrade ever runs `apt install influxdb2`. When `influxdb2`'s postinst later tries to enable/start it, systemd no-ops on the masked unit.
2. Add an explicit re-mask-if-enabled check after packages are upgraded, mirroring the `telegraf.service` block immediately above it (`influxdata/telegraf#14052`) — `influxdb2` is built with the same debhelper packaging as `telegraf` and may exhibit the same postinst behavior of restoring its shipped unit file (and reversing a pre-upgrade mask) during a package upgrade/reinstall, not just a fresh install. This is defensive parity with the telegraf case rather than a confirmed reproduction of the identical upstream issue against `influxdb2` specifically.

This function already runs in both places needed to close the gap PR #863 left open: `create_upgrade_container()` (not-in-place upgrade) and the in-place upgrade execute path.
</details>

<details>

- [ ] Verify `should_mask_service influxdb.service` masks correctly on an engine that has never had `influxdb2` installed (simulates both ESCL-6163 engines' pre-upgrade state)
- [ ] Upgrade such an engine through this code path and confirm `influxdb.service` stays masked and `delphix-influxdb.service` binds port 8086 cleanly across a reboot
- [ ] Confirm no regression for engines that already have `performance-diagnostics` (and therefore `influxdb2`) installed from a prior upgrade
</details>

## Jira

[ESCL-6163](https://perforce.atlassian.net/browse/ESCL-6163) / [DLPX-99011](https://perforce.atlassian.net/browse/DLPX-99011)


upgrade/upgrade-scripts/common.sh
```diff
@@ -592,6 +592,24 @@ function fix_and_migrate_services() {
 		mask_service telegraf "$container"
 	fi
 
+	#
+	# DLPX-99011: influxdb2 is built with the same debhelper packaging as
```
sebroy: Please refrain from referencing Jira IDs in code comments. The comment should be self-contained.

sebroy (approved):

upgrade/upgrade-scripts/common.sh
dbshah12: Good catch, thanks — removed the Jira ID reference from the comment in #909 (targets this branch, since direct pushes here are blocked by branch protection).

dbshah12:
Closing in favor of #909, which now targets `develop` directly and carries all the changes from this PR plus the follow-up fix for the review comment above.

